PR #1, opened 2022-03-18
Feature/base design

chloebrett:
Screenshots

Screen Shot 2022-03-18 at 4 43 12 pm

Screen Shot 2022-03-18 at 4 43 32 pm

### DIFF
--- a/src/renderer/components/NewProjectModal.tsx
+++ b/src/renderer/components/NewProjectModal.tsx
@@ -1,0 +1,20 @@
+import { Box, Modal, Typography } from "@mui/material";
+import SelectMediaBlock from "./SelectMediaBlock";
+
+interface Props {
+  isOpen: boolean;
+  closeModal: () => void;
+}
+
+const NewProjectModal = (props: Props) => {
+  const { isOpen, closeModal } = props;
+
+  return <Modal open={isOpen} onClose={closeModal}>
+    <Box>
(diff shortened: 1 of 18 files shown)


ghost:
Nice work! 
I think it might be a good idea to add a little bit of documentation to the README so that everyone knows how to set up their editors to automatically format (on save, etc) according to es-lint rules. At least for VS Code.

Optionally, we can also add a prettier config so that rules match and then people can choose default formatter in preferences.

src/renderer/components/NewProjectModal.tsx
```diff
@@ -0,0 +1,88 @@
...
+  }
+`;
+
+interface Props {
+  isOpen: boolean;
+  closeModal: () => void;
+}
+
+const NewProjectModal = (props: Props) => {
+  const { isOpen, closeModal } = props;
```
ghost: Recognising that this structure will likely get duplicated in future components, could we just put the object destructuring inside the component/function params?
https://stackoverflow.com/questions/52600056/destructuring-props-in-the-function-parameters-using-typescript

I realise its a very little thing. :)

ghost (changes requested):

src/renderer/components/NewProjectModal.tsx
chloebrett: Done

ghost (approved):

PR #2, opened 2022-03-19
Implement redux store, use for creating new projects and handling rec…

chloebrett:
…ent ones

Goals of this PR:
- Set up basic standards for how we do state management in our app
- Create a simple base for the application store that we can expand later as we have more uses for it
- Open the conversation about how we handle state - hooks or higher order components? Pure redux or redux-toolkit? Open to suggestions on all these fronts

Changes:
- Added redux store
- Using redux store for storing current project as well as list of recent projects
- Wired this into the app
- On form save for new project, adds the project to "recent projects" list (doesn't do anything about the actual media file yet, just mock data)
- If on macOS, starts up latest version of redux devtools in the electron window. Support for other OSs will come later, feel free to contribute this if you're on windows or linux. Basically just need to know the file path for chrome extensions on your OS
- Also made a couple other misc changes - typo fixes, removed the open collective message, etc


src/main/main.ts
```diff
@@ -9,12 +9,19 @@
...
 import path from 'path';
-import { app, BrowserWindow, shell, ipcMain } from 'electron';
+import os from 'os';
+import { app, BrowserWindow, shell, ipcMain, session } from 'electron';
 import { autoUpdater } from 'electron-updater';
 import log from 'electron-log';
 import MenuBuilder from './menu';
 import { resolveHtmlPath } from './util';
 
+// Load Redux DevTools on macOS (TODO: support other OSs)
```
fwang1703, line 23: Could we consider using electron-devtools-installer to load in redux devtools? https://github.com/MarshallOfSound/electron-devtools-installer

src/renderer/store/reducers.ts
```diff
@@ -0,0 +1,47 @@
...
+import { ApplicationStore, initialStore, Project } from './helpers';
+
+const currentProjectReducer: (
+  currentProject: ApplicationStore['currentProject'],
+  action: Action<any>
+) => ApplicationStore['currentProject'] = (
+  currentProject = initialStore.currentProject,
+  action
+) => {
+  if (action.type === PROJECT_CREATED) {
```
ghost, line 17: Just because this will be a template for future reducers, would it be a good idea to have a switch statement instead of multiple if statements - for readability ?

ghost:
Love your work

src/renderer/store/reducers.ts
chloebrett: To be honest I prefer if because switch kind of forces a particular format that gets kind of crowded if the logic gets complex

How about we leave it as if for now and if switch makes more sense we can change it later?

ghost: Yep, I'm okay with that

ghost (approved):

fwang1703 (approved):
Just fix up the merge conflict and all good to merge

src/main/main.ts
chloebrett: made a trello card for this

chloebrett: https://trello.com/c/kOalGhZu/39-use-electron-devtools-installer

PR #3, opened 2022-03-19
Python server

MutazAshhab:
Goal of this PR is to set a basic python backend to enable the development of the ML stuff

Changes:
- Added `child_process` dependency
- Added flask server and `requirements.txt`
- On app launch, the server is started and logs if the server started correctly
- add `__pycache__` to `.gitignore`

Notes:
1. For the app to launch now you have to have python installed on your machine
2. You also have to go into `mlvet/src/py_server` and run `pip install -r requirements.txt`


src/main/main.ts
```diff
@@ -90,10 +94,30 @@ const createWindow = async () => {
     } else {
       mainWindow.show();
     }
+
+    pyServer = startServer();
+
+    if (pyServer !== null && pyServer.stderr !== null) {
+      pyServer.stderr.once('data', (data) => {
```
chloebrett, line 105: Is there a way to get this logging to stdout instead of stderr? Using stderr seems like a bit of a hack

src/main/main.ts
```diff
@@ -90,10 +94,30 @@ const createWindow = async () => {
...
+        console.log(data.toString()); // prints address that the server is running on. TODO: remove before it is electron app is built
+        get('http://localhost:5000', (res) => {
```
chloebrett: Can we extract the port into a constant so it's configurable?

src/main/main.ts
```diff
@@ -90,10 +94,30 @@ const createWindow = async () => {
     } else {
       mainWindow.show();
     }
+
+    pyServer = startServer();
+
+    if (pyServer !== null && pyServer.stderr !== null) {
+      pyServer.stderr.once('data', (data) => {
+        console.log(data.toString()); // prints address that the server is running on. TODO: remove before it is electron app is built
```
chloebrett: I guess we can leave this in for now, but might want to check NODE_ENV before logging so we don't log in prod

src/main/main.ts
```diff
@@ -90,10 +94,30 @@ const createWindow = async () => {
...
+            );
+          }
+        });
+      });
+    }
   });
 
   mainWindow.on('closed', () => {
     mainWindow = null;
+    if (pyServer !== null && typeof pyServer.pid === 'number') {
```
chloebrett: Why do we need to check the process ID's type?

src/py_server/background_removal.py
```diff
@@ -0,0 +1,2 @@
+def remove_background():
+    raise NotImplementedError
```
chloebrett, line 2: Nice 👍🏻

src/py_server/requirements.txt
```diff
@@ -0,0 +1 @@
+Flask==2.0.3
```
chloebrett, line 1: Please add install instructions to the readme

chloebrett (changes requested):
Looking good! Just a few small things and a couple of Q's

Also, please target `develop` (I know my open PR was hitting main as well, I've updated it)

src/py_server/requirements.txt
MutazAshhab: Done in [f73da1f](https://github.com/patrickbrett/mlvet/pull/3/commits/f73da1f43abb305633ebdb397931a95d0950249a)

src/main/main.ts
MutazAshhab: The flask app logs to both stdout and stderr.

This gets logged to stdout
```
 * Environment: production
   WARNING: This is a development server. Do not use it in a production deployment.
   Use a production WSGI server instead.
 * Debug mode: off
```
This gets logged to stderr
```
 * Running on http://127.0.0.1:5000/ (Press CTRL+C to quit)
```

It should ping the server once flask displays this `Running on http://127.0.0.1:5000/ (Press CTRL+C to quit)` since that indicates that the server is now up and running.

Flask also logs out the requests it gets to stderr.

If I where to change it to log everything to stdout then I would have to change the event listner from `once` to `on` (since it will no longer be the first thing that is logged) and only execute the health check if `Running on http://127.0.0.1:5000/ (Press CTRL+C to quit)` is part of what is logged. 

It would also cause the event listener to fire off unnecessarily for every request it recieves.

src/main/main.ts
MutazAshhab: good idea, im going to use the dotenv package just so we don't have to deal with exporting env variables on all 3 OS's

src/main/main.ts
MutazAshhab: Oops, removed it.

src/main/main.ts
MutazAshhab: Of course, its now in the .env file.

MutazAshhab:
Note:

In `src/main/py_server.ts`

I tried:
```typescript
  let options = '';

  if (process.env.FLASK_PORT !== undefined) {
    options = `--port=${process.env.FLASK_PORT}`
  }
  proc = spawn('flask', ['run', options]);
```
But it failed, because of the empty string. So i had to do this:
```typescript
  let proc = null;

  if (process.env.FLASK_PORT !== undefined) {
    proc = spawn('flask', ['run', `--port=${process.env.FLASK_PORT}`]);
  } else {
    proc = spawn('flask', ['run']);
  }
```

src/main/main.ts
chloebrett: Okay, sure - nice analysis. Perhaps add a comment as a quick explainer so this doesn't come up again

chloebrett (approved):
Looks good

src/main/main.ts
MutazAshhab: Okay, Im going to link to this discussion as well.

MutazAshhab:
@patrickbrett I made final changes, ready to merge now.

PR #4, opened 2022-03-24
fix debugging for main and render threads

jhea0008:
Fixes debugging for both main and renderer process when started through visual studio code. Based on https://github.com/electron-react-boilerplate/electron-react-boilerplate/pull/3169/files PR, which has not yet gone through to the boilerplate master, but the changes seem to work for our repository


Mhudlark:
Looks good. Thanks for picking up on their issues :)

Mhudlark (approved):

PR #8, opened 2022-04-15
Feature/open media file

Irmy:
Removed unnecessary boiler plate code regarding ipcrenderer.on
Connected import UI to renderer function and main listener. 
Linked media import to new project action.

PR #9, opened 2022-04-16
Expose ipc methods rather than ipcRenderer

MutazAshhab:
Note, this change depends on https://github.com/patrickbrett/mlvet/pull/8 as it updates the exposed ipcRenderer.

<hr>

Goal of this PR is to expose methods instead of exposing the entire ipcRenderer

Changes:

- Removed ipc.ts as it declared and exported the `requestMediaDialog` function (`requestMediaDialog` now defined in preload.js)
- Added `requestMediaDialog` and removed `ipcRender` from global.d.ts
- Updates SelectMediaBlock component to use the exposed method rather than the exposed ipcRenderer.

### DIFF
--- a/src/main/preload.js
+++ b/src/main/preload.js
@@ -1,6 +1,11 @@
 const { contextBridge, ipcRenderer } = require('electron');
 
+/*
+ * When exposing a new method make sure to update global.d.ts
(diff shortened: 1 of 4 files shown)


chloebrett:
Will wait on #8 merge before reviewing

chloebrett (approved):

PR #11, opened 2022-04-17
Feature/save project

chloebrett:
Depends on #10 

- When transcription is complete, goes to bare bones project page
- Save button in corner
- Clicking save opens save dialog and then writes currentProject state from store into file


chloebrett:
Added more changes to this:

* Support opening project files
* Add file submenu - save and open with keyboard shortcuts
* Refactor IPC so that main can send messages to renderer as well as vice versa
* Pull out react-router; we don't need it since we don't have URLs. Instead use an enum on the redux store root, which is updatable outside the react context. Design doc: https://docs.google.com/document/d/10ACSzFhQGU8RoE0LwWMAwbr2v7NUSyj2SPeXwrzb-a0/edit

PR #12, opened 2022-04-17
Feature/audio extract

MutazAshhab:
Goal of this PR is to add functionality so we can extract audio from a given media file.

Changes:

- demo-video.mp4 and audio.wav are added to `.gitignore`
- 2 new app dependencies
  - fluent-ffmpeg
  - ffmpeg-installer/ffmpeg
- made misc changes. e.g. removed comments and added types


tngu0041 (approved):
Conversions all look good to me.

PR #17, opened 2022-04-29
Feature/ml transcribe backend - transcription handler interface

jhea0008:
Add interfaces and casting code to handle transcription data to be received from python API, and return it from the transcriptionHandler promise. 
For now, this is set to return a hard coded transcription read from SampleTranscript.json, but this will eventually be changed to receive the json from the python thread


LiamTodd (approved):

v1ta1287 (commented):
Looks good

PR #19, opened 2022-04-30
Make IPC listeners work on client

chloebrett:
Gonna add all the optional PR template things to this 3 liner PR 😎

**Why is this change needed?** Because we like it when the save and open file dialogs actually work

**What did you change?** Imported the IPC listeners on the client since they don't get run otherwise

**Does it depend on any other changes?** Nope

**Screenshots of UI changes** N/A

**Explain things** Well, basically if you don't import a file it won't run, so I imported it so that it runs

**Testing details** Hit Ctrl+O from the home screen, open a project and confirm that it takes you to the opened project page

**Any sections in particular to discuss?** The blank line on line 6 is the most important part

**Links to design docs** N/A

### DIFF
--- a/src/renderer/index.tsx
+++ b/src/renderer/index.tsx
@@ -1,4 +1,7 @@
 import { render } from 'react-dom';
 import App from './App';
 
+// Make IPC listeners initialise themselves
+import './ipc';
+
 render(<App />, document.getElementById('root'));


ghost (commented):
Beautiful!

ghost (approved):

PR #20, opened 2022-04-30
Feature/export

zwan0202:
Why is this change needed?
goal of this PR is to add functionality so we can export project to EDL
What did you change?
added export functionality and modified the mock type for Transcription


itsMoxMox (approved):

itsMoxMox (changes requested):
Will just need to address the merge conflict where dev's sharedTypes references a now outdated mockup for Transcription.

ghost (approved):

src/main/export.ts
```diff
@@ -0,0 +1,49 @@
+/* eslint-disable no-plusplus */
+import { join } from 'path';
+import { Project, Transcription } from '../sharedTypes';
+import { padZeros, integerDivide, writeFile, mkdir } from './util';
+
+const secondToTimestamp = (num: number) => {
```
chloebrett: For consistency let's make sure all function signatures have a return type, like this:

`const secondToTimestamp: (num: number) => string = (num) => { ... }`

src/main/export.ts
```diff
@@ -0,0 +1,49 @@
...
+  return `${padZeros(integerDivide(num, 3600), 2)}:
```
chloebrett: Might want to abstract this a little further, up to you what you prefer

e.g. `[3600, 60, 1, 0.01].map(mult => padZeros(integerDivide(num, mult), 2)).join(':')`

That said it's kind of hard to understand as is, maybe add a comment describing what the timestamp looks like, and/or constants for the numeric values (to avoid 'magic numbers')

src/main/export.ts
```diff
@@ -0,0 +1,49 @@
...
+          ${padZeros(integerDivide(num, 1), 2)}:
+          ${padZeros(integerDivide(num, 0.01), 2)}`;
+};
+
+const constructEDL = (title: string, transcription: Transcription) => {
+  let output = `TITLE: ${title}\nFCM: NON-DROP FRAME\n\n`;
+  try {
+    const { words } = transcription;
+    const entries = words.length;
+    for (let i = 0; i < entries; i++) {
```
chloebrett: Could be cleaner to do this with `map` since every iteration of the loop is just producing a new line of output

src/main/export.ts
```diff
@@ -0,0 +1,49 @@
...
+      const currWord = words[i];
+      const editStart = secondToTimestamp(currWord.start_time);
+      const editEnd = secondToTimestamp(
+        currWord.start_time + currWord.duration
+      );
+
+      output += `${edlEntry}\t${editStart}\t${editEnd}\n* FROM CLIP NAME: sample\n\n`;
+    }
+  } catch (e) {
+    console.error(e);
```
chloebrett: What errors are you catching here?

src/main/export.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  if (project.transcription) {
+    writeFile(
+      join(project.savePath, `${project.name}.edl`),
+      constructEDL(project.name, project.transcription)
+    );
+  }
+};
+
+const exportFuncs = {
```
chloebrett: What's wrong with just exporting default exportEDL?

src/main/util.ts
```diff
@@ -16,3 +17,29 @@ if (process.env.NODE_ENV === 'development') {
...
+
+export const integerDivide = (a: number, b: number) => {
+  return Math.floor(a / b);
+};
+
+export const writeFile = (filePath: string, data: any) => {
+  try {
+    writeFileSync(filePath, data, { flag: 'w' });
+  } catch (err) {
+    console.error(err);
```
chloebrett: Hm, where is this being used and would it be better to have that function throw? We shouldn't normally get errors writing a file and if we do we definitely don't want to continue like nothing happened

src/sharedTypes.ts
```diff
@@ -1,10 +1,17 @@
 // TODO: Define transcription schema (this is just a mockup)
+export type Transcription = { words: [Word] };
+
+export type Word = {
+  word: string;
+  start_time: number;
```
chloebrett: Camelcase attributes

src/main/export.ts
```diff
@@ -0,0 +1,49 @@
...
+
+const secondToTimestamp = (num: number) => {
+  return `${padZeros(integerDivide(num, 3600), 2)}:
+          ${padZeros(integerDivide(num, 60), 2)}:
+          ${padZeros(integerDivide(num, 1), 2)}:
+          ${padZeros(integerDivide(num, 0.01), 2)}`;
+};
+
+const constructEDL = (title: string, transcription: Transcription) => {
+  let output = `TITLE: ${title}\nFCM: NON-DROP FRAME\n\n`;
```
chloebrett, line 18: Can you add a link to your write up (I think you made a write up?) on what the structure of an EDL is like? A G Drive link ideally

src/main/export.ts
```diff
@@ -0,0 +1,49 @@
...
+import { padZeros, integerDivide, writeFile, mkdir } from './util';
+
+const secondToTimestamp = (num: number) => {
+  return `${padZeros(integerDivide(num, 3600), 2)}:
+          ${padZeros(integerDivide(num, 60), 2)}:
+          ${padZeros(integerDivide(num, 1), 2)}:
+          ${padZeros(integerDivide(num, 0.01), 2)}`;
+};
+
+const constructEDL = (title: string, transcription: Transcription) => {
```
chloebrett: This is a good candidate for some unit testing, fine to do this in a later PR, would you be confident in adding some tests with jest?

src/sharedTypes.ts
```diff
@@ -1,10 +1,17 @@
...
+  duration: number;
+};
 
-import Transcription from 'sharedTypes/Transcription';
 
 export interface ProjectBase {
   schemaVersion: number;
   name: string;
+  savePath: string;
```
chloebrett, line 4: Not sure if this should be stored on the project entity itself, or just part of the state that initialises when we open a file (since the project file could be moved), but this is fine for now

chloebrett (changes requested):
Looking good! Your yarn.lock is a little messed up, not sure why - maybe discard your changes to it and re-run yarn

Also, we don't want to have a package-lock.json, that's a symptom of running `npm install` which we don't need to do since we are using yarn.

I would suggest
1. Delete package-lock.json
2. Reset your changes to yarn.lock: `git checkout develop; git pull; git checkout feature/export; git checkout develop -- yarn.lock`
3. Re-run yarn
4. If that doesn't work, delete your node_modules directory and try steps 2-3 again

Everything else is pretty nice, just added a few comments for places you could simplify things and some minor style stuff

src/sharedTypes.ts
ghost: Might be useful to store the prev location of export. At the very least, maybe rename to `exportPath` or `exportDefaultPath` rather than `savePath`. Might be confusing is all.

src/main/util.ts
ghost: I reckon Patrick is right. Fail fast. 
I have added a ticket to our team backlog to display an error to the user when export fails.

ghost (commented):
Waiting on suggested changes.

src/main/export.ts
zwan0202: Right, I'll keep that in mind

src/main/export.ts
zwan0202: Thanks, that looks much cleaner

src/main/export.ts
zwan0202: Got it

src/main/export.ts
zwan0202: I can try

chloebrett: Great, feel free to reach out (to myself or anyone else familiar with jest) if you need a hand

src/main/export.ts
zwan0202: Not me, I think Max wrote that

zwan0202:
> Looking good! Your yarn.lock is a little messed up, not sure why - maybe discard your changes to it and re-run yarn
> 
> Also, we don't want to have a package-lock.json, that's a symptom of running `npm install` which we don't need to do since we are using yarn.
> 
> I would suggest
> 
> 1. Delete package-lock.json
> 2. Reset your changes to yarn.lock: `git checkout develop; git pull; git checkout feature/export; git checkout develop -- yarn.lock`
> 3. Re-run yarn
> 4. If that doesn't work, delete your node_modules directory and try steps 2-3 again
> 
> Everything else is pretty nice, just added a few comments for places you could simplify things and some minor style stuff

I deleted node_modules and redo the steps 2-3, not sure why all the `resolved` path in my `yarn.lock` is `npmjs` instead of `yarnpkg`

src/main/export.ts
```diff
@@ -0,0 +1,48 @@
...
+  title,
+  transcription
+) => {
+  let output = `TITLE: ${title}\nFCM: NON-DROP FRAME\n\n`;
+  const { words } = transcription;
+  const entries = words.length;
+
+  output += words
+    .map((word, i) => {
+      const edlEntry = `${padZeros(i + 1, entries)}\tAX\tAA/V\tC`;
```
chloebrett, line 24: Nice!

src/sharedTypes/Word.ts
```diff
@@ -1,7 +1,7 @@
 interface Word {
   word: string;
 
-  start_time: number;
+  starTime: number;
```
chloebrett: startTime

chloebrett (changes requested):
Nearly there, still need to delete package-lock.json though. Great work taking the suggestions on board :)

There are a couple of small merge conflicts now by the way, so will need to fix those (merge develop into this branch again)

chloebrett (approved):
LGTM!

chloebrett:
@itsMoxMox conflicts resolved, please re check

itsMoxMox (approved):

PR #21, opened 2022-04-30
Clean up transcription code syntax; rename some files; move some impo…

chloebrett:
…rts around

- camelCased some variable names
- camelCased some file names
- Folded transcription and word types back into main sharedTypes declaration
- Deleted background removal placeholder code


src/sharedTypes.ts
```diff
@@ -1,14 +1,21 @@
...
-import Transcription from 'sharedTypes/Transcription';
-
 export interface ProjectBase {
   schemaVersion: number;
   name: string;
   filePath: string;
   transcription: Transcription | null;
 }
 
+export interface Transcription {
```
jhea0008, line 8: Wondering what the rationale is for having all of our shared types in one file instead of having a folder of individual files?

jhea0008 (approved):

src/sharedTypes.ts
chloebrett: When this gets bigger we might want to split it - for now I couldn't find a way to split it and have an index so that imports all happen from the same file, so I felt the tradeoff of including some extra stuff in here was fine. I'm sure there's a way to do it, but couldn't find it in like 10 mins of experimentation

The file is like 30 lines, it's perfectly fine to have the types together for now. I would've left them split if the imports were working as expected, but yeah it was making us import twice if we wanted, say, Transcription as well as something in the main sharedTypes file

PR #22, opened 2022-04-30
Refactor reducers into separate files

chloebrett:
Why? Reducers are getting complex, we need to separate them so they're not all in the same file

What?
- Reducers split into multiple files
- Added a transcription reducer for handling transcription related actions


ghost:
Great stuff. Was thinking about this the other day. Do you think it is worth refactoring the actions into separate files too? For the same reason.

One possible file structure might be to have a folder for each reducer/set of actions eg: 


/store
    /Page
        page.actions.ts         (or just       actions.ts)
        page.reducer.ts
    /Project
        project.actions.ts
        project.reducer.ts

chloebrett:
> Do you think it is worth refactoring the actions into separate files too? For the same reason.

Maybe - the hard things is that sometimes multiple reducers handle a single action. E.g. PROJECT_CREATED both updates the currentProject and adds it to the recentProjects list - two different reducers.

I'm sure if/when the actions file gets too big we'll find a sensible way to split it, it's well under 100 lines for now so I'm fine with leaving it as is

> Great stuff

Is that an approval? 😛

ghost (approved):

PR #23, opened 2022-04-30
Framework for undo/redo functionality

chloebrett:
See this design doc: https://docs.google.com/document/d/1fBLBj_I3Y4AgRnIHzJ-grsXvzoKUBA03KNRv3DzABAg/edit

Basically everything is covered there. (e.g. what the heck is an op, etc, etc)

That said:

Brief reasoning on why we want to do this now -> because it will make our lives WAY easier later on (undo/redo is not a "tack it on when we already have stuff mostly built" kind of feature)

Details of changes:

I originally was going to put through a bigger PR - something like 500 line additions. Obviously that's hard to review, so I've tried (key word is tried) to split it up. I don't really think it can get much more granular than this ~250 line addition PR, so apologies to the reviewers, but I think you'll live (following the "would you review this yourself rule", ya I would, undo/redo is cool 😎)

PR includes:
- adding functional undo/redo buttons to the edit menu, and disabling them when there's nothing to undo/redo
- adding an undo stack to the redux store and generalised reducers for manipulating it
- helpers for dispatching undos, redos and ops
- also a small, unrelated change to the project schema to broaden the types a bit because the narrow typing of video/audio projects was getting annoying

What it doesn't include:
- the actual "dummy" ops - the two I suggested in the design doc were (a) deleting half the words in the transcription and (b) replacing a random word with "swamp". Both dumb, but also both great for demoing this functionality before we have the actual edit/delete stuff implemented.
- any kind of UI for this demo

See the demo PR which includes these two things: https://github.com/patrickbrett/mlvet/pull/24

That's probably the one you want to check out locally and play around with while you review this one :)


src/renderer/store/reducers/undoStackReducer.ts
```diff
@@ -0,0 +1,57 @@
...
+    action.type === PROJECT_OPENED ||
+    action.type === CURRENT_PROJECT_CLOSED
+  ) {
+    return initialStore.undoStack;
+  }
+
+  if (action.type === UNDO_STACK_PUSHED) {
+    const { stack, index } = undoStack;
+
+    const newStack = stack
```
ghost, line 26: Nice. So just to confirm, doing an action overwrites any possible redo actions from further up the stack?

chloebrett: yup - otherwise we'd have an undo tree and that would get messy fast lol

fwang1703 (approved):
Looks good!

src/renderer/store/reducers/undoStackReducer.ts
chloebrett: https://www.youtube.com/watch?v=D9laah5bcio&ab_channel=PatrickBrett

if you have time, this explains in detail (check the timestamps for the specific section)

ghost: agreed.

lol didn't see vid until 1 min ago.

I drew up a diagram if you want to include in resources.

https://www.figma.com/file/ya0gV3UM0nQmXWc8SfkneQ/Design-Docs?node-id=8%3A18

ghost (approved):

MutazAshhab (approved):

PR #24, opened 2022-04-30
Provide demo usage of undo stack

chloebrett:
~~Depends on: https://github.com/patrickbrett/mlvet/pull/23 (line count will drop to like +250 when that one gets merged)~~ 

Provides demo usage of undo stack, with two dummy ops:
- delete half the words in the transcript
- replace a random word with 'swamp'

To test:
- open a project
- press the op buttons (delete half the words, replace a word with swamp) in various combinations
- check that the undo and redo react buttons work
- check that using ctrl+z / ctrl+shift+z also works
- check that the edit menu buttons work
- check that they're greyed out at the same time as the react buttons
- check that undoing a few times, then hitting an op button again (e.g. delete half the words), wipes out the option to redo the things that you undid
- check that no matter what you do, if you hit undo enough you'll end back with the initial state of the transcription

While you're at it feel free to check out the redux state from the debugger if you wanna play with it live, I find it kind of satisfying lol


chloebrett:
PR that this depends on was merged, ready for review

Taz17:
Here is the current state of the UI whilst running the app in linux.

MutazAshhab:
Here is the current state of the UI whilst running the app on linux

image

I put this here to show that the "Menu" option is not visible.

Additionally, the shortcuts CTRL + Z and CTRL = SHIFT = Z do not work.

fyi, i did also test that other shortcuts work e.g. CTRL + ALT + I to open the dev tools

chloebrett:
Weird. I will test on my linux machine and get back to you. Looks like the whole edit menu isn't showing

chloebrett:
Fixed in this PR https://github.com/patrickbrett/mlvet/pull/37/files

MutazAshhab:
Yup, I can now see the `edit` menu and the shortcuts work now.

image

MutazAshhab (approved):

emmapainter:
Didn't see this at the time, but do we really want to be merging demos into our develop branch?

PR #25, opened 2022-05-01
Feature/deepspeech model

tngu0041:
**Pytest** -- Using pytest for the python unit tests. But if anyone has any strong opinions for a different testing library feel free to suggest.

**Pre-trained model and scorer** -- I'm not pushing these to git because they're quite large (about 1Gb). But remember we probably need to include these in the installation package for the final app.

**Why am I using the Deepspeech command-line API instead of in-code library?** -- For some reason they give different outputs. The in-code library outputs per-letter transcription (which is more information than we need), whereas command-line API outputs per-word transcription. There's no way to make the in-code API output per-word from what I've read in the documentation.

**Why am I using subprocess and writing the output to a text file??** -- The raw output contains a load of unneeded metadata (like warnings, tensorflow version, etc), which is why I'm redirecting the output to a text file, reading the text file, and then deleting the text file (for cleanliness).


kateahyde:
A few things:
- I'm happy with Pytest although I think your test is more testing deepspeech itself than your own code - in terms of testing our own code, I think we should be assuming that the packages we're using are doing what they're supposed to (e.g. in this case outputting the correct/expected words and confidence ratings) as it's not our responsibility to check that code - that's on the people writing it (otherwise we could go on for ages testing everything in all the packages we're using). Here maybe you could mock the output of deepspeech (e.g. use a text file generated by running the code so you know it's exactly right) and then check that the final output of the function matches what you're expecting based on what's in the text file, as your code is more about extracting the output than the deepspeech model returning the correct results. This would also mean we wouldn't have to upload an audio file (not sure if this has similar issues using across different OS like video files?). Let me know if this doesn't make sense or if you have a different opinion.

- Totally on board with including the download of the model itself in the instructions and requirements, although should we maybe be including it in the repo regardless? I think I could be persuaded either way, I'm just not completely sure which is better so thought I'd open the discussion

- Mutaz and I were discussing a potential alternative solution for the API I'm working on where the output is pushed to a file rather than returned directly, although I'm thinking it probably still makes sense for this code to return the json object as it'll mean the API is free to do whatever is most suitable/abstracts this decision away from the transcription itself which I think makes sense.... basically I think what you've got (txt and converting back) is good just wanted to check with @Taz17 that we're on the same page

More than happy to discuss any of this and to hear anyone else's thoughts and opinions if they're different to mine - these are just my initial thoughts :)

chloebrett:
> I'm happy with Pytest although I think your test is more testing deepspeech itself than your own code

Mostly agree, though I think in this case it does validate that the model is installed and working as expected which is valuable - it's an integration test really

> should we maybe be including [the 1gb model] in the repo regardless?

I think installing it separately is probably the way to go, otherwise this repo will take forever to clone. Maybe add a shell script to install it and just tell people to run that. Also github max repo size is 2gb so we'd be pushing that limit somewhat (though still within it)

> Mutaz and I were discussing a potential alternative solution for the API I'm working on where the output is pushed to a file rather than returned directly

What's the proposed benefit of this?

--

Code looks really good btw Thien, and love your explanations in the PR description!

kateahyde:
> Mostly agree, though I think in this case it does validate that the model is installed and working as expected which is valuable - it's an integration test really

Agree with this, I was more focusing on the test as a unit test but you're right, it's also valuable to include the integration test. Maybe we should have both in there? Also is there anything we need to be mindful of with .wav files across different OS or is it not an issue?

> Also github max repo size is 2gb so we'd be pushing that limit 

Ah right I wasn't aware of this - a very valid reason not to push it so let's stick with that then, thanks for the clarification

> What's the proposed benefit of this?

I'm planning to do a GET endpoint which will return the transcription directly, however we were discussing the potential for the request to timeout if it's a large video file. We were contemplating a different approach where the initial call to the GET endpoint starts a thread that runs the transcription in the background and returns a 200 OK (or similar) immediately after starting the thread, and then the calling Node code will have to poll another endpoint for the status on the transcription and when it's complete it will have a reference to a file containing the output. This is a bit rough atm, and I was planning to flesh it out a bit more and look at if and how it would work once I have the basic GET fully implemented.

chloebrett:
> the calling Node code will have to poll another endpoint

Sockets are a good solution for this use case! That way python can push the completed transcription instead of node having to poll

kateahyde:
True, good point! I can't remember exactly why we thought an API would be better but if it's not going to do what we need then a socket approach is defs worth looking into (and will hopefully be simpler than creating threads and trying to keep track of their status) :)

tngu0041:
>Here maybe you could mock the output of deepspeech (e.g. use a text file generated by running the code so you know it's exactly right) and then check that the final output of the function matches what you're expecting based on what's in the text file, as your code is more about extracting the output than the deepspeech model returning the correct results. This would also mean we wouldn't have to upload an audio file (not sure if this has similar issues using across different OS like video files?). Let me know if this doesn't make sense or if you have a different opinion.

Personally, I still think there's value to testing the deepspeech library, even if its more of an integration like Patrick said. I'm happy to mock the function call if there's issues with handling the audio file on other OS though.

kateahyde:
Yeah sounds good, I agree there's merit to the integration test. If you could do a unit test as well though mocking the txt file I think that would cover everything :)

README.md
```diff
@@ -30,6 +30,17 @@ pip install -r requirements.txt
 
 This will install all dependencies required to run the python server.
 
+For Deepspeech transcription to work, you will need to also add the pre-trained model and scorer to `src/py_server`
```
MutazAshhab: Since there was a pull request that refactored the code and directories to follow a camelCase naming convension could you change `src/py_server` to `src/pyServer`

MutazAshhab:
Other than that ^ small naming fix, the code looks very good to me. 

@patrickbrett @kateahyde it would be nice to continue the conversation of web sockets vs polling in a github issue so that everyone that is not watching this PR could keep track.

MutazAshhab:
its not allowing me to re-review/approve. but its fine by me now!

PR #31, opened 2022-05-01
Feature/home screen UI

chloebrett:
~~Depends on #35~~

Includes:
* Recent projects UI to match figma designs with example thumbnail (not yet loading actual thumbnails)
* Date modified and media size are real
* Clicking the project, or clicking the delete button, doesn't do anything yet
* Wired up some redux stuff to get the project metadata injected into the front end
* Some cheeky subtle animation
* The delete button doesn't do anything yet
* Clicking a recent project opens it

Screenshots:
Screen Shot 2022-05-04 at 3 55 27 pm
Screen Shot 2022-05-04 at 4 04 54 pm


src/renderer/components/NewProjectBlock.tsx
```diff
@@ -1,34 +1,69 @@
...
-    <NewProjectBox>
-      <GenericSquareBox onClick={showModal}>
-        <AddIcon fontSize="large" />
-        <Typography fontWeight="bold">New Project or Drag Media</Typography>
-      </GenericSquareBox>
+    <NewProjectBoxWrapper>
+      <BoxComponent onClick={showModal} elevation={3}>
+        <Stack alignItems="center" className="MuiStack-root">
+          <AddIcon sx={{ color: colors.yellow[500], fontSize: 72 }} />
+          <Typography>New project</Typography>
```
emmapainter: Change to "New Project"

src/renderer/components/RecentProjectsBlock.tsx
```diff
@@ -1,36 +1,121 @@
...
+import { projectOpened, pageChanged } from 'renderer/store/actions';
+import { ApplicationPage, ApplicationStore } from '../store/helpers';
+import colors from '../colors';
+import { formatDate } from '../util';
+import exampleThumbnail from '../../../assets/example-thumbnail.png';
 
 const RecentProjectsBox = styled(Box)`
-  background: ${colors.grey[900]};
-  width: calc(100vw - 80px);
+  width: calc(100vw - 40px);
```
emmapainter, line 11: The recent project listings and new project box should be the same width.

src/renderer/theme.tsx
```diff
@@ -20,6 +20,9 @@ const theme = responsiveFontSizes(
       },
     },
     typography: {
+      body1: {
```
emmapainter, line 23: what is this for?

src/renderer/components/RecentProjectsBlock.tsx
```diff
@@ -1,36 +1,121 @@
...
 
   return (
     <RecentProjectsBox>
-      <Typography fontWeight="bold">Recent Projects</Typography>
-      <ProjectsListBox>
-        {recentProjects.map(({ id, name }) => (
-          <GenericSquareBox key={id}>
-            <Typography fontWeight="bold">{name}</Typography>
-          </GenericSquareBox>
+      <Typography variant="h6" style={{ marginBottom: 20, fontWeight: 'bold' }}>
```
emmapainter: Use font styles defined by @hhen0005 in theme.tsx - I believe this is h3

emmapainter (commented):
Looking much better! Just a few small comments.

emmapainter:
If possible, could you change the scrolling indicator to use dark mode? Is there a default dark mode option or would we choose the colours explicitly? 
image

src/renderer/theme.tsx
chloebrett: Defaults text to light grey instead of black

src/renderer/components/RecentProjectsBlock.tsx
chloebrett: Done

src/renderer/components/NewProjectBlock.tsx
chloebrett: Done

src/renderer/components/RecentProjectsBlock.tsx
chloebrett: H3 is too small, let's do what appears right for now and match it up once the sizes are properly defined

chloebrett:
> If possible, could you change the scrolling indicator to use dark mode? Is there a default dark mode option or would we choose the colours explicitly? image

Haven't looked into this, I think we'd have to set an app wide dark mode. It's also possible to make a custom scrolling indicator though will need testing cross platform, seems out of scope for this but good point to note

src/renderer/components/RecentProjectsBlock.tsx
emmapainter: Figma has defined everything - do the sizes/measurements given on Figma look different when you try to implement them?

chloebrett: Okay, initially I thought this was just because the figma frame size was quite small (height = 752, 13 inch macbook = 900 and most monitors higher)

But, actually it was because the theme variants were not working and everything was defaulting to 16px when I used them.

I've fixed them though this has involved ditching the whole 'h-100' and 'h-300' idea - MUI creates elements with those types which is not a good idea since they are not actually html element types! I've changed it back to use h1 and h3. @hhen0005 please review, you're more across this theming stuff than I am

chloebrett: Anyway, using h1 and h3 now and they look fine

src/renderer/theme.tsx
```diff
@@ -20,16 +20,19 @@ const theme = responsiveFontSizes(
...
+        color: colors.grey[300],
+      },
       fontFamily: ['Rubik', 'sans-serif'].join(','),
-      'h-100': {
+      h1: {
         fontSize: '24px',
-        lineHeight: '36px',
+        lineHeight: 36 / 18, // 36px
```
ghost, line 29: Just curious, what is the benefit of this?

src/renderer/components/NewProjectBlock.tsx
```diff
@@ -1,34 +1,69 @@
...
-import GenericSquareBox from './GenericSquareBox';
 import ModalContainer from './NewProjectModal/ModalContainer';
+import colors from '../colors';
 
-const NewProjectBox = styled(Box)`
-  background: ${colors.grey[900]};
-  width: calc(100vw - 80px);
+const NewProjectBox = styled(Paper)`
+  background: ${colors.grey[700]};
+  width: calc(100% - 40px);
```
ghost, line 9: No need to change but for future reference. Because css property `box-sizing` is set to "border-box", can just add padding to parent element and set child to be width 100% instead of doing margin and then calculating width as 100% - margin in the child element

src/renderer/components/NewProjectBlock.tsx
```diff
@@ -1,34 +1,69 @@
...
   display: flex;
   align-items: center;
   justify-content: center;
+
+  color: ${colors.grey[300]};
+
+  transition: 0.5s background;
+
+  .MuiStack-root {
+    transition: 0.5s transform;
```
ghost, line 24: love it

src/renderer/theme.tsx
chloebrett: The alternative throws an error lol. Try it

src/renderer/components/NewProjectBlock.tsx
chloebrett: True true

src/renderer/components/NewProjectBlock.tsx
chloebrett: 🌹

src/renderer/theme.tsx
ghost: yeah weird. thanks for the fix

src/renderer/components/RecentProjectsBlock.tsx
ghost: Annoying that it doesn't work as advertised. I understand the reason for changing. Can you also please remove the type declarations for h-300 and h-100 if they no longer exist (if you haven't done so already) :). Thanks

src/renderer/pages/Project.tsx
```diff
@@ -25,17 +25,20 @@ const ProjectPage = () => {
...
-      const project = await window.electron.openProject();
-      dispatch(projectOpened(project));
+      const { project, filePath } = await window.electron.openProject();
+      dispatch(projectOpened(project, filePath));
     } catch (err) {
       console.error(err);
     }
   };
 
+  const handleSaveProject = async () => {
```
ghost: Do we still need these buttons in the project screen? Isn't the logic now integrated into the file menu?

Also, I believe they have been removed in Team Shrek's feature/edit-transcription-interface PR.

src/renderer/pages/Home.tsx
```diff
@@ -1,11 +1,17 @@
...
 const HomePage = () => {
+  const hasRecentProjects = useSelector(
+    (store: ApplicationStore) => store.recentProjects.length > 0
+  );
+
   return (
     <>
-      <NewProjectBlock />
-      <RecentProjectsBlock />
+      <NewProjectBlock isFullSize={!hasRecentProjects} />
```
ghost, line 13: Could the home page just get the entire store of recentProjects from redux and then pass it into the <RecentProjectsBlock/> child component? (rather than accessing the store in both the parent and child component)?

Are you looking to avoid re-renders of the entire RecentProjectsBlock component?

chloebrett: Yeah the home page only uses it to check if there ARE recent projects, so it only reloads when that changes. If we make the selector load the whole recent projects list, it will re render its whole component every time the recent projects change. No practically important but slightly nicer implementation this way

Also kind of defeats the point of redux reducing pass-down props for stuff that is in the store

src/renderer/pages/Project.tsx
chloebrett: Nah we dont I'll remove

src/renderer/pages/Home.tsx
ghost: Fair points. Thanks

src/renderer/pages/Project.tsx
chloebrett: Done

chloebrett:
Ready for re-review, up to date with develop now as well

ghost (approved):

PR #32, opened 2022-05-02
Update `pathToServer` path after refactoring was done

MutazAshhab:
Changed the `pathToServer` in `src/main/pyServer.ts` variable such that it matches the path we currently have after the refactor.


chloebrett (approved):
ah good catch

PR #35, opened 2022-05-02
Feature/project metadata

chloebrett:
~~Depends on #30~~

Why:
- To match UI, we need to load metadata on each project - thumbnails, media file size, project last modified date

Changes:
- Adds a service/handler that does this
- Cleans up IPC initialisation into own folder on back end
- New project schema including unique ID (UUID) for each project (to maintain reference to project even if file name, project name, etc changes)
- Not doing anything with this on the front end yet - see #31 for that


chloebrett:
PR this depends on was merged, ready for review now

src/main/handlers/projectMetadataHandler.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  const mediaSize =
+    project.mediaFilePath === null
+      ? null
+      : await retrieveSize(project.mediaFilePath);
+
+  return {
+    dateModified: projectLastModified,
+    mediaSize,
+  };
```
zwan0202, line 36: Nice work, just a small thing. Think this can be more consistent, like remove `dateModified` or add `mediaSize` to `mediaSize`

zwan0202 (approved):

src/main/handlers/projectMetadataHandler.ts
chloebrett: What do you mean by this? Like consistency between the shorthand and long form? `mediaSize` is just a shorthand for `mediaSize: mediaSize`

itsMoxMox (approved):

src/main/handlers/projectMetadataHandler.ts
itsMoxMox: I think it's fine to leave as is.

zwan0202: Yes, and it's fine  to leave as is

chloebrett: Yeah in general, if a shorthand is available we should use it, I think eslint is set up to get mad if we don't anyway

PR #38, opened 2022-05-03
WIP - feature/new project ui

MutazAshhab:
Frames 102 and 99 from https://www.figma.com/file/vPh49OC7hJpMD7cQMAnBUo/UI-Design?node-id=416%3A2134

Includes:
- Close button on the top of every modal actually closes the modal
- The back button now takes you back a step rather than closing the modal, however, it will not retain the project name.
- after clicking transcribe you'll navigate to the old transcription UI.
- `continue` and `transcribe` are disabled when the relevant data in not yet uploaded/typed in

What has not been included in this PR:
- Delete button functionality (now does nothing when you click it)
- The text color is black (couldnt get it to change to white, not kidding, :( )
- Drag and drop is not implemented.
- The keyboard key `Return` and does not navigate to the next view

### notes for reviewers
- my css skills are whack so if something could be done in a simpler way bring it up.
- i have fallen in love with the `Stack` component and might have overused it. so again, bring it up in cases were we dont need stacks
- in the 2nd screenshot, you can see there is an orange outline, thats only there when i use the keyboard outside of the textfield and when you dont click on it.
- last note. im sorry for the 478 line code review that you'll have to go through

Screenshots:

image

image

image

image


MutazAshhab:
As of [fbd4db85](https://github.com/patrickbrett/mlvet/pull/38/commits/fd4db85c62683af57e51e0ad2c503a9d9232a13d) there are still things that need to be done.

Things to be done:
- [x] Update the height and width of the modal container to match the designs on figma.
- [x] Fixing up the styles (including the white/black coloring of text and small parts of components.)
- [ ] Drag and drop.

chloebrett:
Tested out the UI locally, looking good so far!

src/renderer/components/ActionButton.tsx
```diff
@@ -0,0 +1,9 @@
+import { Button, styled } from '@mui/material';
+import colors from '../colors';
+
+const ActionButton = styled(Button)`
```
emmapainter: This button style should be called PrimaryButton

src/renderer/components/CancelButton.tsx
```diff
@@ -0,0 +1,9 @@
+import { Button, styled } from '@mui/material';
+import colors from '../colors';
+
+const CancelButton = styled(Button)`
```
emmapainter: This should be SecondaryButton

src/renderer/components/ModalTitle.tsx
```diff
@@ -0,0 +1,9 @@
+import { Box, colors, styled } from '@mui/material';
+
+const ModalTitle = styled(Box)`
```
emmapainter: The title should use h1 font style

src/renderer/components/NewProjectModal/ModalContainer.tsx
```diff
@@ -13,9 +15,9 @@ const CustomModal = styled(Modal)`
 `;
 
 const CustomModalInner = styled(Box)`
-  width: calc(80vw - 40px);
-  background: ${colors.grey[900]};
-  padding: 20px;
+  width: calc(30vw - 40px);
```
emmapainter: Not sure on this (@hhen0005 help) but I think the modal should be constant width - currently the modal scales poorly for smaller window sizes.

src/renderer/components/NewProjectModal/NewProjectView.tsx
```diff
@@ -0,0 +1,110 @@
...
+    }
+  };
+
+  const continueButton = (
+    <ActionButton onClick={handleContinue} disabled={isAwaitingProjectName}>
+      Continue
+    </ActionButton>
+  );
+
+  const cancelButton = <CancelButton onClick={closeModal}>Cancel</CancelButton>;
```
emmapainter: Do you know why these button labels are displaying in all caps?

src/renderer/components/NewProjectModal/NewProjectView.tsx
```diff
@@ -0,0 +1,110 @@
...
+          alignItems="center"
+          justifyContent="space-between"
+        >
+          <ModalTitle>New Project</ModalTitle>
+          <Box onClick={closeModal}>
+            <CloseIcon sx={{ color: colors.yellow[500], fontSize: 36 }} />
+          </Box>
+        </CustomStack>
+        <CustomStack>
+          <CustomTextField
```
emmapainter: Figuring out how to change the non-focus and input text colours for this is pretty crucial - the current contrast is really hard to see.

src/renderer/components/NewProjectModal/NewProjectView.tsx
```diff
@@ -0,0 +1,110 @@
...
+            id="standard-basic"
+            label="Project Name"
+            variant="standard"
+            value={projName}
+            onChange={(event) => handleProjectNameInput(event)}
+          />
+        </CustomStack>
+        <CustomStack
```
emmapainter: Would love it if the buttons were equal width and took up the width of the modal like this. Also could you add the drop shadows?
Screen Shot 2022-05-05 at 1 21 58 pm

src/renderer/components/SelectMediaBlock.tsx
```diff
@@ -1,10 +1,17 @@
...
 
 const SelectMediaBox = styled(Box)`
-  background: ${colors.grey[400]};
-  color: ${colors.grey[900]};
+  width: 100%;
+`;
+
+const InnerBox = styled(Box)`
+  border-style: dashed;
+  border-color: ${colors.grey[500]}
```
emmapainter: @hhen0005 / @fwang1703 / someone experienced with front end - the border colour is set to grey 500.... so why does it show up as black in the app?

src/renderer/components/SelectMediaBlock.tsx
```diff
@@ -1,10 +1,17 @@
...
+  color: ${colors.white};
```
emmapainter: I know this colour isn't even showing up as white but the text colour here is grey 300

src/renderer/components/SelectMediaBlock.tsx
```diff
@@ -14,34 +21,45 @@ const SelectMediaBox = styled(Box)`
...
-      </ImportFileBox>
+      <InnerBox>
+        <Stack
+          direction="column"
+          alignItems="center"
+          justifyContent="space-between"
+          sx={{ height: '150px' }}
+        >
+          <Typography fontWeight="bold">Drag and drop file here</Typography>
+          <Typography fontWeight="bold">or</Typography>
```
emmapainter: This text should use p-300 style

src/renderer/components/NewProjectModal/UploadVideoView.tsx
```diff
@@ -0,0 +1,123 @@
...
+          />
+        </CustomStack>
+      </CustomStack>
+      <CustomStack
+        direction="column"
+        alignItems="baseline"
+        justifyContent="flex-start"
+        sx={{ height: '45%' }}
+      >
+        <MediaDisplayOnUpload fileName={mediaFileName} />
```
emmapainter: The location of this is independent of other elements - on smaller screen sizes it glitches out and displays over the top of the select media block
Screen Shot 2022-05-05 at 1 55 29 pm

emmapainter (changes requested):
Added a few UI comments - also not sure if this is expected behaviour, but if you exit out from the select media modal (via the x or clicking outside of modal bounds) then click new project again it will take you back to the select media modal instead of the project name modal. When you're doing UI stuff it's worth testing it at a smaller window size (even just laptop screen size) to make sure everything still looks as expected.

src/renderer/components/NewProjectModal/NewProjectView.tsx
MutazAshhab: Nope, but its probably a setting from the UI library

src/renderer/components/ActionButton.tsx
ghost: We shouldn't need to define primary and secondary button components. 
Because of the custom theme that we are using globally, if we define the color prop of a button as 'primary' or 'secondary' according to the MUI docs, then it should automatically already be styled according to our designs.

Hence, we don't actually need to do any other styling.

ghost: Now realising, theme is currently not working. Please use the method described above for now and I'll fix the theme.

MutazAshhab: I changed the functionality that is executed when the CloseIcon is clicked. It now closes the modal and resets back to the first view.

src/renderer/components/SelectMediaBlock.tsx
chloebrett: MUI really do be like that

src/renderer/components/NewProjectModal/NewProjectView.tsx
chloebrett: Yeah it's a default part of the button styling - they have `text-transform` applied

chloebrett:
Emma has made some good points, I'll wait until you've addressed them before reviewing

src/renderer/components/NewProjectModal/ModalContainer.tsx
MutazAshhab: Resolved in this commit: https://github.com/patrickbrett/mlvet/pull/38/commits/903905ef5e1001fef54cf9b154e015b8835560ac

src/renderer/components/CancelButton.tsx
MutazAshhab: Switched to standard button in: https://github.com/patrickbrett/mlvet/pull/38/commits/664fdb48cc831149ec5dc99c4985f94fff44bf5e

src/renderer/components/SelectMediaBlock.tsx
MutazAshhab: Missing semicolon on line 13, this is now fixed

src/renderer/components/NewProjectModal/UploadVideoView.tsx
MutazAshhab: Responsiveness is now fixed in: https://github.com/patrickbrett/mlvet/pull/38/commits/903905ef5e1001fef54cf9b154e015b8835560ac

src/renderer/components/NewProjectModal/NewProjectView.tsx
emmapainter: Figured this out - in themes.tsx under typography -> button add `textTransform: 'none',`

src/renderer/components/NewProjectModal/UploadVideoView.tsx
```diff
@@ -0,0 +1,132 @@
...
+  const cancelButton = (
+    <StandardButton color="secondary" onClick={prevView}>
+      Back
+    </StandardButton>
+  );
+
+  return (
+    <Container>
+      <CustomStack
+        direction="column"
```
ghost: Nice work defining the custom stack! All 4 of these properties can actually go in the custom stack styled part.
Off the top of my head:
direction = flex-direction
alignItems = align-items
justifyContent = justify-content
height = height

src/renderer/components/NewProjectModal/UploadVideoView.tsx
```diff
@@ -0,0 +1,132 @@
...
+        alignItems="flex-start"
+        justifyContent="space-between"
+        sx={{ height: '50%' }}
+      >
+        <CustomStack
+          direction="row"
```
ghost: same as above :)

src/renderer/components/SelectMediaBlock.tsx
MutazAshhab: Fixed, again, it was the semicolon issue.

src/renderer/components/ModalTitle.tsx
MutazAshhab: now using h-100, done in: https://github.com/patrickbrett/mlvet/pull/38/commits/edcab9ea48b51441c79e70ab36a1c046b144fcaa

src/renderer/components/NewProjectModal/NewProjectView.tsx
MutazAshhab: Done in: https://github.com/patrickbrett/mlvet/pull/38/commits/9f8bba6985635aa8197025abb5cc262102356eb7

src/renderer/components/NewProjectModal/UploadVideoView.tsx
MutazAshhab: done in: https://github.com/patrickbrett/mlvet/pull/38/commits/9c3bb86622a4c497c326922f4841c70ba1e791ac

src/renderer/components/NewProjectModal/UploadVideoView.tsx
MutazAshhab: done in: https://github.com/patrickbrett/mlvet/pull/38/commits/9c3bb86622a4c497c326922f4841c70ba1e791ac

src/renderer/components/SelectMediaBlock.tsx
MutazAshhab: done in: https://github.com/patrickbrett/mlvet/pull/38/commits/724472e1594ea09a26fe28ffdedd877554a9e3ec

src/renderer/components/ActionButton.tsx
MutazAshhab: @hhen0005 update on this? currently im using Standard button and using color="primary"/"secondary". You mentioned something about variants? - sorry i forgot

src/renderer/components/NewProjectModal/NewProjectView.tsx
MutazAshhab: Turns the color attribute in the TextField component does not take actual colors. Check the docs. https://mui.com/material-ui/api/text-field/

Color just changes the outline of the TextField and not the text color that the user types.

src/renderer/components/NewProjectModal/NewProjectView.tsx
MutazAshhab: done in: https://github.com/patrickbrett/mlvet/pull/38/commits/ab49c26e4721cc25c00fd971de2e26f92a3ed5cd

src/renderer/components/NewProjectModal/NewProjectView.tsx
```diff
@@ -0,0 +1,117 @@
...
+    <Button color="secondary" onClick={closeModal} sx={{ width: '40%' }}>
+      Cancel
+    </Button>
+  );
+
+  return (
+    <Container>
+      <CustomColumnStack>
+        <CustomRowStack>
+          <Typography variant="h-100">New Project</Typography>
```
MutazAshhab: @hhen0005  @emmapainter When using the variant option, it gives us the correct size but it does not apply the correct coloring. Should that be applied separately or should it be done via the theme?

MutazAshhab: Answered on zoom, coloring should be applied when Typography is used

src/main/util.ts
```diff
@@ -37,5 +39,17 @@ export const mkdir = (dirPath: string) => {
   }
 };
 
+export const handleOSQuery: () => OperatingSystems = () => {
```
chloebrett: Maybe return an enum instead, as this will always (or at least should, assuming the user is on a supported OS) have exactly one 'true' value

src/renderer/components/NewProjectModal/ModalContainer.tsx
```diff
@@ -34,36 +37,53 @@ const ModalContainer = ({ isOpen, closeModal }: Props) => {
     [dispatch]
   );
 
-  const viewComponents = [ImportMediaView, RunTranscriptionView];
+  const viewComponents = [
+    NewProjectView,
+    UploadVideoView,
```
chloebrett: This was originally called `ImportMediaView`, for two reasons:
- It's not a really an upload, since there's no actual file transfer taking place - hence 'import' 
- It's not guaranteed to be a video file, it could be an audio file - hence 'media'

src/renderer/components/NewProjectModal/NewProjectView.tsx
```diff
@@ -0,0 +1,127 @@
...
+  background-color: ${colors.grey[700]};
+  height: 200px;
+`;
+
+const CustomButton = styled(Button)`
+  filter: drop-shadow(0px 8px 16px rgba(0, 0, 0, 0.8));
+`;
+
+const NewProjectView = ({ closeModal, nextView }: Props) => {
+  const [projName, setProjName] = useState<string>('');
```
chloebrett: In general it's a good idea to avoid abbreviations in variable names unless they're absolutely necessary - this in itself is fine but it's a slippery slope to obtuse names like 'genImpExpStats' that become meaningless (trust me I've seen it happen 😅)

src/renderer/pages/Project.tsx
```diff
@@ -1,4 +1,4 @@
-import { Stack } from '@mui/material';
+import { Button, Stack } from '@mui/material';
```
chloebrett: IIRC the 'StandardButton' thing you deleted was made by Hugh as a placeholder for when we want to bring in custom styling. I'm fine with replacing with a regular MUI button for now as we can style it in the theme file instead (and currently have no custom styles for it), but maybe check with @hhen0005 and see what he thinks (that said I don't think this is a blocker for the merge)

src/renderer/pages/Project.tsx
```diff
@@ -84,24 +83,21 @@ const ProjectPage = () => {
...
+            </Button>
+            <Button onClick={changeRandomWordToSwamp}>
               Change a random word to &lsquo;swamp&rsquo;
-            </StandardButton>
-            <StandardButton
-              onClick={dispatchUndo}
-              disabled={undoStack.index <= 0}
-            >
+            </Button>
+            <Button onClick={dispatchUndo} disabled={undoStack.index <= 0}>
```
patrickbrett, line 92: Again not a blocker, but if you have a minute can you remove the undo/redo buttons please as their functionality exists in the menu now (I know this is technically not part of your PR but just noticed it)

src/renderer/util.ts
```diff
@@ -19,6 +19,32 @@ export const extractFileExtension: (filePath: string) => string | null = (
   return extension;
 };
 
+export const extractFileNameWithExtension: (
```
chloebrett, line 23: I think this should be a backend function, that way the renderer doesn't need to care about OS stuff. We might need a getOS() down the line but we're trying to make the platforms have feature parity so I can't foresee anything specific

src/renderer/util.ts
```diff
@@ -76,6 +102,60 @@ export const makeProject: (
...
+    transcription: null,
+    mediaType: 'audio',
+    mediaFileExtension: 'mp3',
+    thumbnailFilePath: null,
+  };
+
+  return project;
+};
+
+export const updateProjectWithMedia: (
```
chloebrett, line 128: Nice way of splitting it!

src/sharedTypes.ts
```diff
@@ -33,3 +33,9 @@ export interface Word {
 }
 
 export type MediaFileExtension = AudioFileExtension | VideoFileExtension;
+
+export interface OperatingSystems {
```
chloebrett: Same idea, if we do this on the back end then no shared types needed (also again, would make this an enum)

chloebrett (changes requested):
Styling is looking really good! Just a few back end comments

src/renderer/pages/Project.tsx
emmapainter: Can confirm we will be removing the StandardButton

emmapainter:
There is extra space at the top and bottom of the new project modal and at the top of the upload media modal - otherwise the styling is looking great! Sorry for being so picky 😅
image
On Figma if you select an element then select "inspect" on the top right then it will show you the pixel distance between the item you have selected and any item you hover over with your mouse
Screen Shot 2022-05-07 at 4 27 24 pm

MutazAshhab:
Here is what it looks like now: 

image

corresponding commit: https://github.com/patrickbrett/mlvet/pull/38/commits/0f0ae23e5d6d70f8488f0a84e7c266b344c1d756

src/renderer/components/NewProjectModal/ModalContainer.tsx
MutazAshhab: refactor done in: https://github.com/patrickbrett/mlvet/pull/38/commits/40010a865dd4a23174225bea93e2d2876e78655d

src/renderer/util.ts
MutazAshhab: thank you!

src/renderer/components/NewProjectModal/NewProjectView.tsx
MutazAshhab: Ah yes, this is something i overlooked before pushing, there is no need for an abbreviated name here. I refactored the name here: https://github.com/patrickbrett/mlvet/pull/38/commits/5545d5ec5937464dc3b775d37bfe19d4e5b354a6

Although its fine to have an abbreviated variable name, changing this just so others dont see an abbreviated variable name and start getting ideas.

src/renderer/pages/Project.tsx
MutazAshhab: I have already removed all occurrences of StandardButton in this branch. Once it gets merged, `develop` will no longer have StandardButton and will only use Button.

src/sharedTypes.ts
MutazAshhab: enum refactor here: https://github.com/patrickbrett/mlvet/pull/38/commits/f54cbe3641987ce328de2f5e4091eb81a6eb50d4

src/renderer/util.ts
MutazAshhab: enum refactor here: https://github.com/patrickbrett/mlvet/pull/38/commits/f54cbe3641987ce328de2f5e4091eb81a6eb50d4

I chose to keep getOS since we might need it down the line

src/main/util.ts
MutazAshhab: nice catch

enum refactor: https://github.com/patrickbrett/mlvet/pull/38/commits/f54cbe3641987ce328de2f5e4091eb81a6eb50d4

emmapainter:
One last thing (sorry) before I approve! Could you line up the right edge of the x button with the right side of the other elements on the modal for both pages?
Screen Shot 2022-05-07 at 7 27 31 pm

MutazAshhab:
When i remove the right padding it inches closer to where the line you drew is. 

This is the closest it goes since the icon itself has empty space around it.

image

image

Is this acceptable?

emmapainter:
Oh wow I forgot that the icon had padding inside it - I hate that but it's MUI's fault so I guess it's too bad. Love your work, approving now!

emmapainter (approved):

src/renderer/components/NewProjectModal/ImportMediaView.tsx
```diff
@@ -1,116 +1,136 @@
...
+            setMediaFileName={setMediaFileName}
+            setMediaFilePath={setMediaFilePath}
+            setIsAwaitingMedia={setIsAwaitingMedia}
+          />
+        </CustomRowStack>
+      </CustomColumnStack>
+      <CustomColumnStack
+        alignItems="baseline"
+        justifyContent="flex-start"
+        sx={{ height: '45%' }}
```
ghost: Love the stacks. Flexbox convert. Looks like this stack (which displays the uploaded media) is a little bit big and its pushing the back/transcribe buttons down. Maybe reduce the height a bit. 

image

Being aware of multiple files in the future or even just ultra-long filenames now, could be a good idea to add overflow-y: 'auto' (overflowY ?) to this stack. This will add a scrollbar when the content overflows out of the designated height for the container element.

src/renderer/components/NewProjectModal/ImportMediaView.tsx
```diff
@@ -1,116 +1,136 @@
...
+      onClick={handleTranscribe}
+      disabled={isAwaitingMedia}
+      sx={{ width: '40%' }}
+    >
+      Transcribe
+    </CustomButton>
+  );
+
+  const cancelButton = (
+    <CustomButton color="secondary" onClick={prevView} sx={{ width: '40%' }}>
```
ghost: Rather than set width 40% on both the cancel button and the transcribe button. I suggest you set the width of both to be 100% and then set the gap property (CSS) of the container element to be 32px.

This means that both buttons will try and take up as much space in their parent component as possible (so will be equal) and the gap styling will force the 32px between them.

MutazAshhab: Done in this commit: https://github.com/patrickbrett/mlvet/pull/38/commits/db258538dce61ed27b415f8350b7af96a08cf2f0

I also went ahead and adjusted it for `NewProjectView.tsx`

src/renderer/components/NewProjectModal/ImportMediaView.tsx
MutazAshhab: Nice catch, I went ahead and adjusted the hieght such that it lines up. I also removed `marign-top` in other components.

Most importantly i also added the `overflowY: 'auto'` in this commit: https://github.com/patrickbrett/mlvet/pull/38/commits/eaf8056a93593b9f6945a1618fecd871dbe21670

src/renderer/pages/Project.tsx
ghost: Great!

ghost (approved):

src/main/util.ts
```diff
@@ -37,5 +39,23 @@ export const mkdir = (dirPath: string) => {
   }
 };
 
+export const handleOSQuery: () => OperatingSystems | null = () => {
+  const isDarwin = os.platform() === 'darwin';
+  const isWindows = os.platform() === 'win32';
+  const isLinux = os.platform() === 'linux';
+
+  if (isDarwin) {
```
chloebrett, line 47: Nitpick - could just map OS identifier strings to enum values and return that, with null as a default fallback

src/renderer/components/NewProjectModal/RunTranscriptionView.tsx
```diff
@@ -70,6 +70,11 @@ const RunTranscriptionView = ({ closeModal, nextView }: Props) => {
     ) {
       return;
     }
+
+    if (currentProject.filePath === null) {
```
chloebrett: I don't think 'filePath' is an attribute in 'project' any more - we have 'mediaFilePath' and 'saveFilePath'

src/sharedTypes.ts
```diff
@@ -42,3 +42,9 @@ export interface Clip {
 }
 
 export type MediaFileExtension = AudioFileExtension | VideoFileExtension;
+
+export enum OperatingSystems {
```
chloebrett, line 46: This will store the values as numbers, can we instead do like Linux = 'LINUX' etc so the runtime values are human readable

Also, maybe say MacOS instead of Darwin, Darwin is just the string identifier and not many people are familiar with it

chloebrett (approved):
Couple of super small things, approved once those are done

src/renderer/components/NewProjectModal/RunTranscriptionView.tsx
MutazAshhab: refactor done here: https://github.com/patrickbrett/mlvet/pull/38/commits/5d57bfca861469a5898825e6172929ba38319e0c

src/sharedTypes.ts
MutazAshhab: human readable refactor + macOS instead of darwin done here: https://github.com/patrickbrett/mlvet/pull/38/commits/d0e83b46c50d1224dbed8e815b067fd19d3ae308

src/main/util.ts
MutazAshhab: Mapped enum valies with the corresponding OS identifiers here: https://github.com/patrickbrett/mlvet/pull/38/commits/d0e83b46c50d1224dbed8e815b067fd19d3ae308

MutazAshhab:
I also refactored `MediaDisplayOnUpload` component and imports to `MediaDisplayOnImport` since @patrickbrett pointed out that the user will be importing not uploading.

src/main/util.ts
```diff
@@ -37,5 +39,23 @@ export const mkdir = (dirPath: string) => {
...
+  const isWindows = os.platform() === OperatingSystems.WINDOWS;
+  const isLinux = os.platform() === OperatingSystems.LINUX;
+
+  if (isDarwin) {
+    return OperatingSystems.MACOS;
+  }
+  if (isWindows) {
+    return OperatingSystems.WINDOWS;
+  }
+  if (isLinux) {
```
chloebrett, line 53: This isn't quite what I meant, I was more referring to using a map/object so you don't need three if statements, something like this:

```
const platformToOperatingSystem: Record<string, OperatingSystems> = {
    'darwin': OperatingSystems.MACOS,
    'windows': OperatingSystems.WINDOWS,
    'linux': OperatingSystems.LINUX
};

return platformToOperatingSystem[os.platform()] || null;
```

But - don't worry about it, this is approach is fine. There's a fine line between giving a review and saying "do it how I'd do it". There is nothing wrong with the approach you've taken given the number of options, just consider the object approach if you encounter this problem again with say, 5-10 options as it will cut down on if statements :)

chloebrett (approved):
Looking gooood! Sorry for holding up your merge for so long hehe

PR #39, opened 2022-05-03
Feature/edit delete

LiamTodd:
Put a call to preProcess in handleTranscription, so that when the json is read from the ML model it is processed before it is put into the store. Front end won't know about this.

Changed the sharedType 'Word' to include a 'key' field of type 'string' and a 'deleted' field of type 'boolean'.
Used concrete implementations of Word ('TranscriptWord') and Transcription ('JSONTranscription') shared types process the json data. These class names are pretty bad, forgive us pls.

Updated postProcess to be compatible with our updated types, but it still isn't being used.

We currently don't have access to the duration of the video file (we need this for pre processing), so its hardcoded to '0'.

### DIFF
--- a/src/main/edit_delete/postProcess.ts
+++ b/src/main/edit_delete/postProcess.ts
@@ -1,3 +1,4 @@
+import { Transcription } from 'sharedTypes';
 import Clip from './Clip';
 
 /**
@@ -22,7 +23,9 @@ const removeDeleted = (wordList: Array<any>): void => {
  */
 const FILENAME = 'placeholder_filename.mp4';
 
-const postProcess = (wordList: Array<any>): Array<Clip> => {
+const postProcess = (jsonTranscript: Transcription): Array<Clip> => {
+  const wordList = jsonTranscript.words;
+
   removeDeleted(wordList);
   const res: Array<Clip> = [];
   let currentStartTime = -1;
@@ -30,13 +33,13 @@ const postProcess = (wordList: Array<any>): Array<Clip> => {
 
   for (let i = 0; i < wordList.length - 1; i += 1) {
     if (currentStartTime === -1) {
-      currentStartTime = wordList[i].start_time;
+      currentStartTime = wordList[i].startTime;
       currentDuration = wordList[i].duration;
     }
 
     if (
-      wordList[i].start_time + wordList[i].duration ===
-      wordList[i + 1].start_time
+      wordList[i].startTime + wordList[i].duration ===
+      wordList[i + 1].startTime
     ) {
       currentDuration += wordList[i + 1].duration;
     } else {
@@ -47,10 +50,10 @@ const postProcess = (wordList: Array<any>): Array<Clip> => {
   }
 
   if (res.length === 0) {
-    currentStartTime = wordList[0].start_time;
+    currentStartTime = wordList[0].startTime;
     res.push(new Clip(currentStartTime, currentDuration, FILENAME));
   } else if (currentStartTime === -1) {
-    currentStartTime = wordList[wordList.length - 1].start_time;
+    currentStartTime = wordList[wordList.length - 1].startTime;
     currentDuration = wordList[wordList.length - 1].duration;
     res.push(new Clip(currentStartTime, currentDuration, FILENAME));
   } else {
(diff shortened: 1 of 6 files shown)


jhea0008 (approved):

src/main/edit_delete/preProcess.ts
```diff
@@ -1,11 +1,47 @@
...
+
+  const processedTranscript: Transcription = new JSONTranscription(
+    [],
+    jsonTranscript.confidence
+  );
+
+  for (let i = 0; i < numberOfWords - 1; i += 1) {
+    // duration includes the white space between current and next word
+    const wordDuration =
+      jsonTranscript[i + 1].start_time - jsonTranscript.words[i].start_time;
```
rileykeane: `.words` is missing here after `jsonTranscript` which is causing the function to crash 

should be `jsonTranscript.words[i + 1]`

src/main/edit_delete/postProcess.ts
```diff
@@ -22,21 +23,23 @@ const removeDeleted = (wordList: Array<any>): void => {
  */
 const FILENAME = 'placeholder_filename.mp4';
 
-const postProcess = (wordList: Array<any>): Array<Clip> => {
```
rileykeane, line 25: What exactly are we passing here? Is it just the the array of from a Transcription? e.g `Word[]`

src/main/edit_delete/preProcess.ts
```diff
@@ -1,11 +1,47 @@
-const preProcess = (wordList: Array<any>, duration: number): Array<any> => {
```
rileykeane, line 1: All instances of start_time needs to be changed to startTime

src/main/edit_delete/preProcess.ts
```diff
@@ -1,11 +1,47 @@
...
-  for (let i = 0; i < wordList.length - 1; i += 1) {
-    wordList[i].duration = wordList[i + 1].start_time - wordList[i].start_time;
-    wordList[i].deleted = false;
+import { Transcription } from 'sharedTypes';
+import JSONTranscription from './JSONTranscription';
+import TranscriptWord from './TranscriptWord';
+
+const preProcessTranscript = (
+  jsonTranscript: any,
```
rileykeane, line 6: should `jsonTranscript` be of type `Transcript`?

rileykeane (changes requested):
There are a few bugs here which are preventing the code from running. 

1. all cases of `start_time` need to be changed to `startTime`. `postProcess` seems to be fixed by `preProcess` still has the old notation. 
2. fix the bug in `preProcess` on line 19
3. Avoid using `any` for function parameter types.

src/main/edit_delete/preProcess.ts
LiamTodd: Transcript and Word are interfaces to represent the transcription, we made JSONTranscription and TranscriptWord as concrete implementations of these interfaces.

src/main/edit_delete/preProcess.ts
LiamTodd: should be fixed now

src/main/edit_delete/preProcess.ts
LiamTodd: Don't think so, as the input is raw JSON (from the ML model) and is given as an "any" type in transcriptionHandler. This function processes the raw json and spits out a Transcript instance to be put into store.

LiamTodd:
1. The raw json is the input to preProcess(), and its contents take the form {"word": string, "start_time": number, "duration": number}. We process these to create Word instances which have the fields "word":string, "startTime": number, "duration": number, "key": string, and "deleted": bool. Where you're seeing 'start_time' is when we are accessing the raw json, to access the start time data. With this data, we create Word instances which will have their 'startTime' field assigned the 'start_time' value from the corresponding json. Hope this makes sense. If this code is causing errors please let me know!

3. Good pick up, it's been fixed
4. The input is raw json, how do we specify this type?

src/main/edit_delete/preProcess.ts
rileykeane: sorry my bad! You're right - I forgot you are taking the raw json from the ml team - not a Transcript object. This is fine 🙂

rileykeane (approved):
The `startTime` is undefined in the returned `Transcript` but its not a blocking issue - can fix this after the merge.

PR #40, opened 2022-05-03
Feature/export fe

ghost:
Need an export UI and the capability to show the export progression on the frontend. 

Added the UI (a small card with a progress bar), a redux store for IO related states and the IPC channels for the backend to update on the frontend on the progress of the export.

image

*One small section of unnecessary code to be removed as part of a later PR. Will comment to indicate it and reason for it.


src/main/export.ts
```diff
@@ -33,14 +34,22 @@ const constructEDL: (title: string, transcription: Transcription) => string = (
...
   mkdir(project.savePath);
 
   if (project.transcription) {
     writeFileSync(
       join(project.savePath, `${project.name}.edl`),
       constructEDL(project.name, project.transcription)
     );
+
+    // Placeholder for BE - to be copied and removed
+    mainWindow?.webContents.send('export-progress-update', 0.5);
```
ghost, line 54: This line & comment to be removed in a later PR. Probably by @itsMoxMox 

Included so that backend team can easily copy the functionality into the export logic so that export progress can be updated at intervals throughout.

itsMoxMox: Thanks for the code nugget!

itsMoxMox (approved):
Looks good!

src/renderer/store/helpers.ts
```diff
@@ -33,6 +33,15 @@ export interface UndoStack {
   index: number; // Used for supporting redo
 }
 
+/**
+ * All input/output user actions states
+ * Import / Export progress states
+ */
+export interface IO {
```
chloebrett: Maybe call this ExportIO or similar as we may want other IO state later

chloebrett (changes requested):
My understanding of exportEDL is that it's basically synchronous, right?

Like we take the transcription state, shuffle some json around and basically construct an EDL in linear time, where each word we're processing takes maybe 20 microseconds or something. Which means for even say a 1000 word transcript, there's hardly any point showing a progress bar to the user since it'll be done in a fraction of a second.

Basically: just want to confirm, for EDL export, no ffmpeg or similar is required, right? So this is basically instant (or we can at least treat it like it is)

Am I right in saying that?

If so then is this necessary?

That said, even if its not, I love the building blocks you've made, I'm sure they will be relevant for reusing in other long running processes - like mp4 export down the line, and maybe even for the transcription progress.

src/renderer/store/helpers.ts
ghost: Yep, sure. I actually named it IO to allow for future IO states. But I guess we can always rename it back if we decide to go with that.

ghost: Done.

chloebrett: Ah I see what you mean - I had imagined it as we would add new root level state for other IO stuff, rather than reusing the IO state. I suppose either are fine

src/renderer/store/reducers/index.ts
```diff
@@ -9,6 +10,7 @@ const rootReducer = combineReducers({
   recentProjects: recentProjectsReducer,
   currentPage: currentPageReducer,
   undoStack: undoStackReducer,
+  io: exportIoReducer,
```
chloebrett: You need to change this, otherwise the whole thing throws an error.

chloebrett (commented):
Approved once you fix this one thing >

Also, am I correct in saying there's currently no way to test this UI locally?

src/renderer/store/reducers/index.ts
ghost: Whoops. Fixed now. Thanks for pointing that out.

ghost: In relation to above comment regarding local testing of UI:

Not really. Another PR is coming with the export file menu option which will trigger the entire process.
If you want to see the export card you can edit the conditional rendering to be `true`.

chloebrett (approved):

PR #41, opened 2022-05-04
Feature/edit transcription interface

rileykeane:
Sorry @hhen0005 this is quite a large PR...

**Why is this change needed?** 
This is needed to display the transcript on the project page so the user can edit the video. Also includes functionality to delete section of the transcript.  

**What did you change?**
- Added backend logic to pre process the transcription model and add fields to the object that are required for rendering the transcript in the front end and for generating the edl 
- Added backend logic to post processes the transcript and get it ready for export
- Calls the preprocess function before adding the transcription to store 
- Added a TranscriptionBlock that takes a Transcript object and renders it as text 
- Added functionality to delete text from the transcript and update the transcript in store 
- Cleaned up the Project page UI (note this is still not final) 
- Changed SampleTranscript.json back to snake_case to be inline with the correct output from the ML model.


fwang1703 (approved):

src/main/edit_delete/postProcess.ts
```diff
@@ -0,0 +1,66 @@
+import { Transcription } from 'sharedTypes';
+import Clip from './Clip';
+
+/**
+ * Removes deleted words from edited trancription
+ * @param wordList
+ */
+const removeDeleted = (wordList: Array<any>): void => {
+  let i = 0;
+  while (i < wordList.length) {
```
chloebrett: Can do this immutably like so:

`return wordList.filter(word => !word.deleted)`

Bit easier to follow. Also will be O(n) instead of O(n^2)

src/main/edit_delete/postProcess.ts
```diff
@@ -0,0 +1,66 @@
...
+ * Processes edited data to be ready for export
+ * @param wordList
+ * @returns
+ */
+const FILENAME = 'placeholder_filename.mp4';
+
+const postProcess = (jsonTranscript: Transcription): Array<Clip> => {
+  const wordList = jsonTranscript.words;
+
+  removeDeleted(wordList);
```
chloebrett: if you do it immutably you will then need to reassign wordList - probably easies to just do it in the initial assignment

`const wordList = removeDeleted(jsonTranscript.words)`

src/main/edit_delete/Clip.ts
```diff
@@ -0,0 +1,16 @@
+export default class Clip {
```
chloebrett: Is this class going to have any specific methods attached to it? If not, can it just be an object with an interface instead? (better for logging, passing around, serialising to a string, reduces boilerplate, etc)

src/main/edit_delete/JSONTranscription.ts
```diff
@@ -0,0 +1,12 @@
+import { Transcription, Word } from 'sharedTypes';
+
+export default class JSONTranscription implements Transcription {
```
chloebrett: Same idea, in this case instead of calling `new JSONTranscription` we can just make an object like `({ confidence, words })` and pass that around, unless there's something specific this class needs to do (and even then could just be simpler to have a top level function that takes a Transcription as argument and returns whatever is needed, simpler and purer that way)

src/main/edit_delete/TranscriptWord.ts
```diff
@@ -0,0 +1,34 @@
+import { Word } from 'sharedTypes';
+
+export default class TranscriptWord implements Word {
```
chloebrett: I think you know what I'm going to say here

src/main/edit_delete/postProcess.ts
```diff
@@ -0,0 +1,66 @@
...
+  const res: Array<Clip> = [];
+  let currentStartTime = -1;
+  let currentDuration = 0;
+
+  for (let i = 0; i < wordList.length - 1; i += 1) {
```
chloebrett: This can be simplified too, though the actual implementation might be a bit tricky so I'm happy if you want to leave this for now. Just in general, outside of some very specific circumstances (like if you're implementing KMP or something lol) you generally don't need to use explicit for loops like this, `forEach` or `map` are a bit cleaner and easier to read

Think of the steps you could break this down into - e.g. firstly adding dummy words at the start and end to represent the max boundaries, then pairing each word with its successor, then combining words that overlap each others' boundaries, or something like that.

src/main/edit_delete/postProcess.ts
```diff
@@ -0,0 +1,66 @@
...
+    ) {
+      currentDuration += wordList[i + 1].duration;
+    } else {
+      res.push(new Clip(currentStartTime, currentDuration, FILENAME));
+      currentStartTime = -1;
+      currentDuration = 0;
+    }
+  }
+
+  if (res.length === 0) {
```
chloebrett: That would also reduce the need for these cleanups which have a bit of code repetition

src/main/edit_delete/preProcess.ts
```diff
@@ -0,0 +1,49 @@
...
+      i.toString(),
+      jsonTranscript.words[i].word,
+      jsonTranscript.words[i].start_time,
+      wordDuration,
+      'PLACEHOLDER FILENAME'
+    );
+    processedTranscript.words.push(word);
+  }
+
+  // last word in transcript
```
chloebrett: Same idea here regarding cleanups, this is very similar to the code above, see if you can DRY this out

src/main/handlers/transcriptionHandler.ts
```diff
@@ -19,16 +20,24 @@ const handleTranscription: (
...
+  const duration = 0; // TODO: get actual duration from video
+  const processedTranscript = preProcessTranscript(
+    jsonTranscript.transcripts[0],
+    duration
+  );
 
-  return newTranscript;
+  return processedTranscript;
+
+  // const newTranscript: Transcription = processedTranscript;
```
chloebrett: remove

src/renderer/components/TranscriptionBlock.tsx
```diff
@@ -0,0 +1,47 @@
...
+  onWordClick: (wordIndex: number) => void;
+}
+
+const TranscriptionBlock = ({ onWordClick, transcription }: Props) => {
+  return (
+    <TranscriptionBox>
+      <p>
+        {/* eslint-disable-next-line prettier/prettier */}
+        {transcription.words.map((word, index) => (
+          !word.deleted ? (
```
chloebrett: Flip this ternary so the else isn't a double negative - or better yet, just filter the deleted words before mapping

src/renderer/pages/Project.tsx
```diff
@@ -1,28 +1,80 @@
...
+  const saveTranscription: (transcription: Transcription) => void = useCallback(
+    (transcription) => dispatch(transcriptionCreated(transcription)),
+    [dispatch]
+  );
+
+  const deleteWord = (firstWordIndex: number, numberOfWords: number) => {
+    if (currentProject && currentProject.transcription) {
+      // eslint-disable-next-line no-plusplus
+      for (let i = firstWordIndex; i < firstWordIndex + numberOfWords; i++) {
+        currentProject.transcription.words[i].deleted = true;
```
chloebrett, line 42: If you want these actions to be undoable, then dispatch them as ops with 'do' and 'undo' actions (see: [undo stack doc](https://docs.google.com/document/d/1fBLBj_I3Y4AgRnIHzJ-grsXvzoKUBA03KNRv3DzABAg/edit), [video if you're up for it](https://www.youtube.com/watch?v=D9laah5bcio)

Especially don't mutate state - this isn't how redux (or react in general) is supposed to work

Happy to run you through this if you like

src/renderer/pages/Project.tsx
```diff
@@ -1,28 +1,80 @@
...
+section to the side among other things.
+*/
 const ProjectPage = () => {
+  const dispatch = useDispatch();
   const currentProject = useSelector(
     (store: ApplicationStore) => store.currentProject
   );
 
-  const dispatch = useDispatch();
+  // RK: I really shouldn't use transcriptionCreated here - but i'm lazy and it works
```
chloebrett, line 32: Yeah, it works for now but it will start doing funky stuff if we change it (e.g. for the original transcription to be backed up into a different part of the state, which is likely at some point).

easy solution - make a new action that does the same thing for now, just call it `transcriptionEdited` or similar

harder (but much better) solution - refactor to use ops so the undo works and so you're not mutating the application state

src/renderer/pages/Project.tsx
```diff
@@ -1,28 +1,80 @@
...
+  const saveTranscription: (transcription: Transcription) => void = useCallback(
+    (transcription) => dispatch(transcriptionCreated(transcription)),
+    [dispatch]
+  );
+
+  const deleteWord = (firstWordIndex: number, numberOfWords: number) => {
+    if (currentProject && currentProject.transcription) {
+      // eslint-disable-next-line no-plusplus
+      for (let i = firstWordIndex; i < firstWordIndex + numberOfWords; i++) {
```
chloebrett, line 41: Not that you should be doing this anyway, but for reference it can be done as a one liner: `currentProject.transcription.words.slice(firstWordIndex, numberOfWords).forEach(word => word.deleted = true);`

src/renderer/pages/Project.tsx
```diff
@@ -1,28 +1,80 @@
...
+        currentProject.transcription.words[i].deleted = true;
+      }
+      saveTranscription(currentProject.transcription);
+    }
+  };
 
-  if (currentProject === null) {
+  const deleteText = async () => {
+    const highlightedWords = window.getSelection();
```
chloebrett, line 49: Using the native selection seems fine for now, might need to build our own at some point though

src/renderer/pages/Project.tsx
```diff
@@ -1,28 +1,80 @@
...
+      const anchor = await Number(
+        highlightedWords?.anchorNode?.parentElement?.dataset.index
+      );
+      const focus = await Number(
+        highlightedWords?.focusNode?.parentElement?.dataset.index
+      );
+
+      const start = Math.min(anchor, focus);
+      const end = Math.max(anchor, focus);
+      deleteWord(start, end - start + 1);
```
chloebrett, line 63: If this is how you're using deleteWord, why not just make it take `start` and `end`?

src/renderer/pages/Project.tsx
```diff
@@ -1,28 +1,80 @@
...
+    if (event.key === 'Delete' || event.key === 'Backspace') {
+      deleteText();
+    }
+  };
+
+  useEffect(() => {
+    window.addEventListener('keydown', onKeyDown);
+
+    return () => {
+      window.removeEventListener('keydown', onKeyDown);
```
chloebrett, line 77: very nice, someone's been watching react tuts 😛

src/renderer/pages/Project.tsx
```diff
@@ -1,28 +1,80 @@
...
 
-  const handleOpenProject = async () => {
-    try {
-      const project = await window.electron.openProject();
-      dispatch(projectOpened(project));
-    } catch (err) {
-      console.error(err);
-    }
+  const onWordClick = (wordIndex: number) => {
+    console.log(currentProject.transcription?.words[wordIndex]);
```
chloebrett: remove if this is just for debug

src/renderer/store/helpers.ts
```diff
@@ -22,6 +22,7 @@ const baseMockProject: Omit<Project, 'name'> = {
   filePath: 'fakepath',
   fileExtension: 'mp4',
   transcription: null,
+  savePath: '',
```
chloebrett: what is savePath referring to? There was a `savePath` we had before that referred to the export path, its name has been changed to `exportFilePath` I believe.

src/sharedTypes.ts
```diff
@@ -17,6 +17,8 @@ export interface Word {
   word: string;
   startTime: number;
   duration: number;
+  deleted: boolean;
```
chloebrett, line 33: Why do we need to soft-delete words? Why can't we just remove them entirely?

src/main/edit_delete/preProcess.ts
```diff
@@ -0,0 +1,49 @@
...
+  duration: number
+): Transcription => {
+  const numberOfWords: number = jsonTranscript.words.length;
+
+  const processedTranscript: Transcription = new JSONTranscription(
+    [],
+    jsonTranscript.confidence
+  );
+
+  for (let i = 0; i < numberOfWords - 1; i += 1) {
```
chloebrett: This one actually returns a word for each iteration so you can change it to a `map`

src/main/handlers/transcriptionHandler.ts
```diff
@@ -2,6 +2,7 @@ import path from 'path';
 import fs from 'fs';
 import { app } from 'electron';
 import { Transcription } from '../../sharedTypes';
+import preProcessTranscript from '../edit_delete/preProcess';
```
chloebrett: Nitpick, but we're using camelCase for file and folder names, not snake_case

src/renderer/components/TranscriptionBlock.tsx
```diff
@@ -0,0 +1,47 @@
...
+              data-index={index}
+              data-type="word"
+              onClick={() => {
+                onWordClick(index);
+              }}
+            >
+              {`${word.word} `}
+            </Word>
+          ) : null
+          /* eslint-disable-next-line prettier/prettier */
```
chloebrett: What does this need to be disabled for?

src/main/edit_delete/postProcess.ts
```diff
@@ -0,0 +1,66 @@
...
+ * @param wordList
+ * @returns
+ */
+const FILENAME = 'placeholder_filename.mp4';
+
+const postProcess = (jsonTranscript: Transcription): Array<Clip> => {
+  const wordList = jsonTranscript.words;
+
+  removeDeleted(wordList);
+  const res: Array<Clip> = [];
```
chloebrett: `Clip[]`

src/main/edit_delete/preProcess.ts
```diff
@@ -0,0 +1,49 @@
...
+    // duration includes the white space between current and next word
+
+    const wordDuration =
+      jsonTranscript.words[i + 1].start_time -
+      jsonTranscript.words[i].start_time;
+    // unique identifier for each word
+    jsonTranscript.words[i].key = i.toString();
```
chloebrett: This is a pretty good way to go about it, nice

src/renderer/components/TranscriptionBlock.tsx
```diff
@@ -0,0 +1,47 @@
...
+  transcription: Transcription;
+  onWordClick: (wordIndex: number) => void;
+}
+
+const TranscriptionBlock = ({ onWordClick, transcription }: Props) => {
+  return (
+    <TranscriptionBox>
+      <p>
+        {/* eslint-disable-next-line prettier/prettier */}
+        {transcription.words.map((word, index) => (
```
chloebrett: Nice work using the map. Instead of logging each word with a space next to it, maybe `join` the map result with a space. Easier to follow (I was confused for a sec about why you had the backticks) and also won't put a space at the end of the last word

chloebrett (changes requested):
Nice work overall, left a few comments to simplify some things. Happy to run through them with you if you like.

Also, for future, might be better to gradually merge things to develop instead of trying to get 500 lines in one go. It looks like there's about 5 people who've contributed to this over several weeks - would be easier to keep things in sync with the rest of the project if it was integrated more frequently, and also easier to review. Wasn't really a problem this time, just keep in mind

src/renderer/components/TranscriptionBlock.tsx
rileykeane: @fwang1703 @emmapainter

src/sharedTypes.ts
rileykeane: Let's say I deleted a section then made 4 edits after that. What if I want to restore the section I deleted 5 edits ago without having to revert all the way back to the state at that time? This way we at least have the option to restore deleted sections without needing to undo every change that came before it. 

It hasn't been implemented because the UI for the restore action is tricky - Zac doesn't want deleted text to have a strikethrough (which would be the easy way) so it needs some brainstorming before its implemented.

src/renderer/components/TranscriptionBlock.tsx
rileykeane: Thats a smart idea and will fix my hover bug showing the space. Thanks!

src/renderer/store/helpers.ts
rileykeane: We had hacked around with helpers to make testing easier - I reverted back to an older version before raising the PR. This whole file can just be overwritten with whatever is on develop.

src/sharedTypes.ts
chloebrett: Makes sense, I can see the value in this.

Are these requirements documented somewhere? I had never heard about it

rileykeane: We had strikethrough in the original designs, Zac asked us to scrap it. I didn't want to fully remove the possibility - I'm going to try and convince him otherwise.

src/main/edit_delete/postProcess.ts
LiamTodd: Thanks for the tip! I ended up deleting 'removeDeleted' and just used your one-liner.

src/main/edit_delete/postProcess.ts
LiamTodd: Makes sense, thanks. I've done it as
 "const wordList = jsonTranscript.words.filter(word => !deleted)".

Am I correct in saying that this wouldn't work:
const wordList = jsonTranscript.words
const wordList = wordList.filter(word => !deleted)
?

src/renderer/components/TranscriptionBlock.tsx
emmapainter: Looks like the parentheses it was complaining about can actually be removed - for some reason before when we tried removing them it kept complaining, it might have just been super laggy to update though.

src/renderer/pages/Project.tsx
rileykeane: Will need to go with the harder option for undo anyway. Ok to just leave it as is for now?

src/main/edit_delete/Clip.ts
LiamTodd: I get what you mean - Clip doesn't really need to be a class. Sorry but I'm not too sure how to implement the change though, my JS and TS is worse than my Japanese, and I don't speak Japanese. 
Applies for the next two comments as well regarding JSONTranscript and TranscriptWord.

src/renderer/components/TranscriptionBlock.tsx
emmapainter: @rileykeane are you changing it to filter before mapping?

src/main/edit_delete/postProcess.ts
chloebrett: That would throw a syntax error because you'd be reassigning a const. You could use let, but in general it's preferable to avoid let where possible, so either inlining or giving the vars different names is the usual approach

src/main/edit_delete/Clip.ts
chloebrett: Hahah, that's unfortunate - my Japanese isn't too good either 😢

Basically just instead of calling `new Clip`, create an object directly with the fields it needs - e.g. `{ startTime: 123, duration: 456, fileName: 'bla' }` and that should be the only change you need (accessing fields is the same as with a class)

If you can't get it working, don't worry too much, it's not a blocker.

src/renderer/pages/Project.tsx
chloebrett: Hahah yeah leave it as is, I'd say merge this when the other minor stuff is resolved then make a new PR to switch to using the op system

src/renderer/pages/Project.tsx
emmapainter: @rileykeane since you're the PM for this feature I'll let you make the call, but if I remember correctly undo is out of scope for this sprint.

src/main/edit_delete/Clip.ts
chloebrett: Oh, should probably add, specify the type of the object as `Clip` and then make an interface for it:

```
interface Clip {
    startTime: number;
    duration: number;
    fileName: string;
}
```

src/renderer/components/TranscriptionBlock.tsx
rileykeane: yep. although won't it still be a double negative using filter? 

`transcription.words.filter((word) => !word.deleted).map(...)`

Am I missing something here?

chloebrett: Nope, because if we use filter we don't need the ternary :)

src/renderer/pages/Project.tsx
chloebrett: Sure, it's more about conforming to the existing approach for reversible actions, but I also realise that standard was only implemented a few days ago and so it's fine if you want to merge this as is and then update later.

rileykeane: undo/redo was not part of the MVP. However since it seems a big chunk of the heavy lifting to implement this is already done, we can potentially look into it next sprint. Of course need to prioritise actual MVP features over this.

src/renderer/components/TranscriptionBlock.tsx
rileykeane: This is fixed now with the solution discussed over slack.

src/renderer/components/TranscriptionBlock.tsx
rileykeane: Ternary has been flipped. Using filter sounds nice on paper but ended up being more difficult than it's worth.

src/renderer/pages/Project.tsx
rileykeane: removed.

src/renderer/pages/Project.tsx
```diff
@@ -1,28 +1,81 @@
...
+    }
+  };
+
+  useEffect(() => {
+    window.addEventListener('keydown', onKeyDown);
+
+    return () => {
+      window.removeEventListener('keydown', onKeyDown);
+    };
+  });
```
ghost, line 79: This use effect will currently add the event listener on every render and remove it at "end" of render. 
Just checking that this was the intention?

If you just wanted to add the event listener when the page loads, and remove it when the page closes, then add an empty dependency array after the function argument.

src/renderer/pages/Project.tsx
```diff
@@ -47,29 +96,20 @@ const ProjectPage = () => {
...
+          <TranscriptionBlock
+            transcription={currentProject.transcription}
+            onWordClick={onWordClick}
+          />
         </Stack>
         <Box sx={{ width: '2px', backgroundColor: 'gray' }} />
         <Stack justifyContent="center" sx={{ width: 'fit-content' }}>
-          <Box sx={{ width: '400px', height: '280px', backgroundColor: 'red' }}>
+          <Box
+            sx={{ width: '400px', height: '280px', backgroundColor: 'black' }}
```
ghost, line 133: lol :)

src/renderer/components/TranscriptionBlock.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+const TranscriptionBox = styled(Box)`
+  background: ${colors.common.white};
+  height: 100%;
+  padding: 20px;
+`;
+
+const Word = styled('span')`
+  &:hover {
+    background: orange;
```
ghost, line 13: I believe this is supposed to be `yellow-500` with 50% opacity. Sorry, thats my bad, didn't add it on figma in hi-fi designs

src/main/editDelete/preProcess.ts
```diff
@@ -0,0 +1,53 @@
...
+const preProcessTranscript = (
+  jsonTranscript: any,
+  duration: number
+): Transcription => {
+  const numberOfWords: number = jsonTranscript.words.length;
+
+  const processedTranscript: Transcription = {
+    confidence: jsonTranscript.confidence,
+    words: []
+  }
```
ghost: This is tiny but this object declaration is missing a semicolon. Has linter / prettier been run on this file?

ghost (changes requested):
Nice work. Got some very sexy bits of code in there. Left a few little comments.

Definitely not in this PR:
Just wanted to say, I'm now really leaning towards gaps between words being treated as their own 'word' object. Would make highlighting to delete easier (highlight could end in a gap) and would allow for possible functionality on those gaps later. Realising this would require rework but reckon it might be worth it.

src/renderer/pages/Project.tsx
chloebrett: Ooh good catch on that one hugh

src/renderer/pages/Project.tsx
rileykeane: Bright red wasn't great on the eyes 😬

src/renderer/components/TranscriptionBlock.tsx
rileykeane: Yep, styling isn't complete. Just a random color for now.

src/main/editDelete/preProcess.ts
rileykeane: It has and there were no errors. Not sure why this wasn't picked up.

rileykeane: @LiamTodd

rileykeane:
> Definitely not in this PR:
> 
> Just wanted to say, I'm now really leaning towards gaps between words being treated as their own 'word' object.


100% on this - I think we also need to consider line breaks and punctuation also.

src/main/editDelete/preProcess.ts
LiamTodd: My bad!

src/renderer/pages/Project.tsx
fwang1703: linter isn't happy with the use of an empty dependency array :'( - will look into it further in another PR

danfeltham (approved):
Got nothing else to add. Big fan of your work

ghost (approved):

src/renderer/components/TranscriptionBlock.tsx
```diff
@@ -0,0 +1,51 @@
...
+                  onWordClick(index);
+                }}
+              >
+                {word.word}
+              </Word>
+            )
+          )
+          .reduce((acc, curr) => (
+            <>
+              {acc} {curr}
```
chloebrett, line 43: 😍

src/renderer/pages/Project.tsx
chloebrett: Yeah justification for it here: https://docs.google.com/document/d/1fBLBj_I3Y4AgRnIHzJ-grsXvzoKUBA03KNRv3DzABAg/edit#heading=h.7c9bao1d81db

On your end all that's needed is do + undo redux actions wrapped in an op, then dispatch that with dispatchOp. All the actual undo functionality is abstracted away

chloebrett: I do understand that it wasn't originally part of the mvp - so defs no need to do it in this PR. All I'm saying is in terms of "make it so the user can undo" we're already 90% of the way there, you just have to provide a reversible action

src/renderer/pages/Project.tsx
chloebrett: Yeah the current impl shouldnt cause any bugs it's just inefficient

src/sharedTypes.ts
chloebrett: Cool, is there an overall scope doc for the edit/delete stuff? Would be good to know what is and isn't being included

chloebrett (approved):
Looking good now, just resolve any open comments from other reviewers and it's all good from me.

Back end pre/post processing might need a bit of a refactor later on but it's fine for now. Just a bit iffy on the code duplication and some of the for loops are a bit hard to read. I can't think of anything immediately better off the top of my head though so happy for it to be merged

chloebrett (approved):
Looking good now, just resolve any open comments from other reviewers and it's all good from me.

Back end pre/post processing might need a bit of a refactor later on but it's fine for now. Just a bit iffy on the code duplication and some of the for loops are a bit hard to read. I can't think of anything immediately better off the top of my head though so happy for it to be merged

src/sharedTypes.ts
rileykeane: The current scope is DWDD (do what Descript does). Happy to write something more formal up (and hopefully make it better then Descript)

PR #42, opened 2022-05-05
Experiment/video preview spike

ghost:
**A spike for video preview.** Note: Don't really need a review on this PR, more just to get it in for my team to use as an example and for everyone in the project to have a look at. Having said that, feel free to comment.

This spike is its own package and you'll need to `yarn` and `yarn start` within the `experiments/video-preview` directory to view it. You'll also need to add a `sample.mp4` file in the `experiments/video-preview/assets` directory

**There is an annoying typescript linting error which I couldn't get rid of (still works despite it). If anyone can fix that I will be indebted for life.**

This spikes makes uses of video streaming. 
A custom video component (`Replay`) makes a request to the express streaming server for a range of bytes from the input video using the video timestamp. The video timestamp is updated when each cut of the input video ends.

A cut of the input video is created whenever the user makes an edit (actually multiple cuts). For example, if they delete a word then 2 cuts are produced on either side of the word. If they move a word then potentially 4 cuts are produced. When the user plays back the current state of the project, we jump around between these cuts and basically pretend the output video has already been created. :)

I estimate that for large files (eg: 1GB), the video preview component needs about 0.2 seconds to load (this could potentially be optimised after MVP). For smaller videos (eg: < 200MB), the video preview component needs about 0.05 - 0.1 seconds to load. Hence, the min clip/cut size/word duration will probably need to be around 0.1 seconds to be recognisable in the video preview.

**Please have a look at these design docs to better understand the process:**
[Design docs diagrams](https://www.figma.com/file/ya0gV3UM0nQmXWc8SfkneQ/Design-Docs?node-id=0%3A1)
[Design doc - potential methods](https://docs.google.com/document/d/1pPkEb5nVR_W6WAQN42NtByDeWRr0r2Lv2mJCV8ck0rY/edit#heading=h.ssh3k2wg6grs)
[Design doc - video streaming](https://docs.google.com/document/d/133W-zR5STUf8zbFyz66H-H5OlwcAB2AFTANI047aB5o/edit)

Thank you!


chloebrett:
> If anyone can fix that I will be indebted for life.

I accept cash credit and american express https://github.com/patrickbrett/mlvet/pull/43

PR #43, opened 2022-05-05
Eslint fix on spike as requested by Hugh

chloebrett:
Don't know why the problem was here in the first place. All I know is that this fixes it and does so really simply.


ghost:
Thank you!

ghost (approved):

PR #44, opened 2022-05-06
Export Progress, EDL now writes source path

itsMoxMox:
Why:

- To link the EDL export progress BE to the progress bar in the FE

Changes:

- EDL construction now updates the frontend with its progress.
- EDL construction now takes in and uses media source path.

Pair Programmed with Manika (@manikag-12)

### DIFF
--- a/src/main/export.ts
+++ b/src/main/export.ts
@@ -12,14 +12,15 @@ const secondToTimestamp: (num: number) => string = (num) => {
     .join(':');
 };
 
-const constructEDL: (title: string, transcription: Transcription) => string = (
-  title,
-  transcription
-) => {
+const constructEDL: (
+  title: string,
+  transcription: Transcription,
+  source: string | null,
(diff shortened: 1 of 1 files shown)


ghost (approved):

src/main/export.ts
```diff
@@ -12,22 +12,27 @@ const secondToTimestamp: (num: number) => string = (num) => {
     .join(':');
 };
 
-const constructEDL: (title: string, transcription: Transcription) => string = (
-  title,
-  transcription
-) => {
+const constructEDL: (
+  title: string,
```
chloebrett, line 16: Instead of passing three distinct parts of a Project type, maybe type this instead as `Partial<Project, 'name' | 'transcription' | 'mediaFilePath'>` and then just pass the whole project as an argument (typescript will block you from accessing other fields you didn't declare)

chloebrett (approved):
Small comment, other than that looks good

PR #45, opened 2022-05-07
Feature/get audio duration

jhea0008:
Adds the audio duration and file name to the output from the transcription handler. 
Currently reads in the audio file from a hardcoded path to assets, but this can be changed to be passed in to the function in the future. 

Also not 100% sure about the interface changes... let me know if anyone is expecting something different


src/main/handlers/transcriptionHandler.ts
```diff
@@ -13,8 +14,21 @@ const sleep: (n: number) => Promise<void> = (n) =>
...
+  filePath: string
+) => Promise<Transcription> = async (filePath: string) => {
+  // TODO: replace hard coded media path with parameter passed in
+  const pathToSaveMedia = path.join(
+    process.cwd(),
+    'assets',
+    'audio',
+    'audio.wav'
+  );
+  const audioDurationPromise = getAudioDurationInSeconds(pathToSaveMedia)
```
chloebrett: Rather than storing as a promise that is immediately awaited, you can just await here:

```
const audioDuration = (await getAudioDurationInSeconds(pathToSaveMedia)) || 0;
```

chloebrett (changes requested):
Looking good, just a small comment

src/main/editDelete/preProcess.ts
```diff
@@ -2,13 +2,15 @@ import { Transcription, Word } from 'sharedTypes';
...
-  duration: number
+  duration: number,
+  fileName: string
 ): Transcription => {
   const numberOfWords: number = jsonTranscript.words.length;
 
   const processedTranscript: Transcription = {
     confidence: jsonTranscript.confidence,
     words: [],
+    fileName,
```
LiamTodd: Everything's looking good!
Was just wondering: Should we attach the filename to a Transcription or a Word object? The Word interface currently has a "filename" field which is unused. I'm unsure of the correct answer, and it would probably matter when we start to have multiple videos layered on top of one another.

LiamTodd (approved):

src/main/editDelete/preProcess.ts
jhea0008: Hmm yea I did see that there's a filename field for each word. I'm open to storing it however it's gonna be easiest. I'm currently not sure what the filename is gonna be used for, something like  identifying which words come from which files? Would it be better to store the full file path for that then?

src/main/handlers/transcriptionHandler.ts
jhea0008: Was thinking the promise can async complete itself while the transcription transcribes, but it's probs not a huge performance difference regardless

src/main/editDelete/preProcess.ts
LiamTodd: I'm not sure exactly how it'll be used either, as I think it will depend on video preview once we have multiple videos going on top of one another.
I suppose it shouldn't matter right now since its a simple change to make (would you agree?) and we don't have a clear picture of how this will be used right now.
If you think it will be better to store the file path instead of just the file name, then that can be passed to preProcess instead.

src/main/handlers/transcriptionHandler.ts
MutazAshhab: Since there is no large hit on performance i think awaiting the promise when the function is called is the way to go.

One less thing to keep track of, and one less variable.

src/main/editDelete/preProcess.ts
MutazAshhab: > Everything's looking good! Was just wondering: Should we attach the filename to a Transcription or a Word object? 

Down the line we might have multiple audio sources. So to make future development easier I think it should be attached to both.

In the case that we end up developing a version with multiple audio source and the `fileName` attribute it unused, then it could be removed.

src/main/handlers/transcriptionHandler.ts
chloebrett: I see what you're saying - if you want to go this route you could use Promise.all, it's a bit cleaner. But I agree with Mutaz, the "get duration" operation should be very fast so I would go with the simpler approach

src/main/editDelete/preProcess.ts
chloebrett: I am fine with using media path as an identifier of sorts right now, but down the line we should have a unique identifier that doesn't change in case the media file gets moved

chloebrett: With regards to where it's used, Word, Transcription, either is fine - it's easily changeable later

src/main/handlers/transcriptionHandler.ts
jhea0008: Ok, I've gone with the immediate await in the latest commit

src/main/editDelete/preProcess.ts
jhea0008: Yea, very easy change if we need it as a path/name and whether it's stored in words or transcriptions. For now, I've pushed a commit to save it in words and not transcriptions, just so we don't have differing ways of accessing the same value if it's not functionally required.

MutazAshhab (approved):

chloebrett (approved):

PR #46, opened 2022-05-07
Refactor pre process logic and add tests

chloebrett:
Origin story: I left a comment on the edit-delete PR suggesting that the pre and post processing logic be refactored to use `map`, but was ambiguous about how.

Riley sent some ideas over message, but to be honest this was a bit too involved to be explained through code snippets - it needed a PR.

I've done the pre process logic as an example - I have left the post process logic for you guys to try. This is a nice-to-have - it makes the code easier to read and easier to modify, but it's not strictly necessary or part of the sprint, so do what you like with this code (including not using it at all if you like) - definitely don't want to stress you out about timelines.

The basic process I followed was:
- Write some unit tests for the existing logic to capture what it's trying to do.
- Think about discrete steps that the logic can be broken down into, and what their inputs and outputs are.
- Try to frame each step as a pure function without any side effects - i.e. takes an input, returns an output, nothing else.
- Write helper types where necessary, and make everything is well typed (e.g. no 'any's).

Also, in doing this I noticed a really minor bug - the floating point subtractions sometimes produce really weird duration values (e.g. 3.199999999993 instead of 3.2), so I've added a util function to round these to the nearest millisecond.

I hope you can follow the same idea for the post processing logic, if you can find some time to do it - it will be a good exercise in writing clean code and contribute to the maintainability of the codebase 🙂  I would be happy to pair on this if that interests any of you.

I think this also serves as a good example of how test driven development can be useful for very back-end problems like this one.

### DIFF
--- a/src/main/editDelete/__tests__/preProcess.test.ts
+++ b/src/main/editDelete/__tests__/preProcess.test.ts
@@ -1,0 +1,102 @@
+import preProcessTranscript from '../preProcess';
+
+describe('Test pre-processing JSON transcript into regular transcript', () => {
+  it('should produce expected transcript when given a JSON transcript', () => {
+    const jsonTranscript = {
+      confidence: 1,
+      words: [
+        {
+          word: 'abc',
+          start_time: 0,
+          duration: 1,
+        },
+        {
+          word: 'def',
+          start_time: 1,
+          duration: 1,
+        },
+      ],
+    };
+    const duration = 2;
+
+    const outputTranscript = preProcessTranscript(jsonTranscript, duration);
+
+    expect(outputTranscript).toEqual({
+      confidence: 1,
+      words: [
+        {
+          deleted: false,
+          duration: 1,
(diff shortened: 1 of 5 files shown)


LiamTodd (approved):
Thanks for this! I got the gist of what you meant about not having repetitive code to clean up edge cases for this sort of thing but didn't know how to implement the changes cleanly. Glad that I can see how to do it now.
Will try and clean up postProcess in a similar fashion if time allows.

chloebrett:
No worries!

I just pushed some more changes re: a discussion with Riley about runtime type validation. If there are other things we need to validate in a similar way, it may be useful to do a more generalised approach using something like https://github.com/YousefED/typescript-json-schema. For now, I think this does the job better than just casting and hoping for the best. Open to suggestions.

rileykeane (approved):

Vita1287 (approved):
Makes sense. We should be able to do the same for postProcess

PR #48, opened 2022-05-09
Project Window UI

danfeltham:
Here's the UI for the project window that reflects the HI-FI designs in Figma.

### Changes:
- Colour matching to the Figma designs
- New VideoController component for video playback
- Scrollbar for the transcription text block


src/renderer/components/VideoController.tsx
```diff
@@ -0,0 +1,41 @@
...
+  column-gap: 24px;
+`;
+
+const VideoController = () => {
+  return (
+    <VideoControllerBox>
+      <div
+        style={{
+          backgroundColor: colors.grey[600],
+          fontWeight: 'bold',
```
emmapainter: Could you change this to regular, size 28?

src/renderer/components/TranscriptionBlock.tsx
```diff
@@ -1,16 +1,30 @@
...
+    border-radius: 10px;
+    background: ${colors.yellow[500]};
+  }
 `;
 
 const Word = styled('span')`
   &:hover {
-    background: orange;
+    color: ${colors.grey['000']};
+    background: rgba(255, 187, 85, 0.5);
```
rileykeane: I just confirmed with @emmapainter that this color code was an error in the designs. It should be yellow 500 at 50% opacity.

emmapainter (changes requested):
Could you decrease the gap between the top of the transcription text and the box it sits in? Also can you change the line divider in the middle of the screen to use grey-600? This looks awesome, love your work.

src/renderer/components/VideoController.tsx
```diff
@@ -0,0 +1,41 @@
...
+          fontSize: '24px',
+          borderRadius: '5px',
+          padding: '0 19px',
+        }}
+      >
+        00:00:00
+      </div>
+      <Replay10Icon sx={{ fontSize: '36px' }} />
```
rileykeane: I believe these will need to be [icon buttons](https://mui.com/material-ui/api/icon-button/) so they can have an `onClick` handler

rileykeane (changes requested):
This looks awesome! Just a few minor fixes

- The word highlight color was incorrect in Figma and needs to be updated
- Make the playback control icons buttons.

src/renderer/components/TranscriptionBlock.tsx
ghost: I believe the new line can be written as `rgba(colors.yellow[500], 0.5)`

rileykeane: I tried that too but couldn't get it to work. According to [this](https://stackoverflow.com/questions/7015302/css-hexadecimal-rgba) very old SO thread 8 digit hex codes were going to be supported. (where the last 2 digits represent opacity). 

This `background: ${colors.yellow[500]}80;` seems to work but feels wrong tbh...

I can't find any official W3C resources that confirm its supported.

src/renderer/components/VideoController.tsx
ghost: Could you also please add an IconButton with a pause icon? Maybe just test it and then comment it out. 
Team Farquaad can add the logic for switching between the two if you like.

ghost: nice job getting all these the right size :)

danfeltham: Done and done, ill push the commits soon

src/renderer/components/VideoController.tsx
danfeltham: Done and done, will push changes momentarily

danfeltham:
I've implemented all the changes you guys asked for. Merge this bad boy

ghost (approved):

rileykeane (approved):

emmapainter (approved):
Amazing!!

PR #49, opened 2022-05-09
add ability to navigate using `enter` key

MutazAshhab:
## What changed?

Now you can navigate to the next view in the new project sequence using the `ENTER` key

### DIFF
--- a/src/renderer/components/NewProjectModal/ImportMediaView.tsx
+++ b/src/renderer/components/NewProjectModal/ImportMediaView.tsx
@@ -47,6 +47,7 @@ const ImportMediaView = ({ prevView, closeModal, nextView }: Props) => {
   const currentProject = useSelector(
     (store: ApplicationStore) => store.currentProject
   );
+  const projectName = currentProject?.name;
 
   const dispatch = useDispatch();
 
@@ -104,8 +105,6 @@ const ImportMediaView = ({ prevView, closeModal, nextView }: Props) => {
   if (currentProject === null) {
     return null;
   }
-
-  const projectName = currentProject.name;
 
   const transcribeButton = (
     <CustomButton
--- a/src/renderer/components/NewProjectModal/ImportMediaView.tsx
+++ b/src/renderer/components/NewProjectModal/ImportMediaView.tsx
@@ -1,4 +1,4 @@
-import { useState } from 'react';
+import { useCallback, useEffect, useState } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import { Box, Button, Stack, styled, Typography } from '@mui/material';
 import colors from 'renderer/colors';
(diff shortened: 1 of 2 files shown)


src/renderer/components/NewProjectModal/ImportMediaView.tsx
```diff
@@ -74,7 +79,33 @@ const ImportMediaView = ({ prevView, closeModal, nextView }: Props) => {
...
+    return () => {
+      window.removeEventListener('keypress', handleKeypress);
+    };
+  }, [handleTranscribe, isAwaitingMedia]);
+
+  if (currentProject === null) {
+    return null;
+  }
+
+  const projectName = currentProject.name;
```
ghost: Not opposed to adding vars to make the tsx more readible. However, might be more readible if the variable was defined at the top of the function, just after `currentProject` is defined. 
Could also use optional chaining (?.) unless we want to fail fast.

src/renderer/components/NewProjectModal/NewProjectView.tsx
```diff
@@ -53,18 +53,35 @@ const NewProjectView = ({ closeModal, nextView }: Props) => {
...
     const project = await makeProjectWithoutMedia(projectName);
     if (project === null) {
       return;
     }
     setProjectInStore(project);
     nextView();
-  };
+  }, [nextView, projectName, setProjectInStore]);
+
+  useEffect(() => {
```
ghost, line 72: While I love the hook, wondering if it is possible to just use a form around the text field and button (like a web app would)? Then add an `onSubmit` handler to the form and make the button `type="submit"`.

src/renderer/components/NewProjectModal/ImportMediaView.tsx
Taz17: It cannot be, I cannot have `useDispatch` beneath a conditional statement (really, and early return). For that reason I am considering removing the `projectName` variabel and using `currentProject.name` instead.

```javascript
  const currentProject = useSelector(
    (store: ApplicationStore) => store.currentProject
  );

  const dispatch = useDispatch(); // eslint + react will cry if it is beneath
                                  // the early return since it wont know how many times to render.

  if (currentProject === null) {
    return null;
  }

  const projectName = currentProject.name;
```

Taz17: Oh sorry, I took that code snippet from the wrong branch.

src/renderer/components/NewProjectModal/NewProjectView.tsx
MutazAshhab: I like that approach, I gave it a shot. I did not like the outcome though. I published a temporary branch to show you what I have tried and so you can run it locally and possibly see what is going on.

Here is the commit where I replaced the hook based approach with the form based approach. https://github.com/patrickbrett/mlvet/commit/9962fce5bf0dec5a3c0195f2e5112101f503a67e

The branch name is temporary/naviagte-with-return (yes there is a typo)

Here is what I noticed:
- TextField was pushed down down closer to the `Cancel` and `Continue` buttons. when the FormControl component was added
- autoFocus on TextField no longer works when the modal is open.
- The who component had to be wrapped with `<form> </form>` tags for it to work

I do agree the form approach is much simpler, and is the way to go. 

Am i using the form incorrectly in this case?

ghost: I have added an example of a working form (though not very sexy) to your branch. If possible, please try and implement it in this branch the same way. :)

src/renderer/components/NewProjectModal/ImportMediaView.tsx
MutazAshhab: Here is the commit for this fix: https://github.com/patrickbrett/mlvet/pull/49/commits/a72a93b92a50ddd3094264aa4feff495f15f4315

MutazAshhab:
I will be pausing development of this branch and here is an explanation of what happened and why:

- The solution i initially had required quite a bit of code and was not straight forward.
- Hugh suggested that we implement `navigating by using the enter key` by leveraging HTML forms since we can attached the handler function to the button of type `submit`.
- However, using HTML forms meant that the user would only be able to navigate using `enter` when the user was focused on one of the forms' elements.
- When it came to wrapping the `SelectMediaBlock` with a `form` tag it didnt work because an element has to be focused on (`SelectMediaBlock` functions like a button, meaning if you focus on it by clicking it would open the file dialog).
- Wrapping the whole component with a form didnt work as well.

In this sprint Irmy will be working on drag and drop, which might change the implementation. So im going to wait on that to happen first before going ahead with this feature.

Here is the branch with the attempt of wrapping the component with a form: https://github.com/patrickbrett/mlvet/tree/temporary/naviagte-with-return

PR #50, opened 2022-05-09
enable auto focus on textfield when user selects to create new project

MutazAshhab:
## What changed:
- When a user selects `New project` the modal will pop up as normal but now with the cursor focused on the TextField component.

## Why has this change been made?

Since we will be adding functionality to navigate between modal views using the `enter` key soon (soon, like in this PR #49) it makes sense to just already be focused on the textfield such that the user can type, hit enter, import video, hit enter. without touching the mouse.

## Here is what it looks like:

image


emmapainter (approved):
Nice, love the attention to detail.

PR #53, opened 2022-05-11
Origin/feature/updated start time

Vita1287:
Changed attributes for Word to include start times for the input video and the output video. Note that Patrick's changes for preProcess have been used in this PR. The original startTime has been removed so front-end may need to change.

Changes:
- In sharedTypes, Word interface was updated to include inputStartTime and outputStartTime and startTime is removed
- preProcess was refactored to become compatible with the new Word interface

### DIFF
--- a/src/main/editDelete/__tests__/preProcess.test.ts
+++ b/src/main/editDelete/__tests__/preProcess.test.ts
@@ -1,0 +1,102 @@
+import preProcessTranscript from '../preProcess';
+
+describe('Test pre-processing JSON transcript into regular transcript', () => {
+  it('should produce expected transcript when given a JSON transcript', () => {
+    const jsonTranscript = {
+      confidence: 1,
+      words: [
+        {
+          word: 'abc',
+          start_time: 0,
+          duration: 1,
+        },
+        {
+          word: 'def',
+          start_time: 1,
+          duration: 1,
+        },
+      ],
+    };
+    const duration = 2;
+
+    const outputTranscript = preProcessTranscript(jsonTranscript, duration);
+
+    expect(outputTranscript).toEqual({
+      confidence: 1,
+      words: [
+        {
+          deleted: false,
+          duration: 1,
+          fileName: 'PLACEHOLDER FILENAME',
+          key: '0',
+          startTime: 0,
+          word: 'abc',
+        },
+        {
+          deleted: false,
+          duration: 1,
+          fileName: 'PLACEHOLDER FILENAME',
+          key: '1',
+          startTime: 1,
+          word: 'def',
+        },
+      ],
+    });
+  });
+
+  it('should fill blank spaces between words, and between last word and end', () => {
+    const jsonTranscript = {
+      confidence: -7,
+      words: [
+        {
+          word: 'heat',
+          start_time: 5,
+          duration: 2,
+        },
+        {
+          word: 'from',
+          start_time: 8,
+          duration: 3,
+        },
+        {
+          word: 'fire',
+          start_time: 11.2,
+          duration: 0.9,
+        },
+      ],
+    };
+    const duration = 15.77;
+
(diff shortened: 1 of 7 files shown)


LiamTodd (approved):
Looking good, chad

src/main/editDelete/preProcess.ts
```diff
@@ -1,52 +1,74 @@
...
-  processedTranscript.words.push(lastWord);
+/**
+ * Injects extra attributes into a PartialWord to make it a full Word -
+ * @param word the word to fill attributes for
+ * @param i index of the word in the transcript
+ * @returns full Word object
+ */
+const injectAttributes: MapCallback<PartialWord, Word> = (word, i) => ({
+  ...word,
+  outputStartTime: 0,
```
rileykeane: Can you please adjust the initial `outputStartTime` to be equal to `startTime`

src/sharedTypes.ts
```diff
@@ -28,8 +28,9 @@ export interface Transcription {
 
 export interface Word {
   word: string;
-  startTime: number;
+  inputStartTime: number;
```
rileykeane: Have you confirmed changing this won't break anything with export?

src/main/editDelete/postProcess.ts
```diff
@@ -1,70 +1,70 @@
-import { Transcription, Clip } from 'sharedTypes';
+// import { Transcription, Clip } from 'sharedTypes';
```
ghost: Just wanted to clarify what is going on in this file. Is the code commented out for reason? Is it going to make a comeback?

src/main/editDelete/__tests__/preProcess.test.ts
```diff
@@ -0,0 +1,102 @@
...
+          start_time: 11.2,
+          duration: 0.9,
+        },
+      ],
+    };
+    const duration = 15.77;
+
+    const outputTranscript = preProcessTranscript(jsonTranscript, duration);
+
+    expect(outputTranscript).toEqual({
```
rileykeane, line 72: I'm surprised these test cases passed. I believe they need to be updated to included the new keys in `Word`

src/main/editDelete/postProcess.ts
LiamTodd: We commented it out for now as we are going to refactor it and is not being used. Currently making a comeback lol

rileykeane (changes requested):
Thanks for updating the `Word` model. 

Please just fix the initial `outputStartTime` value and also confirm that changing the `startTime` key won't break anything - it won't effect the project page in the front end but might cause issues else where. 

Also, please remove any commented out code that you don't need anymore. 

@hhen0005 can you please confirm the `outputStartTime` is what you require for video playback?

src/main/editDelete/__tests__/preProcess.test.ts
rileykeane: I'm just now realizing you didn't create these and its just from Patrick's refactor that hasn't been merged with develop.

Vita1287:
> Thanks for updating the `Word` model.
> 
> Please just fix the initial `outputStartTime` value and also confirm that changing the `startTime` key won't break anything - it won't effect the project page in the front end but might cause issues else where.
> 
> Also, please remove any commented out code that you don't need anymore.
> 
> @hhen0005 can you please confirm the `outputStartTime` is what you require for video playback?

Vita1287:
Changes have been made to the following:
inputStartTime has been changed to startTime
outputStartTime is now initially assigned startTime

src/sharedTypes.ts
zwan0202: We will fix the export after this PR is merged

zwan0202 (approved):

rileykeane (approved):

PR #54, opened 2022-05-11
Space between words

LiamTodd:
Changes:
- Added reduce method call to preProcess to add a space between each word as a Word object with an empty string as the word field, as requested from FE
- As well as a space between each word, there is a space padding the beginning and the end of the transcript, for example if the first word begins at 0.5s, there is a space object from 0 to 0.5 (similar at the end if the last word does not reach the end of the video)


Vita1287 (approved):
Looks good, might need to change how we get the duration.

src/main/editDelete/preProcess.ts
```diff
@@ -50,6 +31,82 @@ const injectAttributes: MapCallback<PartialWord, Word> = (word, i) => ({
...
+        spaceChar,
+        word.startTime + word.duration,
+        words[index + 1].startTime - word.startTime - word.duration,
+        word.startTime + word.duration,
+        (index * 2 + 2).toString()
+      )
+    );
+  }
+  // is last word
+  else {
```
rileykeane: It might be better here to use a similar approach to how Patrick did `fillDurationGaps` to avoid duplicating code.

LiamTodd: good call, addressed it here: https://github.com/patrickbrett/mlvet/pull/54/commits/837cda0465a8fef63d9ed07412c3e779d1507079

src/main/editDelete/preProcess.ts
```diff
@@ -50,6 +31,74 @@ const injectAttributes: MapCallback<PartialWord, Word> = (word, i) => ({
...
 });
 
+const constructWord = (
+  word: string,
+  startTime: number,
+  duration: number,
+  outputStartTime: number,
+  key: string
+) => {
+  const newWord: Word = {
```
chloebrett: * Inline this and directly return the object
* Give the function a type definition including a return type
* Can make the function inline as well. See example below

`const returnAnObject: (something: string) => { putHere: string; } = (something) => ({ putHere: something });`

src/main/editDelete/preProcess.ts
```diff
@@ -50,6 +31,74 @@ const injectAttributes: MapCallback<PartialWord, Word> = (word, i) => ({
...
+/**
+ * Adds spaces between words which represent the silence between each word
+ * @param result The output being constructed
+ * @param word The current element of words
+ * @param index The index of word in words
+ * @param words The list of words being reduced
+ * @returns The updated transcript with a silence after word
+ */
+let TOTAL_DURATION = 0;
+const addSpaces = (
```
chloebrett: Maybe make a helper type like we have for MapCallback and give it that: i.e.

`type ReduceCallback<T, U> = (acc: T[], curr: T, index: number; arr: T[]) => U;`

Then the type of this function can just be `ReduceCallback<Word, Word[]>` i.e. it takes a list containing Word objects and returns another list of word objects

Nice work using reduce by the way! It's a tricky one to understand

src/main/editDelete/preProcess.ts
```diff
@@ -50,6 +31,74 @@ const injectAttributes: MapCallback<PartialWord, Word> = (word, i) => ({
...
+  const isLastWord = index === words.length - 1;
+  let silenceDuration;
+  if (isLastWord) {
+    silenceDuration = TOTAL_DURATION - word.startTime - word.duration;
+  } else {
+    silenceDuration =
+      words[index + 1].startTime - word.startTime - word.duration;
+  }
+
+  word.key = (index * 2 + 1).toString();
```
chloebrett: Add comments explaining what this index is for and why it's the n*2+1'th, otherwise they're just magic numbers

src/main/editDelete/preProcess.ts
```diff
@@ -60,12 +109,13 @@ const preProcessTranscript = (
...
 ): Transcription => {
+  TOTAL_DURATION = duration;
   return {
     confidence: jsonTranscript.confidence,
     words: jsonTranscript.words
       .map(camelCase)
-      .map(fillDurationGaps(duration))
-      .map(injectAttributes),
+      .map(injectAttributes)
+      .reduce(addSpaces, []),
```
chloebrett: 😍

src/main/editDelete/preProcess.ts
```diff
@@ -50,6 +31,74 @@ const injectAttributes: MapCallback<PartialWord, Word> = (word, i) => ({
...
+  result: Word[],
+  word: Word,
+  index: number,
+  words: Word[]
+) => {
+  const spaceChar = ' ';
+
+  // is the first word
+  if (index === 0) {
+    result.push(
```
chloebrett: The idea with reduce is that it's immutable - in other words:

* it gives you an accumulator (the result so far) as well as the current value,
* and you return the new value of the accumulator.

So rather than mutating the accumulator, try to build the new accumulator out of the existing one + current value. This will most likely involve using concat() instead of push().

(This works very similarly to redux 'reducers' by the way, which take a 'state' and an 'action', then return the new 'state' after the action is applied, without actually editing the old one. If you're wondering why immutability is sought after here, it's mostly because it's easier to debug & reason about).

src/main/editDelete/preProcess.ts
```diff
@@ -50,6 +31,74 @@ const injectAttributes: MapCallback<PartialWord, Word> = (word, i) => ({
...
+      constructWord(spaceChar, 0, words[index].startTime, 0, index.toString())
+    );
+  }
+
+  const isLastWord = index === words.length - 1;
+  let silenceDuration;
+  if (isLastWord) {
```
chloebrett: Bit of a nitpick, but might be simpler to use an intermediate variable defined with a ternary, that way you can make silenceDuration a const

i.e.
`const endTime = isLastWord ? TOTAL_DURATION : words[index + 1].startTime`
then use endTime in both cases

chloebrett (changes requested):
Looking really cool, love that you're embracing the use of array methods. Just a few comments

* The unit tests for preProcess are failing now - try `yarn test`. Please update the tests so they reflect the expected logic. Don't worry about the tests for other files (currently just App.tsx), those are unrelated
* Please use a prefix for your branch name e.g. feature/, chore/, bugfix/

LiamTodd:
> Looking really cool, love that you're embracing the use of array methods. Just a few comments
> 
> * The unit tests for preProcess are failing now - try `yarn test`. Please update the tests so they reflect the expected logic. Don't worry about the tests for other files (currently just App.tsx), those are unrelated
> * Please use a prefix for your branch name e.g. feature/, chore/, bugfix/

Cheers! Will address the unit tests on Thursday (a bit late right now). 
Sorry about the branch name convention, just missed that -I tried renaming the branch but it closed the PR so I just reopened it and the branch will have to keep its name (unfortunately).

src/main/editDelete/preProcess.ts
LiamTodd: cool trick, addressed here: https://github.com/patrickbrett/mlvet/pull/54/commits/c0d453bab2060cc1c723a21702e63c7782237f36

src/main/editDelete/preProcess.ts
LiamTodd: Good point, added some comments here: https://github.com/patrickbrett/mlvet/pull/54/commits/ff252792d7273be27cb6c53d3faa103ad833af00

src/main/editDelete/preProcess.ts
LiamTodd: Ahhh I see how that words now. Wouldn't have figured that one out myself (I tried lol). Addressed here: https://github.com/patrickbrett/mlvet/pull/54/commits/8edf6d64c9d50d8fba9ae4f007537ad444542a37

src/main/editDelete/preProcess.ts
LiamTodd: Typescript does my head  in, but I think this is what you wanted: https://github.com/patrickbrett/mlvet/pull/54/commits/a759e9bc4fff5160ddc28509ba3b9529624d27b6

src/main/editDelete/preProcess.ts
LiamTodd: Not going to lie, I'm not sure how to implement these changes and also not exactly sure why, as it's working as intended. But: I'm open to making changes if you could give a little more of an explanation on why its better to use concat rather than push

LiamTodd: https://github.com/patrickbrett/mlvet/pull/54/commits/95a6f7be59fa28c3ebc165d1ab048625e448506f

LiamTodd:
@patrickbrett All changes requested have been made, including using .map instead of .reduce, although I used .flat() instead of .flatMap since there was no point writing a new callback function which didn't do anything which .flat() doesn't already do.

LiamTodd:
Updated test cases, swear I'm done now @patrickbrett @rileykeane

chloebrett:
> @patrickbrett All changes requested have been made, including using .map instead of .reduce, although I used .flat() instead of .flatMap since there was no point writing a new callback function which didn't do anything which .flat() doesn't already do.

Oh my that's beautiful, looks like an ES2019 change I missed 😍  great find! Will review properly in a bit

LiamTodd:
> > @patrickbrett All changes requested have been made, including using .map instead of .reduce, although I used .flat() instead of .flatMap since there was no point writing a new callback function which didn't do anything which .flat() doesn't already do.
> 
> Oh my that's beautiful, looks like an ES2019 change I missed 😍 great find! Will review properly in a bit

Awesome, thanks boss. For some reason yarn test isn't running for me, but I'm certain that the expected output matches the actual output from manual testing.

rileykeane (approved):
This looks great! Seems to work fine with the frontend and I can now remove the pesky reduce function I was using to render spaces 🙂.

chloebrett (approved):
Really nice! Looks good to me

PR #57, opened 2022-05-12
EDL Export Cut Number Bugfix

itsMoxMox:
Fixed a bug where the cut number in the EDL isn't padded by the correct number of zeros.

It used to do something like:
`00000000000014 AX AA/V C    00:00:12:10    00:00:13:01`

Now it looks like:
`014 AX AA/V C    00:00:12:10    00:00:13:01`

Literally like a dozen characters of code.

### DIFF
--- a/src/main/export.ts
+++ b/src/main/export.ts
@@ -23,7 +23,7 @@ const constructEDL: (
   const entries = words.length;
   output += words
     .map((word, i) => {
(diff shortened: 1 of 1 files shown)


LiamTodd (approved):
all goods boss

zwan0202 (approved):

itsMoxMox:
Sick thanks boys 👑

PR #58, opened 2022-05-12
Bugfix/fix preprocess test

jhea0008:
Fixes the unit test preProcess.test.ts, which needed to be updated with new return structure. 
Adds testing section to readme
Adds src to jest modules to help with App.test.tsx

Note: App.test.tsx does not work and is expected to fail


Vita1287 (approved):
Look's good for now. Might need some changes when Team Shrek implements spaces/punctuation.

chloebrett (approved):

jhea0008:
> Look's good for now. Might need some changes when Team Shrek implements spaces/punctuation.

Yea, hopefully we can get things to a point where we have everyone updating the test cases as they update the functions

PR #60, opened 2022-05-12
Save and save-as functionality

chloebrett:
"But wait!" I hear you say. "We already have save!"

Well, concerned friend, fear not because the save we had was actually a really bad save as in disguise, which had no notion of saving to the same file multiple times (like, you know, most software does), and would even let you try to save when you didn't have a project open (cringe!)

This new and improve saving solution, will not only _**save**_ you time, but it will _**save as**_ well yourself, from getting headaches when you try to save.

Okay enough faff, what does this PR do?

- Reworks the whole 'open a project' command to take an optional file path, so that we can use it for both opening arbitrary mlvet files as well as opening known ones
- Reworks the 'save a project' command so that if the project knows where it's meant to be saved, you can bet it will save it exactly there (and if it doesn't, it will ask the user, as they are all-knowing)
- Adds a 'save as' command which creates a copy of the project (including a new ID, new name, etc) and prompts the user to save it to a new location. If they are cheeky and try to overwrite an existing project file, unspecified behaviour may occur, but I have heard the worst case scenario here is simply the computer may faint (better than some outcomes). Oh! And it puts this in the recent projects as well, nice!
- Keeps track of saved / opened files in the recent projects list (no more unsaved files trying to sneak in there, causing all sorts of bugs and mayhem when they are clicked)
- Turns the menu save / save as buttons on and off as necessary as projects are opened and closed - people shouldn't be able to 'save as' projects that they haven't saved yet, for example - or try to save when they don't have a project open.

That is all!

I have tagged:
- Max because he has done some back endy stuff and this is back endy stuff
- Tristan because he's on the import team
- Riley because he's about to tag me in a PR so I thought I'd return the favour


src/main/handlers/openProjectHandler.ts
```diff
@@ -46,13 +46,18 @@ const openProjectFromFile: (filePath: string) => Promise<Project> = async (
...
+  filePath: string | null,
   mainWindow: BrowserWindow | null
-) => Promise<{ project: Project; filePath: string }> = async (mainWindow) => {
-  const filePath = await getOpenFilePath(mainWindow);
+) => Promise<{ project: Project; filePath: string }> = async (
+  filePath,
+  mainWindow
+) => {
+  const openFilePath =
+    filePath === null ? await getOpenFilePath(mainWindow) : filePath;
```
itsMoxMox: IIRC there's a coalescing operator that you could use here that'll shorten this.

```suggestion
    filePath ?? await getOpenFilePath(mainWindow);
```

src/main/handlers/saveProjectHandler.ts
```diff
@@ -32,15 +40,35 @@ const saveProjectToFile: (
   await writeFile(filePath, projectAsString);
 };
 
-const handleSaveProject: (
+export const handleSaveProject: (
   mainWindow: BrowserWindow | null,
   project: Project
 ) => Promise<string> = async (mainWindow, project) => {
-  const filePath = await getSaveFilePath(mainWindow);
+  const filePath =
```
itsMoxMox, line 13: Similar thingo with the `??` operator in place of a null-comparing ternary
```suggestion
  const filePath = project.projectFilePath ?? await getSaveFilePath(mainWindow, project.name)
```

src/main/handlers/recentProjectsHandler.ts
```diff
@@ -3,20 +3,28 @@ import fs from 'fs/promises';
...
   const recentProjectsJson = JSON.stringify(recentProjects);
 
   await fs.writeFile(RECENT_PROJECTS_PATH, recentProjectsJson);
 };
 
+/**
+ * Type for project persisted to disk in the recent projects list -
+ * includes the intersection of project and recent project fields (i.e. no metadata
+ * as this can go stale)
+ */
```
exquisitebin: Maybe increase the clarity of the comment. I'm still not entirely sure of the difference between a RecentProject and a PersistedRecentProject. Or perhaps change the type name?

exquisitebin (commented):
image

I had the above tests not pass.

image

Additionally, when I enter the transcription editing page as seen above, and make some changes, I cannot save the project.

image

Potentially due to this, I cannot see any recent projects (unless the schema version has changed, I think I should have some recent projects that already exist).

Not entirely sure what the cause is or if is only on my end, or something I have done wrong.

chloebrett:
> image
> 
> I had the above tests not pass.
> 
> image
> 
> Additionally, when I enter the transcription editing page as seen above, and make some changes, I cannot save the project.
> 
> image
> 
> Potentially due to this, I cannot see any recent projects (unless the schema version has changed, I think I should have some recent projects that already exist).
> 
> Not entirely sure what the cause is or if is only on my end, or something I have done wrong.

Will check up on all this soon, this branch is a little out of sync with develop now so pulling in the latest should fix tests etc. Thanks for doing manual testing on this

chloebrett:
@exquisitebin ready for re-review.

1) Regarding the tests, I've merged in the local test fixes and everything is working on my end, please test on yours
2) Regarding the recent projects, these will only be displayed if they have a `projectFilePath` specified, which is most likely why yours weren't working. Try saving a new project and see if that comes up in the recents list
3) Regarding the save not working, nice catch, this was because the windows/linux file menu item was missing an identifier. I've fixed and it works on my linux machine, please test again

exquisitebin (changes requested):
Tests are passing, save and save as are both functioning well, and recent projects are also appearing! 

image

Just having one issue. In the recent projects list there is "Copy of New Project 2", but I had actually opened "New Poject 2" and 'saved as' naming it "New Project 3" as you can see on my desktop. Just need to fix this discrepancy in file names (I'm guessing when it saves the metadata isn't changed?).

chloebrett:
Nice find, I hadn't thought of the case where the user would want to rename it. Fixed so that the name in the recents list matches the saved file name. Ready for re-review

exquisitebin (commented):
Excellent, the 'save as' name is working now!

I have found some other things but not sure whether it is currently in the scope of this branch or MVP:

_If you manually change the filename of the last opened project file (after closing the application), when you open the application up again, all recent projects disappear, including those that didn't have a name change._

Now while the above might not be worth fixing at this stage I think the following probably is:

_After reopening a file (that isn't in recent files) and then working on it and pressing 'save', the project doesn't appear in recent files when you restart the application._

Otherwise its all working great!

chloebrett:
Great finds, love the QA initiative. Will investigate and fix these two

chloebrett:
Fixed both problems, ready for re-review.

I've identified a couple of minor things which I will address later:
- Projects persist their own file path to disk, which is not ideal as if that file path changes that data is instantly stale (and if it doesn't, there's no point persisting it). So remove that field and instead calculate it on load, and
- When projects are copied manually (i.e. from the OS file system), their UUID gets copied, which means it's no longer unique. So perhaps we should instead think of UUIDs as metadata solely for the recent projects list - either that or not have them at all. I'll need to think more about where and when they are useful and if we need them

src/renderer/components/NewProjectModal/ImportMediaView.tsx
```diff
@@ -4,19 +4,18 @@ import { Box, Button, Stack, styled, Typography } from '@mui/material';
 import colors from 'renderer/colors';
 import CloseIcon from '@mui/icons-material/Close';
 import IconButton from '@mui/material/IconButton';
-import { ApplicationStore } from 'renderer/store/sharedHelpers';
-import { projectCreated } from 'renderer/store/currentProject/actions';
-import { recentProjectAdded } from 'renderer/store/recentProjects/actions';
-import { Project } from 'sharedTypes';
+import { ApplicationStore } from '../../store/sharedHelpers';
+import { projectCreated } from '../../store/currentProject/actions';
+import { Project } from '../../../sharedTypes';
```
itsMoxMox, line 9: Appreciate the cleanups, we should also put somewhere that this is the preferred styling. Maybe in #process or something.

src/renderer/util.ts
```diff
@@ -161,3 +161,8 @@ export const formatDate: (date: Date) => string = (date) => {
 
   return [dd, mm, yy].map(pad).join('/');
 };
+
+export const removeExtension: (fileName: string) => string = (fileName) => {
+  const split = fileName.split('.');
+  return split.slice(0, split.length - 1).join('.');
```
itsMoxMox, line 167: Is there a necessity to do `.join('.')` instead of `.join('')`? 

They seem to do the same thing:

image

itsMoxMox (approved):

src/renderer/components/NewProjectModal/ImportMediaView.tsx
chloebrett: Yeah we could automate this, idk if people prefer absolute or relative though

src/renderer/util.ts
chloebrett: yeah because if the file has multiple dots 

e.g.

someproject.2022.mlvet

would change to 

someproject2022

if we join('')

edge case but may as well account for it, dots in file names are allowed

exquisitebin (approved):
Seems to be working great and is all looking good. Happy to approve

PR #63, opened 2022-05-13
Extract audio after media is imported 

MutazAshhab:
- The app requires the audio of the imported media to be available for use.

### Some context
- Before this PR audio extract was just a function being called when the app started (this was a demo when Patrick and I worked on that initial feature)

### What did I change?
- I changed the implementation the `extractAudio` function which now takes in a project as input rather than (simply because the project object has the fields that are required by the method at that stage).
- Now saves the extracted audio under `mlvet/extractedAudio/<project-id>.wav`
    - I used the `project id` as the name for the audio file instead of project name since it is unique and we will never have to deal with `project id` changing

### reviewers
Justin, because he brought this up during the SA meeting + his work depends on it
Kate, because she also works on the transcription team
Dan, because i've never had him in any of my PR's before

### DIFF
--- a/src/main/handlers/audioExtract.ts
+++ b/src/main/handlers/audioExtract.ts
@@ -1,22 +1,29 @@
 import path from 'path';
 import { path as ffmpegPath } from '@ffmpeg-installer/ffmpeg';
 import ffmpeg from 'fluent-ffmpeg';
+import { Project } from 'sharedTypes';
+import { audioExtractStoragePath, mkdir } from '../util';
 
 ffmpeg.setFfmpegPath(ffmpegPath);
 
-export default function extractAudio(
-  absolutePathToMediaFile: string
-): Promise<string> {
+export default function extractAudio(project: Project): Promise<string> {
+  if (project.mediaFilePath === null || project.id === null) {
+    return new Promise((_resolve, reject) => {
+      const errorMessage = 'Project mediaFilePath or ID cannot be null';
+      reject(errorMessage);
+    });
+  }
+
+  mkdir(audioExtractStoragePath());
+
   const pathToSaveMedia = path.join(
-    process.cwd(),
-    'assets',
-    'audio',
-    'audio.wav'
+    audioExtractStoragePath(),
+    `${project.id}.wav`
   );
 
   console.log('Started audio extraction');
 
-  const command = ffmpeg(absolutePathToMediaFile)
+  const command = ffmpeg(project.mediaFilePath)
     .audioChannels(1)
     .outputOptions('-ar 16000') // Sample rate of 16kHz
     .noVideo()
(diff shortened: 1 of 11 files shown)


kateahyde:
Awesome work Mutaz! Just a quick question (which is maybe looking too far ahead but if we can come up with a solution now I think it will save time down the road) - how will this work going forwards if there is more than one media file associated with a project? I haven't looked into exactly how that would work in terms of storage within the project file structure, but if the audio file name is based on the project ID I think that could be problematic for projects with multiple media files down the line as you wouldn't know which media file the audio is referring to. The only reason for mentioning it is it may mean any projects created with the MVP won't work in a later release, although not sure if that's actually a problem.

danfeltham (approved):
Love your work, keep it up King.

src/main/handlers/audioExtract.ts
```diff
@@ -1,22 +1,29 @@
...
+  }
+
+  mkdir(audioExtractStoragePath());
+
   const pathToSaveMedia = path.join(
-    process.cwd(),
-    'assets',
-    'audio',
-    'audio.wav'
+    audioExtractStoragePath(),
```
jhea0008, line 20: This change will break the hardcoded file path in transcriptionHandler.ts (line 70) which it uses to grab the audio file. We need to change the transcriptionHandler so the path is passed in or something

jhea0008 (changes requested):

src/main/handlers/audioExtract.ts
MutazAshhab: Nice catch, I wasnt able to spot this since i always had the audio.wav file! Here is the commit where I changed the method: https://github.com/patrickbrett/mlvet/pull/63/commits/4129bdb2e424bec62d06ec0a42db625973e0df24

You can find the rest of the commits which update the variable names and datatypes within the IPC related files in the next commits below.

jhea0008: Fantastic, that's a much better solution all round. Great work

jhea0008 (approved):

MutazAshhab:
@kateahyde 

>how will this work going forwards if there is more than one media file associated with a project?

If we have more than 1 audio file per project then this approach will not work.

To accommodate for more than 1 audio file we could create a directory under `extractedAudio` using the `uuid`. e.g. `extractedAudio/<project id>/audio1.wav` or `extractedAudio/<project id>/<media name>.wav`

I think this is enough for the MVP however, we could (easily) implement the method for multiple audio sources now but we just 1 one.

I should also mention that the current Project object is structured for 1 media file. Check out `mediaFilePath` it is a single value rather than an array. Which is what made me design this feature/task for 1 single media file rather than a collection of media files

kateahyde:
Very fair @Taz17 I wasn't sure what had been decided for the project object so thought it would be worth asking in case it was worth pre-empting the change. In some ways I think it'd be great if we could use an array/list so that it will barely require any changes to have multiple files, but if it's already been implemented in the project object and given no-one will actually be using the MVP and have existing projects that won't work with the updated version, I think it's reasonable to keep it to 1 for now.

kateahyde (approved):

PR #64, opened 2022-05-13
Update get file name method

MutazAshhab:
### Why did i make this change ?
- The current implementation in this PR is much simpler than that was before...
- I could have just use path.basename() but instead i did this whole OS thing.

### Side note:
- I was thinking of changing the `extractFileExtension` but thought it was not worth making an IPC for a some simple code. What are your thoughts?


chloebrett (approved):
Yeah that is a bit simpler.

Also (and this is not as much of a clean up as you've done, but would have also simplified): you can just use `path.delimiter` instead of switching based on the OS

PR #66, opened 2022-05-14
Feature/video stream express server

zwan0202:
Why is this change needed?
To add express server sends the video in bytes for video preview

What did you change?

- Created an express server
- Add an endpoint for get video in bytes with selected range


package.json
```diff
@@ -14,7 +14,8 @@
     "start:main": "cross-env NODE_ENV=development electron -r ts-node/register/transpile-only ./src/main/main.ts",
     "start:renderer": "cross-env NODE_ENV=development TS_NODE_TRANSPILE_ONLY=true webpack serve --config ./.erb/configs/webpack.config.renderer.dev.ts",
     "test": "jest",
-    "prepare": "husky install"
+    "prepare": "husky install",
+    "start-express": "nodemon ./src/main/expressServer.ts"
```
chloebrett: What's this used for? Looks to me like we're launching from the main process anyway, is there ever a need to start it separately?

src/main/expressServer.ts
```diff
@@ -0,0 +1,49 @@
...
+    };
+
+    res.writeHead(206, head);
+    file.pipe(res);
+  } else {
+    const head = {
+      'Content-Length': fileSize,
+      'Content-Type': 'video/mp4',
+    };
+    res.writeHead(200, head);
```
chloebrett: Nicee!!

src/main/main.ts
```diff
@@ -140,6 +141,8 @@ const createWindow = async () => {
     }
   });
 
+  require('./expressServer');
```
chloebrett: Use import syntax instead of require()

package.json
MutazAshhab: Here is what I have to say about this.

- `nodemon` is tool which helps the developer by automatically restarting the node application when file changes in the directory are detected.
    - Once the app is built the files will not be changing.
    - Starting the server via `scripts` is something we can do now in development, however when we build the app these scripts will not be there and thus the server will not be running.
    - The solution to this is to start the server using a childprocess. Look at the how the python server is ran here: [main/pyServer.ts - line 4](https://github.com/patrickbrett/mlvet/blob/develop/src/main/pyServer.ts#L4)
    - One thing I recommend is using the [cwd option in the child_process.spawn function](https://nodejs.org/api/child_process.html#child_processspawncommand-args-options) (I have not yet updated the code to use it. Changing directories is not the correct way to go about it)
    - When you start using child_process.spawn you'll have to package the code into a function and use that instead of running the entire file

MutazAshhab (changes requested):
Now that you are using `yarn` you can get rid of the `package-lock.json` since that is used by npm

package.json
chloebrett: @Taz17 do we need a child process though? Unless I'm missing something we should just be able to start the server in the same process as the electron back end which will be easier to manage (no need to shut it down, health check it, etc)

MutazAshhab: I remember having a chat about that and it was mentioned that running the server through the electron/nodejs backend was buggy.

Just to be cautious, its better to have @zwan0202 @itsMoxMox to verify that they have tried running the server from the electron/nodejs backend and what the outcome was.

zwan0202: Sorry for the late reply.

I refactored the express server and let it run as a child process. The server runs fine and I can get response from it. 

But every time the project start, it will pop up an error window says 'Connection refused', not sure why. 

I put server and route file as standard `js` files, otherwise Node will found an `import` statement in a file that it does not consider an ECMAScript (ES) module. I don't want to set the module system at the project level. I tried to use the `.mts` and `.cts` extensions, but it not working and not showing any error or warning message.

chloebrett (changes requested):
Let's decide on an outcome regarding [this](https://docs.google.com/document/d/1B7iapmBAZp2Zfnuifp5gWCT8X7wDZdomvmGgmEM5WTk/edit) before we merge

PR #67, opened 2022-05-14
Format handlers consistently so that gencode can be applied

chloebrett:
This PR contains a bunch of preparatory work to make the handlers consistent, so that the code generator (which generates types, etc and is not in this PR) can be run.

Looks like a lot of line changes, but a lot of these are just reshuffling files, such as handlers that needed to be split up or moved to different parts of the codebase.

Requirements:
* Handlers must be declared using `const` syntax
* They must have separately declared types in the same file with the same name
* Their arguments must not use destructuring or default parameters
* Every handler must have its own file
* All files in the `main/handlers` folder will be exposed via IPC
* Each handler must be the default export in its file

These requirements are all documented in the gencode readme which is part of the gencode PR.

Other changes:
* Instead of simply injecting the `mainWindow`, handlers can access an `ipcContext` object which is passed in from `main/main.ts`
* 'START' and 'END' comments tell gencode where to inject generated code
* There is no `handlers/index` anymore

@hhen0005 and @Taz17 you two are the ones I'll be tagging in the gencode PR, so I've tagged you here as well.


ghost:
preProcess tests are failing on this branch. Working on develop. Is that intended?

src/main/main.ts
```diff
@@ -84,7 +84,11 @@ const createWindow = async () => {
     },
   });
 
-  initialiseIpcHandlers(mainWindow);
+  const ipcContext: IpcContext = {
```
ghost: Does `IpcContext` need to be imported here?

ghost:
Added @manikag-12 as this PR will affect her current export-file-menu PR

chloebrett:
> preProcess tests are failing on this branch. Working on develop. Is that intended?

Hm no, will investigate

> Added @manikag-12 as this PR will affect her current export-file-menu PR

Sure, happy to do the rework myself if hers gets merged first

src/main/main.ts
chloebrett: Yup, done thanks

chloebrett (commented):
Fixed the test, good catch. Was due to using app.getPath before app was ready

src/sharedUtils.ts
```diff
@@ -0,0 +1,28 @@
...
+  ];
+
+  const recentProject: Partial<RecentProject> = {
+    ...metadata,
+    projectFilePath: filePath,
+  };
+
+  // Programmatically assign fields from project
+  projectFields.forEach((field) => {
+    recentProject[field] = project[field] as any; // try to give this proper types, I double dare you
```
ghost, line 22: lol

ghost (approved):

ghost:
Great stuff! Approved

chloebrett:
Will wait until at least one other approval before merging

MutazAshhab (approved):

MutazAshhab (changes requested):
I just ran the tests and the App.test.tsx isnt passing.

image

chloebrett:
@Taz17 please pull latest and try again

I did fix some errors since your comment though I think the error you found was from an older commit

MutazAshhab (approved):
It passes all test cases now!

PR #68, opened 2022-05-14
Developer experience - IPC gencode

chloebrett:
Depends on: #67 

Adds a tool for auto wiring IPC stuff every time we edit the handlers in the back end.

Made this because I (and a few other people working with IPC) got pretty sick of having to update like 5 files every time we wanted to change a handler.

Design doc: https://docs.google.com/document/u/1/d/1pFvt4_HpjJYvx0w24IPfeKU7-NLqrl9eFwu0f2UOVpM/edit?usp=drive_web&ouid=114326772776807781082

Relevant XKCD https://xkcd.com/1205/ (where are we in this matrix? no idea)

Basically see the included readme for details

Run it: `yarn gen`

Also the whole thing is based off string matching stuff, no actual code or type parsing, but I've tested it and it works pretty consistently provided the handlers are formatted as it expects (again see the readme for this)

Regarding PR line count, can't really split this up any more, my apologies. Since this doesn't really interact with anyone else's code (it's in a completely separate 'tools' folder), and in the worst case we can revert to the old way of updating types ourselves if we have issues with it, I'm happy if you just want to treat the gencode logic as a black box and go "yup it works" or "no it doesn't". The alternative being to give a detailed analysis of my very ad hoc "look in a string until we find a semicolon, but only if we're not currently in a bracket, and also arrows don't count as brackets" logic 😉  although go for your life if you feel like it


tools/gencode.ts
```diff
@@ -0,0 +1,306 @@
...
+
+/**
+ * Load everything we need to know about all the IPC handlers we've written, so we can put info in the other files as necessary
+ */
+const extractHandlersMetadata: () => Promise<
+  GencodeIpcHandler[]
+> = async () => {
+  // Load list of file paths in handlers directory, skipping index and subfolders
+  const fileNames = (await fs.readdir(HANDLERS_DIRECTORY)).filter(
+    (name) => name.includes('.ts') && name !== 'index.ts'
```
MutazAshhab: Checking if name is equal to `index.ts` is no longer necessary since that file got removed in https://github.com/patrickbrett/mlvet/pull/67

MutazAshhab (changes requested):

tools/gencode.ts
chloebrett: Yep true 👍🏻  changed

MutazAshhab (approved):

PR #69, opened 2022-05-14
Feature/live processing

Vita1287:
This PR seeks to implement live processing so that the video previewer shows the correct time after each edit/delete, as requested by the export team.

**What was changed in this PR?**
liveProcess.ts created under editDelete. This file contains the functionality for live processing.
liveProcess.test.ts created under editDelete/__tests__. Tests are included for both deletion and editing.


src/main/editDelete/liveProcess.ts
```diff
@@ -0,0 +1,42 @@
...
+ * calculateTime calculates the outputStartTimes of a word based on the
+ * current state of all words in a Transcript. It takes into account edits/deletes
+ * @param word word in Transcript
+ * @param i the current index position in the words array
+ * @param words an array of words from Transcript
+ * @returns the input word with an updated outputStartTime
+ *
+ */
+
+const calculateTime = (word: Word, i: number, words: Word[]): Word => {
```
LiamTodd: Great stuff, would be good to add some comments in this function since the logic is quite complex and not super easy to grasp

LiamTodd (commented):
Great work on the test cases - they're easy to understand and are all passing. Might be worth adding some more test cases for edge cases/more complicated series of operations on the transcript in future.

Vita1287:
Yeah I could add some test cases combining edit and delete, just kinda ceebs

src/main/editDelete/liveProcess.ts
```diff
@@ -0,0 +1,42 @@
...
+  return word;
+};
+
+/**
+ * Processes a Transcript so that all words have the correct outputStartTime
+ * @param transcript The original Transcript object
+ * @returns The Transcript object with updated outputStartTime for all words
+ */
+const liveProcessTranscript = (transcript: Transcription): Transcription => {
+  transcript.words.map(calculateTime);
```
chloebrett: Rather than mutating each word, how can you use `map` to produce a new transcript with the changed words?

src/main/editDelete/liveProcess.ts
```diff
@@ -0,0 +1,42 @@
...
+ * current state of all words in a Transcript. It takes into account edits/deletes
+ * @param word word in Transcript
+ * @param i the current index position in the words array
+ * @param words an array of words from Transcript
+ * @returns the input word with an updated outputStartTime
+ *
+ */
+
+const calculateTime = (word: Word, i: number, words: Word[]): Word => {
+  if (!word.deleted) {
```
chloebrett: If you flip this and do an early exit, the rest of the logic will be nested one level less deep which will make it easier to read

src/main/editDelete/liveProcess.ts
```diff
@@ -0,0 +1,42 @@
...
+    if (i === 0) {
+      word.outputStartTime = 0;
+    } else {
+      let j = i - 1;
```
chloebrett: Like Liam said, a comment would be great here. Might be good to give j a descriptive name as well If you can.

src/main/editDelete/liveProcess.ts
```diff
@@ -0,0 +1,42 @@
...
+      while (j > -1 && words[j].deleted) {
+        j -= 1;
+      }
+      if (j === -1) {
+        word.outputStartTime = 0;
+      } else {
+        word.outputStartTime = words[j].outputStartTime + words[j].duration;
```
chloebrett: Notice that you're only ever updating the outputStartTime, so how could you redesign this function to just return the outputStartTime you want?

chloebrett (commented):
Looking good, just gave a couple of suggestions on how you might make this a little cleaner. Feel free to ask if you need further clarification

src/main/editDelete/liveProcess.ts
v1ta1287: Good point. liveProcessTranscript now returns a new Transcript object.

src/main/editDelete/liveProcess.ts
v1ta1287: Conditional statements definitely easier to read now with this change. Thanks

src/main/editDelete/liveProcess.ts
v1ta1287: j has been renamed to nextNotDeleted, comments have been provided explaining the algorithm.

src/main/editDelete/liveProcess.ts
v1ta1287: This is the only change request I'm not 100% sure about. I've refactored calculateTime to return the updated outputStartTime, but I now have to call another function processWord for the mapping. If there's any better suggestions please let me know and I'll try implement it.

src/main/editDelete/liveProcess.ts
v1ta1287: Comments have been added to calculateTime

v1ta1287:
Test case has been added which includes both edit and delete operations

LiamTodd (approved):

src/main/editDelete/liveProcess.ts
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Processes a Transcript so that all words have the correct outputStartTime
+ * @param transcript The original Transcript object
+ * @returns The Transcript object with updated outputStartTime for all words
+ */
+const liveProcessTranscript = (transcript: Transcription): Transcription => {
+  return {
+    confidence: transcript.confidence,
+    words: transcript.words.map(processWord),
```
chloebrett, line 59: Nice work! You can remove the need for `processWord` like this:

`words: transcript.words.map((word, i, words) => ({ ...word, outputStartTime: calculateTime(word, i, words), })),`

Or make `return { ...word, outputStartTime: calculateTime(word, i, words), }` the body of processWord, but with proper line breaks probably not necessary

chloebrett (approved):

ghost (approved):

rileykeane (approved):
Looks great!

PR #70, opened 2022-05-14
Chore/local tests

chloebrett:
Basically, just kept running `yarn test` and fixed whatever complaint it had until everything was working.

Notable things:
* StoreChangeObserver has some async logic that needs to be `waitFor`'d otherwise testing-library says "hey you should wrap that in an act()" even when you already did.
* Unfortunately, the MUI Modal component does not work at all with `waitFor` so they can't be tested in the same test.
* Therefore, I have made two separate tests for the main App render - one for everything except StoreChangeObserver, and one for StoreChangeObserver.
* IPC stuff is just mocked as returning empty lists etc for now, also I had to move some of the IPC things on the front end around
* Tests are a bit stricter than our linter so stuff like "h-300" which isn't a valid style anymore makes it throw an error, I fixed those types of problems when I found them

Acceptance criteria: should be able to run `yarn test` and everything passes

Tagging Thien, Andy and Liam as all on architect team and haven't given you all any PRs to review yet

### DIFF
--- a/src/renderer/__tests__/App.test.tsx
+++ b/src/renderer/__tests__/App.test.tsx
@@ -1,0 +1,17 @@
+import '@testing-library/jest-dom';
+import { render } from '@testing-library/react';
+
+import App from '../App';
+
+describe('App', () => {
+  /**
+   * We test StoreChangeObserver separately from the rest of the App component as it does some async logic
+   * (loading recent projects from IPC). This can be worked around with 'waitFor' but unfortunately MUI's Modal
+   * component breaks when we do that. So for now we test separately
+   */
+  it('should render successfully without wait when StoreChangeObserver is disabled', async () => {
+    const rendered = render(<App hasStoreChangeObserver={false} />);
+
+    expect(rendered).toBeTruthy();
+  });
+});
--- a/package.json
+++ b/package.json
@@ -124,9 +124,6 @@
     ],
     "testPathIgnorePatterns": [
       "release/app/dist"
-    ],
-    "setupFiles": [
-      "./.erb/scripts/check-build-exists.ts"
     ]
   },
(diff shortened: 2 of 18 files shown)


tngu0041 (approved):

src/renderer/__tests__/App.test.tsx
```diff
@@ -0,0 +1,17 @@
...
+describe('App', () => {
+  /**
+   * We test StoreChangeObserver separately from the rest of the App component as it does some async logic
+   * (loading recent projects from IPC). This can be worked around with 'waitFor' but unfortunately MUI's Modal
+   * component breaks when we do that. So for now we test separately
+   */
+  it('should render successfully without wait when StoreChangeObserver is disabled', async () => {
+    const rendered = render(<App hasStoreChangeObserver={false} />);
+
+    expect(rendered).toBeTruthy();
```
LiamTodd, line 15: Not a critique but a question: How does toBeTruthy() work?

LiamTodd (approved):

src/renderer/__tests__/App.test.tsx
chloebrett: It converts the value to a boolean and returns whether or not the result is true

e.g.

"hello", 6, true, [1,2,3], { key: "val" } all become true

and

"", [], 0, null, false, undefined all become false

So what it's checking is if the app render produces something meaningful other than just null/undefined basically

chloebrett: https://developer.mozilla.org/en-US/docs/Glossary/Truthy

truthiness is a gross javascript thing that causes lots of problems haha

Vita1287 (approved):
Tests are working for me now.

PR #71, opened 2022-05-14
Chore/ci tests

chloebrett:
Depends on #70 

Adds a CI test step which will run on every push. We get 33 free build hours a month and current the CI takes 1 minute so unless we average more than 60 pushes per day (idk we might get close) then we should be fine within the free tier.

Worst case we can make it only when being merged into develop or something, but I think it's fine for now

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -22,4 +22,5 @@ jobs:
           node-version: ${{ matrix.node-version }}
           cache: 'npm'
       - run: npm i -g yarn
+      - run: yarn
       - run: yarn test
--- a/src/renderer/components/ExportCard.tsx
+++ b/src/renderer/components/ExportCard.tsx
@@ -18,7 +18,7 @@ interface Props {
 const ExportCard = ({ progress }: Props) => {
   return (
     <Card>
-      <Typography variant="body1" color="primary.main">
+      <Typography variant="h3" color="primary.main">
         Exporting...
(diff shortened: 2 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,26 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node-version }}
+          cache: 'npm'
+      - run: npm i -g yarn
+      - run: yarn
+      - run: yarn test
```
LiamTodd, line 26: Cool to see 2107 stuff in action

LiamTodd (approved):

src/renderer/components/ExportCard.tsx
```diff
@@ -18,7 +18,7 @@ interface Props {
 const ExportCard = ({ progress }: Props) => {
   return (
     <Card>
-      <Typography variant="h-300" color="primary.main">
+      <Typography variant="body1" color="primary.main">
```
ghost: Thanks for the fix. But please change to h3.

chloebrett: Ah yes

chloebrett: Changed

PR #72, opened 2022-05-15
Feature/file representation

chloebrett:
~~Depends on #60~~

- Represents the open file (I think this might only work on Mac)
- Sets file as 'dirty' when it's been edited
- Prompts the user to save if they try to quit with a dirty file

Was a nice test run of gencode, it definitely made the dev experience way nicer

**Acceptance criteria**

- All OS's: editing a file and then trying to close it gives a prompt with save, don't save, cancel
- All OS's: "save" in prompt saves the file and then exits
- All OS's: "don't save" in prompt exits without saving
- All OS's: "cancel" aborts the exit operation and the app stays open
- All OS's: can see open file name in the top bar.
- Mac: file is marked as 'dirty' (round circle in the red traffic light button) when it's been edited. When you save or save as, it goes back to un-dirty
- Mac: can option+click or cmd+click the top bar to see directory structure of the open file (i.e. its parent dir, the parent dir of its parent dir, etc up until the root dir)


Irmy:
Doesn't seem to work on windows OS - no menu option to exit and windows close (X) closes entire application immediately.

rileykeane:
There seems to be an issue eveytime I save a file and the window closes without the application quitting. 

All these cases lead to the same issue:
1. Close files without saving -> save (don't quit)
2. Close files without saving -> don't save (don't quit)
3. save file then close (don't quit)

After doing any of the above cases, when I open any file I get this error and can no longer save at all. Via the menu option or being asked to save on close:
Screen Shot 2022-05-19 at 1 18 40 PM

**Note: Quitting and opening again resolves the issue.**

chloebrett:
@rileykeane I think this just points to the fact that for now we should quit the app when the window closes, as we only support a single window anyway. I have updated it to do this. We can change this later if/when we support multiple windows

chloebrett:
@Irmy please give it another try on Windows now, had to make some pretty big changes to get it working actually (isEdited, window title and represented filename not only don't work on windows, but don't even save to the window state, so I had to build a custom state class)

chloebrett:
Okay, I've been playing around with this and found a bug.

It seems on Windows, the project's listing in the "recent projects" gets duplicated when you save via the "save" dialog that appears when you try to quit without saving. I'm not sure why this happens and haven't been able to replicate it on Mac, though I haven't tried on Linux.

On mac, a different bug occurs: the "recent projects" list actually disappears entirely. This is due (to the best of my understanding) to the fact that the recent projects file is being written when the app quits, and so the operation doesn't finish and the file is left in a blank state.

Obviously not ideal, but avoidable for the sake of demo by making sure you save _before_ hitting the red traffic light or close button. Compared to the previous state of affairs: if you clicked close without saving, the project would simply not save at all.

I've thought about how to resolve these issues and I think the solution will be too involved to get done before the demo, and it's not a priority feature for us to demonstrate. However, showing the open file and whether it's been edited is quite a nice UX, so I'd like to suggest this gets merged and we deal with the bug I just mentioned in a separate PR.

Irmy (approved):
I agree getting it bug free by MVP is too close - just add the known issues as a git issue so other can see it incase they don't read your comment. 
If a Linux user can show what errors they get on their end that would be good before merge - just so we know if it causes any big breaks on their OS.

rileykeane:
Looks like the error I previously posted has been fixed, however I'm now getting some weird behavior with the recents page.

I can't actually create new projects on my computer and have to use a .mlvet file to get to the project page. I could be using an old schema version. I have attached the file I'm using for testing. 

Anyway, two things are happening. 

1. If I open the file - it appears in my recents. Great! However if I close MLVET and open the file again via File --> Open and make a change and save, the project appears twice in recent. (and three times if I do it again and so on...)

2. If I open a project via File --> Open and retart MLVET **all** my recents open the previously opened project. I would assume this doesn't happen if you create new projects in the app (as opposed to opening a .mlvet file) and then only open recent projects via the recent projects section - but as I said, I can't create new projects ☹️ (without hacking the code to manually do it). 

**Files used for testing (converted to txt due to GitHub support)**
[sample.txt](https://github.com/patrickbrett/mlvet/files/8755236/sample.txt)
[sample2.txt](https://github.com/patrickbrett/mlvet/files/8755238/sample2.txt)

rileykeane:
> 1. If I open the file - it appears in my recents. Great! However if I close MLVET and open the file again via File --> Open and make a change and save, the project appears twice in recent. (and three times if I do it again and so on...)

Ok so I can't seem to reproduce this, but it definitely happened a few times. The second issue is happening consistently though.
 
Screen Shot 2022-05-24 at 12 21 57 AM

rileykeane:
> Ok so I can't seem to reproduce this, but it definitely happened a few times. The second issue is happening consistently though.

Update - looks like this is happening on develop and not just this PR.

PR #73, opened 2022-05-15
Feature/delete project

chloebrett:
~~Depends on #72 (and hence a bunch of others)~~

~~Will tag and add a proper description when those are merged~~

To test:
* make some recent projects
* delete them


Irmy:
Sent you the video of screen record - doesn't seem to persist delete on restart - not sure if its because of my broken flask server?

chloebrett:
Good find, there was a bug in the store change observer so when you delete the only recent project it doesn't write to the file. Fixed, please test again

Irmy (approved):
Works well on my end. Project created and deleted and upon restart project has been deleted .

PR #74, opened 2022-05-16
delete postProcess.ts as it is not needed

LiamTodd:
What is the purpose of this PR?  
To remove postProcess.ts, as the logic contained in this class is carried out by transcriptToCuts.ts. The double-up occurred to a communication lapse between agile teams, so this PR is just here to amend that.

What was changed in this PR?  
postProcess.ts is deleted and the Clip interface removed from sharedTypes.ts


Vita1287:
Some impressive stuff here Liam. Keep it up.

Vita1287 (approved):
None of the deleted code affects any of the other back-end code for editDelete. All good!

itsMoxMox (approved):

PR #76, opened 2022-05-18
Feature/express server in main

zwan0202:
Why is this change needed?
To add express server sends the video in bytes for video preview. Start express server in the main process.

What did you change?
- Created an express server
- Add an endpoint for get video in bytes with selected range


src/expressServer/routes.ts
```diff
@@ -0,0 +1,52 @@
+import express from 'express';
+import fs from 'fs';
+
+const router = express.Router();
+
+router.get('/', (_, res) => {
+  res.send('Server Running');
+});
+
+const streamVideo: (req: any, res: any) => void = (req, res) => {
```
chloebrett, line 11: Add a todo to type the request and response

chloebrett (approved):
Does this replace https://github.com/patrickbrett/mlvet/pull/66? If so please close it

Other than that, looks good

package.json
```diff
@@ -210,9 +210,11 @@
     "electron-debug": "^3.2.0",
     "electron-log": "^4.4.6",
     "electron-updater": "^4.6.5",
+    "express": "^4.18.1",
     "fluent-ffmpeg": "^2.1.2",
     "get-audio-duration": "^3.1.0",
     "history": "^5.3.0",
+    "nodemon": "^2.0.16",
```
MutazAshhab: nodemon is a dev dependency, Could you move it from `dependencies` to `devDependencies`

chloebrett:
Actually just saw your commit moving nodemon to devDependencies. Why do we need it at all?

zwan0202:
> Actually just saw your commit moving nodemon to devDependencies. Why do we need it at all?

It can automatically restarting the node application when file changes in the directory are detected. I'm not the one who add this, I guess it's just for convenience

chloebrett:
> It can automatically restarting the node application when file changes in the directory are detected. I'm not the one who add this, I guess it's just for convenience

As far as I can tell it isn't linked to our start command, so I don't think it will do anything

Also, I only commented because it's an addition in the PR. It wasn't a dependency before. If we're not using it, no need to include

itsMoxMox (approved):

PR #78, opened 2022-05-18
Feature/python node interface

kateahyde:
The code in this PR allows our node code to communicate with the python process so that it can obtain a transcript of the audio extracted from the provided video file.

**What was changed in this PR?**
The code in app.py was changed so that it runs as a socket server on the Flask server (rather than a regular API on the Flask server). transcriptionHandler.ts was modified to call and obtain the transcription. socketServer.py was initally used for a spike of the socket, then was used for debugging the socket as it was run externally from the application. I'm thinking it should probably be deleted as it isn't actually being used and its functionality is almost entirely covered by app.py but thought I'd check first.

**Other notes/questions**
I'm not sure if/how automated tests can be written for this... if you have any suggestions or ideas (or if you think it's ok to not have tests written for it) please let me know!


jhea0008 (changes requested):
.wav Audio file probs shouldn't be checked in unless there's a good reason for it

src/main/handlers/transcriptionHandler.ts
```diff
@@ -18,6 +19,21 @@ interface JSONTranscriptionContainer {
 const sleep: (n: number) => Promise<void> = (n) =>
```
jhea0008: This function can be deleted now that it's not used

src/pyServer/socketServer.py
```diff
@@ -0,0 +1,29 @@
+from flask import Flask
+from flask_socketio import SocketIO
+from transcription import transcribe
+
+app = Flask(__name__)
+socketio = SocketIO(app)
+
+
+@socketio.event
```
jhea0008: There seems to be a significant chunk of overlap between this file and app.py. Could we compress them into a single file?

src/main/handlers/transcriptionHandler.ts
```diff
@@ -57,15 +73,9 @@ const requestTranscription: RequestTranscription = async (project) => {
...
-  // Read from sample transcript. Replace this section with real transcript input
-  const transcriptionPath = app.isPackaged
-    ? path.join(process.resourcesPath, 'assets/SampleTranscript.json')
-    : path.join(__dirname, '../../../assets/SampleTranscript.json');
-
-  const rawTranscription = fs.readFileSync(transcriptionPath).toString();
-  const jsonTranscript = JSON.parse(rawTranscription);
+  const transcript = await transcribeRequest(project);
+  const jsonTranscript = JSON.parse(transcript);
+  console.assert(jsonTranscript.transcripts.length === 1); // TODO: add more error handling here
```
chloebrett: You can remove this as our runtime type checks are handled by the line below (validateJsonTranscriptContainer)

src/pyServer/socketServer.py
kateahyde: Yeah that's why I asked if I should just delete it - it's only really useful for debugging now and isn't in use so I thought should be deleted but wanted to check

chloebrett:
Got it working and it's really cool to see it all together! I tested the transcription with a pitch I did for another unit, the transcription it came up with is pretty funny at times 😅 

Screen Shot 2022-05-18 at 7 29 56 pm

To be fair though, this was a poorly recorded Zoom audio and I was speaking quite fast, so for slower speech recorded with a proper mic it will probably be more accurate

src/pyServer/socketServer.py
jhea0008: Fine by me to delete if it doesn't serve any purpose that the other file doesn't

jhea0008 (approved):

src/main/handlers/transcriptionHandler.ts
```diff
@@ -15,8 +16,21 @@ interface JSONTranscriptionContainer {
  * @param n seconds to sleep
  * @returns promise resolving after n seconds
```
MutazAshhab: Now that the sleep() function is gone, could you remove the comment?

src/pyServer/app.py
```diff
@@ -1,18 +1,32 @@
...
+def connect():
+    print('connect')
+
+
+@socketio.on('transcribe')
+def transcribeHandler(audio_file_path):
+    audio_file_path = "\"%s\""%audio_file_path
+    print('transcribe', audio_file_path)
+    return transcribe(audio_file_path)
+
```
MutazAshhab: Can you add a comment explaining what line 21 does?

From what we talked about this is probably for windows only, right? Is this check necessary for when you are MacOS/linux ?

src/pyServer/socketServer.py
MutazAshhab: Before merging to develop make sure that `socketServer.py` is the file being deleted not `app.py`.

The reason why we want to keep app.py is because the command `flask run` automatically looks for a `app.py`

MutazAshhab (changes requested):

src/pyServer/app.py
tngu0041: Yeah, could be worth adding a comment for this. Since it's not obvious at a glance what it does.

@Taz17 Wrapping the file path in "" quotes fixes the issue caused by having spaces in the filepaths in Windows, yeah. I have no idea if this is still an issue in MacOS/Linux since both Kate and I use Windows, and we only tested it on Windows.

If anyone with macOS/Linux wants to test this on their machine though, you can just edit the test I wrote in test_transcription.py to take in a file path that has spaces and see if it still passes or fails.

src/pyServer/socketServer.py
kateahyde: Yep all g I'll delete it now, definitely removing socketServer.py not app.py

src/main/handlers/transcriptionHandler.ts
kateahyde: Assuming you just meant `console.assert(jsonTranscript.transcripts.length === 1);` right? I think lines 76 and 77 are required for the validateJsonTranscriptContainer line

src/main/handlers/transcriptionHandler.ts
kateahyde: Sorry missed that, done

MutazAshhab (approved):

PR #79, opened 2022-05-18
Feature/video preview controller

ghost:
Ok, big PR. Apologies for this, trying to get a working video preview in early for testing.

**What features does this include?**
Video preview UI, video preview & timer functionality, linking functionality to the UI written by @Aretle.
All the things that should be working:
- Video Preview (play, pause, skip, onWordClick)
- Timer UI

**What features doesn't this include?**
The express server connection. Will be handled in a later PR once the express server is merged in. Shouldn't be too much linking work for FE and BE.

**What did you change?**
Ok lets break down this down:
- Added a new package ([vimond-replay](https://vimond.github.io/replay/#/)) - this is the video preview UI
- Moved some of the export utility functions so that they could be reused for formatting the timer for the UI
- Reworked _convertTranscriptToCuts_ function
- Partially refactored the video controls UI (_VideoController.ts_) to pass in play, pause, skip functions
- Added the _VideoPreview_ component = the video preview UI
- Added the _VideoPreviewController_ component = the video preview & timer functionality
- Linked up _onWordClick_ with _VideoPreviewController_

**How to test**
Note: You will need to have a video file at `experiments/video-preview/assets/sample.mp4`
In terminal 1:
`cd experiments\video-preview\`
`yarn`
`yarn start-server`
In terminal 2:
`yarn`
`yarn start` (as usual)

**Things to be aware of**
I have used useImperativeHandle twice in this PR. This react built-in hook should not be used without careful consideration. Basically, it allows you to call methods/functions exposed by a child component in a parent component. This is not the common pattern for react. The common pattern is passing in callbacks from the parent which are called by the child (like a button onClick callback).
I feel it is justified here. I wanted to abstract the video preview UI logic and timer functionality by putting that logic in their own components (VideoPreview & VideoPreviewController). To do that, we had to be able to call functions in the child from the parent. This is very similar to how the standard HTML `<video>` element works with its methods `play()`, `pause()`, etc.


ghost:
I don't think that it works too well with deleting while playing. However, this PR is already quite large so will address in another PR.

src/main/timeUtils.ts
```diff
@@ -0,0 +1,42 @@
...
+};
+
+// 00:00:00:00
+export const secondToTimestamp: (num: number) => string = (num) => {
+  return [3600, 60, 1, 0.01]
+    .map((mult) => padZeros(integerDivide(num, mult), 2))
+    .join(':');
+};
+
+export const secondToTimestampUI = (time: number): string => {
```
chloebrett, line 14: What's the difference between the output of this and secondToTimestamp?

chloebrett (commented):
Looking decent from a cursory look over. Will give a thorough review later today

src/main/timeUtils.ts
ghost: Here are some blackbox outputs:
image

I definitely understand what you're getting at. They seem to be very similar and have shared logic. 
I can change the code to use the `secondToTimestamp` function inside of the `secondToTimestampUI` function. However, I'm not sure if the secondToTimestamp function is currently working correctly and it would involve slicing the output string from variable points. Thought it would be less error prone to write some new logic.

chloebrett: Yeah I guess it's fine, it's not really a blocker. It's just that the two functions do extremely similar things but they are written in very different ways.

chloebrett:
Is this expected? If so how do I test locally?

Screen Shot 2022-05-19 at 7 46 25 pm

src/main/timeUtils.ts
ghost: I will resolve for now as it turns out that the other function `secondToTimestamp` (used for export) is not working correctly and will need to be rewritten anyway by @itsMoxMox . I have suggested to him that he combine the logic where possible to avoid duplication.

itsMoxMox: Good News! It's done

https://github.com/patrickbrett/mlvet/pull/83

ghost:
So there is a problem where when you create a cut (by deleting a word) and then click on a word after that deleted word, then the video will not play from the word but rather from the start of the cut. The system time (at the top) however is not affected. Hence, I believe this section of code is the cause:
image

This problem will be addressed in a future PR but cannot be fixed for now. The reason for this is that the video plays off of the cuts and the cuts are created from the transcript as stored in redux. Currently, the transcript stored in redux is not being changed when a user deletes a word (other than hiding the word).

So, for now, going to leave this bug and work with @rileykeane and @Aretle to update the transcript with live processing and then see if the bug still exists.

danfeltham (approved):
Excellent work. Definitely some top tier programming.

Some bugs for future fixes: 
- The timer moves slow. (more noticeable in longer videos)
- The video ends at different times according to the timer on different playbacks

Would also be nice to see some more comments in the code. Keen to see the PR that updates the video playback when an edit is made.

Future developer problems I reckon. Great job.

rileykeane (approved):
Thats some sexy maths

PR #80, opened 2022-05-19
Feature/drag media import

Irmy:
What is the purpose of this PR?
To allow users to import files through a dragging functionality.

What features does this include?
Drag import and UI response to communicate drop zone

What did you change?
UI elements of SelectMediaBlock and corresponding event handlers.


fwang1703 (approved):
Nice! LGTM

PR #81, opened 2022-05-19
Feature/return to home

fwang1703:
**Why is this change needed?**
Currently there is no way for a user to return back to the home page (*legally* - the current 'Reload' menu button is pretty hacky and will most likely get removed/changed in the future). This new menu button will bring a user from the Project page to the Home page, and will also ask the user to confirm any unsaved changes (if there are any). 

image

image

**What did you change?**
- New button to the menubar under 'History' -> 'Home'
- A new `returnToHome` handler
- A new event `initiate-return-to-home` which triggers when Home menu button is pressed/keyed 

**Future enhancement**
Currently, the confirm changes dialog only appears if the user has not saved the project at all. This is because to check for unsaved changes for a saved file, it depends on checking for 'dirty' files (not implemented yet).


src/main/menu.ts
```diff
@@ -224,6 +224,18 @@ export default class MenuBuilder {
         },
       ],
     };
+    const subMenuHistory: MenuItemConstructorOptions = {
```
chloebrett, line 241: Make sure it works on win and linux as well, this will only add the menu item on mac

src/renderer/ipcRegistration.ts
```diff
@@ -123,3 +123,28 @@ ipc.on('initiate-export-project', async () => {
...
+    const filePath = await window.electron.saveProject(currentProject);
+
+    const projectMetadata = await window.electron.retrieveProjectMetadata({
+      ...currentProject,
+      projectFilePath: filePath,
+    });
+
+    store.dispatch(projectSaved(currentProject, projectMetadata, filePath));
+  }
+  store.dispatch(pageChanged(ApplicationPage.HOME));
```
chloebrett, line 153: I think you also need to dispatch a projectClosed (or whatever it's called) action to set the currentProject to null in the store, otherwise the currentProject will still be set which has consequences for save buttons, file representation etc

src/main/handlers/helpers/saveUtils.ts
```diff
@@ -39,3 +39,28 @@ export const saveProjectToFile: (
 
   await writeFile(filePath, projectAsString);
 };
+
+export const confirmSave: (
```
chloebrett, line 43: Up to you if you want to do this now or in a follow up PR, but I've got the 'dirty' and 'confirm save' dialog working in this PR: https://github.com/patrickbrett/mlvet/pull/72 so you could pull that in and reuse those for your feature, or wait til it's in dev so that this PR isn't blocked. Or, if you like, I could make the follow up when both PRs (this one and mine) are in dev

src/main/menu.ts
fwang1703: Anyone want to be the guinea pigs for this? @hhen0005 @Taz17

chloebrett: Haha Mutaz Linux moment 💀

 I have a Linux laptop I’ll do a quick test now

src/renderer/ipcRegistration.ts
fwang1703: Updated!

src/main/menu.ts
chloebrett: image

This is what it looks like - just no history menu at all. See the file and edit menus for the pattern on how to make it appear in both OS's.

Also look into disabling the option when the user is not in a project - as doesn't make much sense to go home when not already home. The pattern for that is also described in menu.ts I believe

src/main/handlers/helpers/saveUtils.ts
fwang1703: Happy to refactor in a follow up PR

src/main/menu.ts
fwang1703: Just updated for it to (hopefully) work on windows and linux too. Will look into the disabling soon

fwang1703: Just added the disabling of the menu option when in home screen

danfeltham (approved):
Faye you're a weapon. Works like a charm on my Windows computer. Clean af, I couldn't break it. You already know that it needs to check for dirty files but that's future dev problems. Awesome work, thankyou!

MutazAshhab:
Hey @fwan0018, I tested it out and it works, I can see the `Home` option in the history drop down. The shortcut also works. 

All on linux btw!

PR #82, opened 2022-05-19
Feature/mvp punctuation

LiamTodd:
To punctuate the transcription so it looks nice in the demonstration next week.
Be aware that this method of punctuating the transcript is (almost certainly) going to be replaced after the MVP, so this PR is meant to implement something which is just 'good enough'.

**What is changed in this PR?**
Added in a step of punctuating the Transcript as part of preProcess. Commas and full stops are concatenated with words which have a silence after them which reach a certain threshold value. The thresholds are calculated as:     
lowerCommaThreshold: averageSilenceDuration * 0.85,
upperCommaThreshold: averageSilenceDuration * 1.8,
The constants 0.85 and 1.8 are determined through trial and error.
Additionally, there is capitalisation after full stops, and for the words "I'm" and "I'.
Capture

### DIFF
--- a/src/main/editDelete/__tests__/preProcess.test.ts
+++ b/src/main/editDelete/__tests__/preProcess.test.ts
@@ -38,7 +38,7 @@ describe('Test pre-processing JSON transcript into regular transcript', () => {
           fileName: 'PLACEHOLDER FILENAME',
         },
         {
-          word: 'Abc',
+          word: 'Abc.',
           duration: 1,
           startTime: 0,
           outputStartTime: 0,
@@ -56,7 +56,7 @@ describe('Test pre-processing JSON transcript into regular transcript', () => {
           fileName: 'PLACEHOLDER FILENAME',
         },
         {
-          word: 'def',
+          word: 'Def.',
           duration: 1,
           startTime: 1,
           outputStartTime: 1,
@@ -137,7 +137,7 @@ describe('Test pre-processing JSON transcript into regular transcript', () => {
           fileName: 'PLACEHOLDER FILENAME',
         },
         {
-          word: 'From.',
+          word: 'From',
           duration: 3,
           startTime: 8,
           outputStartTime: 8,
@@ -155,7 +155,7 @@ describe('Test pre-processing JSON transcript into regular transcript', () => {
           fileName: 'PLACEHOLDER FILENAME',
         },
         {
-          word: 'Fire.',
+          word: 'fire.',
           duration: 0.9,
           startTime: 11.2,
           outputStartTime: 11.2,
--- a/src/main/editDelete/__tests__/preProcess.test.ts
+++ b/src/main/editDelete/__tests__/preProcess.test.ts
@@ -38,7 +38,7 @@ describe('Test pre-processing JSON transcript into regular transcript', () => {
           fileName: 'PLACEHOLDER FILENAME',
         },
         {
-          word: 'abc',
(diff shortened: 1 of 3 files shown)


src/main/editDelete/punctuate.ts
```diff
@@ -0,0 +1,64 @@
...
+    const silenceDuration = endTime - word.startTime - word.duration;
+
+    if (index === 0) {
+      words[index].word = capitalizeFirstLetter(words[index].word);
+    }
+
+    const thresholds: Thresholds = calculateThresholds(averageSilenceDuration);
+
+    let punctuation: string;
+    if (silenceDuration < thresholds.lowerCommaThreshold) {
```
LiamTodd, line 36: I wanted to use a nested turnery here but ESLint said no:(

src/main/editDelete/punctuate.ts
```diff
@@ -0,0 +1,64 @@
+import { MapCallback } from '../util';
+import { Word } from '../../sharedTypes';
+
+type PartialWord = Pick<Word, 'word' | 'startTime' | 'duration'>;
+type Thresholds = { lowerCommaThreshold: number; upperCommaThreshold: number };
+
+// // TODO: use silence averaging to implement punctuation
+// const lowerCommaThreshold = 0.1;
```
chloebrett: Remove

src/main/editDelete/punctuate.ts
chloebrett: I know the let then assign pattern is a bit gross but nested ternaries can be kind of hard to read, I would agree with eslint here

src/main/editDelete/preProcess.ts
```diff
@@ -102,6 +103,21 @@ const addSpaces: (totalDuration: number) => MapCallback<Word, Word[]> =
...
+  let silenceSum = 0;
+  for (let i = 0; i < jsonTranscription.words.length - 1; i += 1) {
+    const endTime = jsonTranscription.words[i + 1].start_time;
+    const silenceDuration =
+      endTime -
+      jsonTranscription.words[i].start_time -
+      jsonTranscription.words[i].duration;
+    silenceSum += silenceDuration;
+  }
+  return silenceSum / jsonTranscription.words.length;
```
chloebrett: If the transcript is empty does this break the app from a div0?

src/main/editDelete/__tests__/preProcess.test.ts
```diff
@@ -38,7 +38,7 @@ describe('Test pre-processing JSON transcript into regular transcript', () => {
           fileName: 'PLACEHOLDER FILENAME',
         },
         {
-          word: 'abc',
+          word: 'Abc.',
```
chloebrett, line 41: Can you add a test that specifically tests for different punctuation types, I can't see any commas in the tests and I think we should have a punctuation specific test to make sure it's being tested explicitly

LiamTodd: Yep, no problem

src/main/editDelete/preProcess.ts
LiamTodd: Good pick up. I've replaced this with:
  return jsonTranscription.words.length !== 0
    ? silenceSum / jsonTranscription.words.length
    : totalDuration;
as the average silence duration just equals the total duration if the transcript has no words

kateahyde (approved):

PR #86, opened 2022-05-21
Feature/move text

danfeltham:
## Summary
You wanted the ability to move text around the transcript. I give you the ability to move text around the transcript. While I'm not stocked with how it is implemented 
<hr/>

It is a key feature of our final product.
<hr/>

### What did you change?
I added a few key bindings (x, c, v) for cut, copy, and paste. I made an operation for pasting text for the undo stack. This was the only op required because cutting is just a combination of deleting text which already exists as an op and pasting.

I also created a clipboard state in react which tracks the indexes of the currently copied block of text. This may not be ideal but it gets the job done and I'm lazy.
<hr/>

No tests because it will take too long to figure out how to do it by myself. Might be good to have a masterclass.
<hr/>

N/A
<hr/>

The `getIndexSelectedWords()` function is a little bit janky. If anyone is inspired to write a better one, the user experience will be greatly increased.
<hr/>

### OS

- [ ] Linux
- [ ] MacOS
- [x] Windows


chloebrett:
It's up to you whether you want to fix these now or later, as we're close to the deadline. But there are some bugs in your code:

1. If you cut and paste a sequence of words, some of the spaces between words disappear.

Before:

Screen Shot 2022-05-22 at 11 43 49 am

After:

Screen Shot 2022-05-22 at 11 44 24 am

2. If you delete words, then cut the words surrounding them and paste them elsewhere, the deleted words return.

Before:

Screen Shot 2022-05-22 at 11 45 06 am
 
After delete:

Screen Shot 2022-05-22 at 11 45 40 am

After paste:

Screen Shot 2022-05-22 at 11 45 58 am

3. If you delete words then cut them, then undo the cut but don't undo the delete, the deleted words return. If you redo the delete and then the cut, the state adjusts as expected. This indicates there might be a problem with your undo but not your do part of the cut op.

This one is a little harder to show in screenshots but if you follow the instructions you should be able to replicate it.

4. Copying words all over the place breaks undo. This might be linked to your known issue of indexes not working - but in theory this should be deterministic even with that bug. Below is the result of copying a bunch of times, undoing and then redoing, then undoing until there is no more left to undo. As you can see this doesn't return us to the original transcript, which should always be the result of undoing to the top of the stack. In fact in the state shown below, if I redo a bunch of times and then undo some more, _even more_ copies of "financial business" appear.

Screen Shot 2022-05-22 at 11 49 14 am

danfeltham:
@patrickbrett Good finds! Thankyou for taking the time to investigate and document those. I reckon those will be something fun to work on over the break and second semester. I don't intend to put any more time into the coding until some of my other assignments are finished.

Since it shouldn't break anyone else's code I'm still inclined to merge with develop and save those bugs for a future PR.

Cheers

PR #87, opened 2022-05-22
Link video preview FE with express server

Mhudlark:
## Summary

Links the video preview frontend to the express streaming server backend.

<hr/>

### What did you change?

Changed the `VideoPreviewController.tsx` to access the input media file path from redux and send it to the express server on the backend. Also, added the [`buffer` package](https://www.npmjs.com/package/buffer) (MIT license)

<hr/>

### Notes

I added a new package `buffer`. This package allows us to encode strings to base64 on the client side of the app (native Buffer only exists in Node.js backend). We use this to encode the media file path for the project so that it can be sent to the backend in the API URI.
There are other methods of encoding strings but most are hacky or deprecated.

<hr/>

### Testing

Just transcribe a video as usual, except now the video preview will also show that video.

### DIFF
--- a/package.json
+++ b/package.json
@@ -206,6 +206,7 @@
     "@fontsource/rubik": "^4.5.6",
     "@mui/icons-material": "^5.5.1",
     "@mui/material": "^5.5.1",
+    "buffer": "^6.0.3",
     "child_process": "^1.0.2",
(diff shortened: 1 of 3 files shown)


zwan0202 (approved):
Looks good! Have you tried it?

PR #90, opened 2022-05-23
Highlight words on playback

rileykeane:
## Summary

Highlights the word in the transcription that is currently playing. 

Please note, I do intended to animate the transition between words, however wanted to get this feature done for the MVP. For now the switch from one word to another feels a little janky. I plan to improve this by having the highlight slide over to the next word. 

I also had to make a tiny change to the PASTE_WORD and UNDO_PASTE_WORD ops to ensure the output start word time was updated - otherwise it caused some super weird behavior with the highlighting. 

<hr/>

This is feature helps the editor easily identify the word that's currently playing and enables them to follow along the transcript as the video is playing. 

The feature was ✨ _inspired_ ✨by an unnamed competitor. 

<hr/>

### What did you change?

Added a use effect hook that's dependent on the playback time. This will find the word thats currently playing and store its index in nowPlayingWordIndex. This is passed to the transcription block which will set the background color of the now playing word to orange 500 with a 50% opacity (same as the on hover color)

Also modified PASTE_WORD and UNDO_PASTE_WORD ops to return the processed transcription to ensure the new output start time was updated in store. 

<hr/>

Nope.

<hr/>

https://user-images.githubusercontent.com/27614346/169824053-713757a6-62cd-4bbc-b71a-dab4700a7480.mov

<hr/>

I had refactored this and moved the logic into the transcription block itself, however moved it back into the project page because having it in the transcription block means I need to pass `time` in as a prop. This would mean the transcription block would re-render every time the `time` value changes (which is a lot...) so I reverted the change. 

I'm not that experienced with react, so someone who is, please advise on the best approach.  

<hr/>

### OS

- [ ] Linux
- [x] MacOS
- [ ] Windows

### DIFF
--- a/src/renderer/pages/Project.tsx
+++ b/src/renderer/pages/Project.tsx
@@ -29,6 +29,9 @@ const ProjectPage = () => {
   // UI states
   const [time, setTime] = useState<number>(0);
   const [isPlaying, setIsPlaying] = useState(false);
+  const [nowPlayingWordIndex, setNowPlayingWordIndex] = useState<number | null>(
+    null
+  );
 
   const videoPreviewControllerRef = useRef<VideoPreviewControllerRef>(null);
 
@@ -78,6 +81,21 @@ const ProjectPage = () => {
     };
   });
 
+  useEffect(() => {
+    if (currentProject !== null && currentProject?.transcription !== null) {
+      const { words } = currentProject.transcription;
+      const newPlayingWordIndex = words.findIndex(
+        (word) =>
+          time >= word.outputStartTime &&
+          time <= word.outputStartTime + word.duration &&
+          !word.deleted
+      );
+      if (newPlayingWordIndex !== nowPlayingWordIndex)
+        setNowPlayingWordIndex(newPlayingWordIndex);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [time, currentProject?.transcription]);
+
   const onWordClick: (wordIndex: number) => void = (wordIndex) => {
     if (currentProject !== null && currentProject?.transcription !== null) {
       const newTime =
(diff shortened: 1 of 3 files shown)


chloebrett:
Task linked: [CU-2buvgk5 Highlight words on playback](https://app.clickup.com/t/2buvgk5)

rileykeane:
test

src/renderer/pages/Project.tsx
```diff
@@ -155,10 +156,22 @@ const ProjectPage = () => {
     };
   });
 
-  // TODO: figure out return type
+  useEffect(() => {
```
ghost, line 160: I understand for MVP but I reckon this should definitely be optimised if possible in a later PR. Please add a 'TODO: <comment-here>' comment to look into optimising this later

src/renderer/store/transcription/reducer.ts
```diff
@@ -61,10 +61,11 @@ const transcriptionReducer: Reducer<Transcription | null, Action<any>> = (
 
     // Have to check this if to get rid of linter error
```
ghost: No need for comment

src/renderer/store/transcription/reducer.ts
```diff
@@ -61,10 +61,11 @@ const transcriptionReducer: Reducer<Transcription | null, Action<any>> = (
...
     if (pasteyWords !== undefined && suffix !== undefined) {
```
ghost: Please update this to 'pasted words' - is this your spelling mistake? 😄

rileykeane: that is left over from @Aretle

ghost: 🚌

ghost (approved):
Nice work. Let's get it merged ASAP

PR #97, opened 2022-05-25
Fix video preview on first play jump bug - IMPORTANT

ghost:
IMPORTANT! Required for demo

## Summary

A bug fix for a bug that occurs when the user first plays the video. To view the bug (not on this branch). Wait a couple of seconds to play the video after the project page loads and the video will jump the amount of time you wait.

<hr/>

Needed for the demo.

### What did I change
Allowed the video preview to account for the first time playing when the video will not have been paused before.


Irmy (approved):
Looks good - Starts from the beginning on first play as well as if i go back to home and go back into project.

PR #128, opened 2022-07-29
Split handlers into multiple folders, update gencode to support this

### LINKED ISSUE #121: Refactor: split handlers into further folders
Handlers folder is getting quite large (21 handlers now)

Would be helpful to split these into further subfolders based on purpose - e.g. save-related things, media-related things, menu-related things, etc

Will also require a gencode refactor to make the code generator look for files in subfolders of the handlers folder (but exclude the 'helpers' folder). This is probably easier than it sounds

chloebrett:
Resolves #121


MutazAshhab (approved):
Good to merge :+1: v2

PR #130, opened 2022-07-29
Make timestamp in editor page monospace

### LINKED ISSUE #96: Monospace font for time tracker
This thing: Screen Shot 2022-05-24 at 7 03 46 pm

Should have a monospace font so that it doesn't change width when the time changes

https://user-images.githubusercontent.com/6735055/169994342-8790e8a9-d087-4c48-b451-6d4d3463bc5b.mov

Small UI thing and fixable in like 2 mins but will make a pretty big difference I reckon

chloebrett:
Currently it changes width depending on what numbers are showing which is a bit janky. This sets the timestamp to use roboto mono instead.

Before:

https://user-images.githubusercontent.com/6735055/169994342-8790e8a9-d087-4c48-b451-6d4d3463bc5b.mov

After:

https://user-images.githubusercontent.com/6735055/181698395-b081789d-2e43-42b9-8fb5-f5cb3527ec81.mov

Resolves #96


tngu0041 (approved):

src/renderer/App.css
```diff
@@ -3,6 +3,8 @@
  *        See https://github.com/webpack-contrib/sass-loader#imports
  */
 
+@import url('https://fonts.googleapis.com/css2?family=Roboto+Mono:wght@500&display=swap');
```
ghost: I would usually install a font as a package from font-source (or other locations) when using. Happy to do it this way if it will not slow down production app. Does it download font during build or at runtime?

chloebrett: Honestly not sure - I figured it might be at runtime but it's small so probably won't matter, and there's probably a caching layer somewhere. All very hand wavy, downloading the actual font sounds like a better idea

ghost: Ok. As long as it works offline at runtime (as ours is specifically an offline app), then sounds good to me.

emmapainter (approved):

ghost (approved):

chloebrett:
Not going to merge this yet, I'll make the font a download into the codebase first as I'm not sure it will work offline otherwise

chloebrett:
@hhen0005 using a downloaded font now, give it a try

ghost (approved):
Nice! Thanks for changing it

PR #132, opened 2022-07-29
Add cut, copy, paste, delete to menu and work into ipc rather than be…

### LINKED ISSUE #111: Cut/copy/paste in menu
“Cut”, “copy” and “paste" (with respect to the transcription) should be in the edit menu, and should also be refactored to use ctrl / command + x/c/v rather than just x/c/v.

They are currently in the edit menu on Mac, but they aren't linked to anything. They're not currently in the edit menu on linux/windows.

chloebrett:
- Puts the cut, copy, paste, delete buttons in the menu bar / makes them work
- Keyboard shortcuts now actually require the ctrl / cmd modifier key to be held for them to work
- Puts the clipboard into redux
- Doesn't fix any of the actual bugs with cut/copy/paste delete, that will come later

Resolves #111


chloebrett (commented):
Q from Riley: what happens if we select outside the project block?

Dan: would be good to do keyboard shortcuts without having to put them in the menu

Riley: would be good if we could make it clear that this is specific to the project page / transcription block - behaviour defined in clipboard.ts is very specific to project.tsx

rileykeane (approved):

PR #135, opened 2022-07-29
Fix bug where blank media shows if import is cancelled

### LINKED ISSUE #102: Browsing for a file to import and then cancelling causes a blank selected file to appear
Screen Shot 2022-05-28 at 3 23 14 pm

The transcription button then gets enabled, but clicking it doesn't do anything

chloebrett:
Resolves #102


Irmy (approved):
Looks good! Works as expected

Issue #100, opened 2022-05-28
Long project name goes off screen on import media page


chloebrett:
Import media page: <img width="702" alt="Screen Shot 2022-05-28 at 3 19 00 pm" src="https://user-images.githubusercontent.com/6735055/170811320-4b676c61-1c2e-48b4-a7cc-86874f458b1d.png">

Transcription page: <img width="629" alt="Screen Shot 2022-05-28 at 3 19 17 pm" src="https://user-images.githubusercontent.com/6735055/170811328-f0ec08e9-e490-47d7-9965-900a6683a4ab.png">

Transcription page looks fine, import media should also truncate

PR #137, opened 2022-07-29
add overflow property to Typography component

MutazAshhab:
Bug #100

Screenshot:
image


MutazAshhab:
Here is the before: 
image

chloebrett (approved):

PR #143, opened 2022-07-29
Fix cut/copy/paste/delete

### LINKED ISSUE #113: Deleting entire transcription leaves start (before first word) and end (after last word)
Think of a way to resolve this issue

One idea is to make it that deleting the first word removes everything before it, and deleting the last word removes everything after it. Another idea is to represent the silences and allow those to be deleted.

### LINKED ISSUE #116: Copy+paste words does not work with delete or undo
Example:

1) base transcription

Screen Shot 2022-05-28 at 3 43 43 pm

2) copied the word "financial" 3 times

Screen Shot 2022-05-28 at 3 44 03 pm

3) deleted entire transcription

Screen Shot 2022-05-28 at 3 44 15 pm

Note: trying to select the copied words again and delete them just doesn't work at all - these words are now invincible 😎 

4) hit undo

Screen Shot 2022-05-28 at 3 46 35 pm

5) hit redo

Screen Shot 2022-05-28 at 3 46 42 pm

6) hit undo (this pattern repeats and never stabilises)

Screen Shot 2022-05-28 at 3 46 50 pm

### LINKED ISSUE #138: Cut and paste does not handle spaces properly
Subissue of #93

If you cut and paste a sequence of words, some of the spaces between words disappear.

Before:

Screen Shot 2022-05-22 at 11 43 49 am

After:

Screen Shot 2022-05-22 at 11 44 24 am

### LINKED ISSUE #139: Deleted words handled improperly by cut and paste
Subissue of #93

If you delete words, then cut the words surrounding them and paste them elsewhere, the deleted words return.

Before:

Screen Shot 2022-05-22 at 11 45 06 am
 
After delete:

Screen Shot 2022-05-22 at 11 45 40 am

After paste:

Screen Shot 2022-05-22 at 11 45 58 am

### LINKED ISSUE #140: Delete -> cut -> undo sequence does not behave as expected
Subissue of #93

If you delete words then cut them, then undo the cut but don't undo the delete, the deleted words return. If you redo the delete and then the cut, the state adjusts as expected. This indicates there might be a problem with your undo but not your do part of the cut op.

This one is a little harder to show in screenshots but if you follow the instructions you should be able to replicate it.

### LINKED ISSUE #141: Multiple copy breaks undo
Subissue of #93

Copying words all over the place breaks undo. This might be linked to the known issue of indexes not working - but in theory this should be deterministic even with that bug. Below is the result of copying a bunch of times, undoing and then redoing, then undoing until there is no more left to undo. As you can see this doesn't return us to the original transcript, which should always be the result of undoing to the top of the stack. In fact in the state shown below, if I redo a bunch of times and then undo some more, _even more_ copies of "financial business" appear.

Screen Shot 2022-05-22 at 11 49 14 am

chloebrett:
Massive PR, will break this up and it's a WIP for now, but it basically fixes the cut/copy/paste/delete editor functionality. It also fixes the issue of spaces disappearing when words are chopped and rearranged.

Video demo:

https://user-images.githubusercontent.com/6735055/181770373-b321364b-69ca-47d2-8953-15a0d620792b.mov

Video looks a bit random without keyboard shortcuts being shown, but basically I start by copying a phrase, then paste it in a few places, then do the same with cutting, then delete some words, then undo everything, then redo everything. All of this works as expected.

However... in doing so it breaks the video preview 😆 so I will have to look into that before finalising this. Hopefully that will be the easy part since I think it's just the transcript to cuts function we need to look into

Some other notes of interest
- Sets up a dummy transcription service to just return the sample transcription, pretty useful for those with M1's or who just don't want to wait a long time for transcriptions to finish
- Moves to UUIDs for word keys so that we don't get duplicate key errors (this causes some potential other problems but there should be workarounds for those)
- Re-creates spaces every time an edit is made so that there are always appropriate spaces between words
- Reworks the clipboard to store the actual words content rather than just indices, since the clipboard exists outside of time with respect to the undo stack, so rather than updating the clipboard every time we undo or redo, we can just store the actual content to be copied
- moves the live process function to the front end since it's only called from there, and calling it from the back end causes some weird polyfill stuff (might help with packaging, that could be part of the build error!)
- instead of testing liveProcess directly, we test it via the transcription reducer and I made a bunch of helper methods to simplify this process
- fixed a bunch of random off by one errors that were causing some of the UI weirdness with cut/copy/paste

Also, this PR builds on #132, so it will become marginally smaller once that's merged.

Resolves #141
Resolves #140
Resolves #139 
Resolves #138
Resolves #116
Resolves #113


src/main/editDelete/preProcess.ts
```diff
@@ -19,89 +20,16 @@ const camelCase: MapCallback<SnakeCaseWord, PartialWord> = (word) => ({
...
-
-/**
- * Adds spaces between words which represent the silence between each word
- * @param result The output being constructed
- * @param word The current element of words
- * @param index The index of word in words
- * @param words The list of words being reduced
- * @returns The updated transcript with a silence after word
- */
-const addSpaces: (totalDuration: number) => MapCallback<Word, Word[]> =
```
LiamTodd, line 63: does the transcription object no longer use spaces to represent silences?

chloebrett: No, this logic was moved to a shared file as it’s used both by live process (front end) and pre process (back end)

see processingShared.ts

However I think we may be able to simplify things and only add in spaces for (a) rendering the text and (b) video preview, which would make manipulating the transcript far easier

chloebrett: Update: no longer using spaces

chloebrett:
Done!!! 🎉

Here's a demo, showing the following:
- Moving around the timeline both when stopped and while video is playing
- Cutting, copying, pasting text
- Some edge cases
- Integration with undo

Demo link: https://drive.google.com/file/d/1X0Sm29Td3CxAPTm1reNrvnztECHh0oIg/view?usp=sharing

As you can see, it all works more or less without a hitch (as far as I can tell from my own testing) - even edge cases like copying a word, pasting it next to itself, then copying both the original word and the pasted word, and pasting both of them somewhere else. (without some deliberate handling, this would create react key errors)

Some design docs which we can discuss when going over this PR together:
- https://docs.google.com/document/d/1lQPJ4-kCI72GhNjpNbguT2iT5ihD_6oBikop-gXziYQ/edit
- https://docs.google.com/document/d/1hJAQW0rriYm4-f_yhoQ0ZUz3SGDxKbtScCl_9j4Pt3U/edit#
- https://docs.google.com/document/d/14xBZMCWNxhK6rJkEZ3WLhEVXluvb92SD_-dHGY9TkRI/edit

Some notes on implementation:
- Word keys aren't UUIDs any more [i.e. I changed them to be at the start of this PR, but have gone back on that decision], since we need reducers to be deterministic. Instead we use a combination of 'originalIndex' and 'pasteCount', where each pasted copy of a word has a unique pasteCount.
- We no longer represent spaces as words themselves; instead we add buffers between words which for words in the middle of the transcription consist of half the silence duration before/after that word. For the start and end words, it includes the entire start and end silence. This makes manipulating the transcription for cut/copy/paste etc way way way easier. It also makes the playback smoother. See the docs above for a full explanation
- 'liveProcess' is replaced with a simplified function, 'updateOutputStartTimes', which does what it says on the tin.
- Lots and lots and lots and lots of transcription reducer tests have been written :)
- Now using inclusive-start, exclusive-end for delete ranges as they cause less off-by-one errors with slice() etc
- Highlighting the currently-playing word now accounts for buffers, so spaces are never highlighted and there is always a word that is highlighted 🙂 
- convertTranscriptToCuts has been rewritten using Array.reduce as it's a bit cleaner

Now obviously this is an ungodly huge PR, unfortunately as it provides more or less integrated functionality that I can't really see being broken down into smaller parts, I think trying to break it into multiple PRs won't be a great idea. Instead, I propose:

1. First we get #132 reviewed, approved and merged, since this builds on that.
2. Then let's all sit down together (whoever is interested in this PR, I'm thinking @rileykeane, @aretle, @LiamTodd maybe?) and we'll go through it, you can ask questions and give feedback and suggestions while I explain the changes I've made
3. As a side note, a significant portion of the line changes are from the test files being changed to fit new object formats, so while it's still a massive PR, don't be _too_ intimidated 🙂

chloebrett (commented):
Don't need space at start or end of rendering

Tag Hugh

Dan: why can't we just use "current length of transcript plus one" as the key each time we paste / create new keys? -> I think there was an issue with this but I can't remember specifically what

Liam: what if we use "original index concatted with new index" as a key

ghost:
Just doing some testing of the app and found an issue. 
If I cut and paste the first word into the middle then the video preview doesn't seem to get updated. 
The video preview will still the play the cut content (the clock and the transcript text preview do get updated though) and will not play the pasted content. 
Of course, once the paste point has been passed in the video playback, the bug is not noticeable.
Let me know what you find please :)

chloebrett:
> Just doing some testing of the app and found an issue. If I cut and paste the first word into the middle then the video preview doesn't seem to get updated. The video preview will still the play the cut content (the clock and the transcript text preview do get updated though) and will not play the pasted content. Of course, once the paste point has been passed in the video playback, the bug is not noticeable. Let me know what you find please :)

Good find, will investigate. I think it's probably a bug in transcriptToCuts since the editor experience is unaffected

chloebrett:
I've fixed all known issues:
- pasteCount is now pasteKey and the algorithm is O(n) -> doesn't need to take into account originalIndex, that was the simplification (thanks Dan)
- the bug that Hugh identified is fixed - it originated from the function that calculates output start times erroneously assuming the first word in the array would also be the first word in the original transcript
- Spaces are no longer rendered before and after the first/last word in the TranscriptBlock component

I think that addresses all issues raised, please re-review and approve if you are happy with the PR :)

src/renderer/store/transcription/reducer.ts
```diff
@@ -26,65 +28,78 @@ const transcriptionReducer: Reducer<Transcription | null, Action<any>> = (
...
-    }
+  if (action.type === PASTE_WORD) {
+    const { startIndex, clipboard } = action.payload as PasteWordsPayload;
+
+    const prefix = transcription.words.slice(0, startIndex + 1);
+
+    // Paste key must be unique for all pasted words - that is, no two pasted words should ever have the same paste key.
+    // We force this invariant by finding the highest paste key in the entire transcription to this point, and then
+    // adding n to it for the nth pasted word, for all words on the clipboard.
+    const highestExistingPasteKey = Math.max(
```
rileykeane, line 75: 😍

rileykeane (approved):

danfeltham (approved):
Happy for this to be merged.

LiamTodd:
Looks good, thanks for explaining everything @chloebrett

LiamTodd (approved):

ghost (approved):

PR #148, opened 2022-07-30
Bugfix/dont show red circle when transcribing

### LINKED ISSUE #92: isDocumentEdited marked true during transcription
Noting this issue for myself

When the transcription is running the traffic light goes circular, indicating the open file is edited

This isn't what we want as if the user abandons at this point we would rather just lose their in progress work - the file is not yet saveable

Screen Shot 2022-05-24 at 6 49 16 pm

### LINKED ISSUE #110: Mac: red traffic light circle appears too early
This appears as soon as the "currentProject" is defined which is perhaps too early.

In general, think about what the saving options should be when the project is only half-formed (e.g. transcription is loading). I don't think it makes sense to allow the user to save the file at this point, so we should prevent that.

chloebrett:
Well that was easier than expected 😆 

Tagging Riley as he has a mac and this is easiest to see on mac

For some reason it won't let me untag taz so we have a linux tester too 😎 

Resolves #92 
Resolves #110 
(oops I think I made duplicate issues)

~~Builds on #147~~ (that PR is now merged)


kateahyde:
Just checking I'm understanding properly, is this basically saying if there's no transcription associated with the current project then there isn't anything to save so saving shouldn't be enabled?

chloebrett:
> Just checking I'm understanding properly, is this basically saying if there's no transcription associated with the current project then there isn't anything to save so saving shouldn't be enabled?

That is correct @kateahyde

chloebrett:
See the linked issues this resolves for a description

kateahyde (approved):

kateahyde:
Thanks for the clarification and for pointing to the issues, I'll be sure to check them properly next time :) @chloebrett

PR #150, opened 2022-07-30
Various file fixes

### LINKED ISSUE #77: Deleting a project in the opening menu removes new project option
When you press the trash can on a project in the main menu it removes everything from the page not just the project you delete

### LINKED ISSUE #91: Miscellaneous file stuff
The interaction of save / save as with recent projects has some bugs

From #72 :

> Okay, I've been playing around with this and found a bug.
> 
> It seems on Windows, the project's listing in the "recent projects" gets duplicated when you save via the "save" dialog that appears when you try to quit without saving. I'm not sure why this happens and haven't been able to replicate it on Mac, though I haven't tried on Linux.
> 
> On mac, a different bug occurs: the "recent projects" list actually disappears entirely. This is due (to the best of my understanding) to the fact that the recent projects file is being written when the app quits, and so the operation doesn't finish and the file is left in a blank state.
> 
> Obviously not ideal, but avoidable for the sake of demo by making sure you save before hitting the red traffic light or close button. Compared to the previous state of affairs: if you clicked close without saving, the project would simply not save at all.
> 
> I've thought about how to resolve these issues and I think the solution will be too involved to get done before the demo, and it's not a priority feature for us to demonstrate. However, showing the open file and whether it's been edited is quite a nice UX, so I'd like to suggest this gets merged and we deal with the bug I just mentioned in a separate PR.
> 

I think it's hard to keep track of what was introduced where, so we should just QA develop and try to fix these bugs as much as possible.

### LINKED ISSUE #108: File location should not be saved to the file
We currently record the `saveFilePath` in each mlvet file, which is redundant since this file is by definition saved somewhere. If the user moves or renames the file, then this information instantly becomes out of date. Instead, we should work out the `saveFilePath` when we load a file based off the file path we used to load it... so the `saveFilePath` can be stored in the recent projects as metadata, but not in each mlvet file itself.

### LINKED ISSUE #109: Export location should not be saved to the file
As every time the user exports, they get to choose the location anyway. We don't need a reference to where the file was exported for any other reason. Perhaps a "default export directory" setting but this sort of requires we have some kind of preferences view (and persist that to disk) so would need additional tasks. For now could just use the save directory as default path for the popup export window.

### LINKED ISSUE #123: Refactor: make clear type definitions for various project embodiments
Relates to #108 and #109

Basically, a project can be one of four things:

- A partially-made project when we're in the middle of creating it (e.g. no transcription yet)
- A fully-made project when we're editing it
- A project saved to disk (that shouldn't contain stuff like the save file location explicitly, since that's known by definition since the file is saved somewhere)
- A set of metadata about a recent project that allows us to locate it and know things like the media size, date modified and thumbnail location

### LINKED ISSUE #127: Error if recent projects doesn't exist
```
[Error: ENOENT: no such file or directory, open '/Users/chloe/Library/Application Support/Electron/mlvet/recentProjects.json'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/Users/chloe/Library/Application Support/Electron/mlvet/recentProjects.json'
}
```

chloebrett:
~~Builds on #147~~ (that PR is now merged)

Resolves #77
Resolves #108
Resolves #91
Resolves #127
Resolves #123
Resolves #109

Changes:
- Introduce notion of a 'runtime project' which knows what file name it originated from, if it's been edited recently, etc. Persisted projects do not need to store this data as it is implied (e.g. file name is implied by what file we save to!)
- Don't store export path, thumbnail path or audio extract path. For the first, it's just determined based off where the user wants to save, and defaults to the project directory. For the last two, they are determined based off the userdata folder location (depends on OS) and the project ID.
- Shows a delete-confirm dialog if the user tries to open a project from recents where the project file doesn't exist
- If the user opens a foreign mlvet file (i.e. one they didn't make themselves), it appears in their recent projects list even if they don't save it
- When a project is saved, make sure no two projects with the same ID end up being saved in the recent projects section

Potential issues
- We're still heavily reliant on project IDs so if the user copies a project file outside of mlvet and tries to open it, we treat it like it's the same project. Can get around this by using save as but it's not ideal behaviour


tngu0041 (approved):
There's too much to take in for a single PR, so I'm mostly just trusting that everything is right.

I'd need more time working on Electron stuff before I'll fully understand everything going on.

PR #152, opened 2022-07-30
Use actual thumbnails instead of example thumbnail

### LINKED ISSUE #124: Thumbnail is currently hard-coded
Please make this not be the case 😂

chloebrett:
Screenshot:

Screen Shot 2022-07-30 at 8 35 34 pm
 
Approach is a very simple one - load the thumbnail on the back end, then send it to the front end using base64.

The promise thing can potentially cause a memory leak, let me know if you have ideas to fix it but it seems like it doesn't really impact the app performance

~~Depends on #150, so don't review until that is merged~~

Resolves #124


LiamTodd (dismissed):

fwang1703 (approved):
Not sure about the memory leak stuff either but everything else looks good! Awesome work

src/renderer/components/RecentProjectsBlock.tsx
```diff
@@ -70,6 +74,25 @@ const RecentProjectsBlock = () => {
...
+    )
+      .then((list) => {
+        const map: Record<string, string> = {};
+        list.forEach(({ id, thumbnail }) => {
+          map[id] = thumbnail;
+        });
+        setThumbnails(map);
+      })
+      .catch();
+  }, [recentProjects]);
```
ghost, line 100: Ok yes there will be a memory leak. There needs to be a cleanup function for the useEffect (basically the useEffect should return a function than cancels any subscriptions in it) that aborts the ipc request. In a fetch request equivalent I would abort the fetch call using an abortController and then the catch would catch the AbortError created. 

However, I'm struggling to figure out a way to abort an IPC request. Initial research indicates that it is not possible. We can come back to this.

The more pressing issue I feel is that the memory leak is apparent when testing with 1 recent project. This shouldn't happen as the user opening a recent project (and thus the page unmounting) should happen well after all of the thumbnails have been loaded (it's not a huge process I think).

If you chuck a `console.log` in the useEffect you'll see that `recentProjects` is getting updated constantly. Again, this should not be happening. The problem is that the state variable `recentProjects` is being modified directly (sorted & sliced). I don't think this is the correct pattern.

If we want to always have access to the sliced / sorted version of recent projects then that should be stored in another state variable (eg: `recentProjectsSortedAndFiltered` which is updated (via another useEffect) whenever the original state variable (`recentProjects`) is modified.

image

In the example above, 'dog' should only get printed whenever `recentProjects` is modified or on page unmounted.

image

Alternatively, if the `recentProjectsSorted` was only needed once in that useEffect then you could simply sort and slice it there rather than having a separate variable.

ghost: In regards to figuring out a way to abort the IPC request, I really don't know. The only resource I found on aborting them said its not possible.

While the memory leak could be a problem, it will not be common as the user would need to have quite a few projects. While, I would like to address it, I think the time required (I already spent like 20 mins) would outweigh the reliability and performance benefits.

If you disagree, I'm happy to discuss in architecture meeting.

ghost (changes requested):
Nice work! Small PR and easy to read. 
Happy to approve once memory leak comment I made is addressed :)

src/renderer/components/RecentProjectsBlock.tsx
chloebrett: Seems reasonable. As for the re-rendering, we can use `useMemo` instead of `useEffect` for the sorted + sliced list as it's a bit more idiomatic. Trying now

chloebrett: Regarding aborting the IPC I have two ideas:
1. find a generic "cancelable promise" library and use that - may not be possible
2. add an `AsyncState` to the thumbnail loading request and prevent page changes if that request is in flight (thereby stopping the component from unmounting)

I still think it's probably not worth the effort to do either of these, clearly any performance issues from the bug were caused by the 'constant rerender' issue which is now fixed. Good catch on that btw, I think I realised when I wrote it it was lazy but figured it wouldn't be an issue since at that time there was no cause for the home screen state to change

chloebrett:
@hhen0005 give it another look :)

src/renderer/components/RecentProjectsBlock.tsx
ghost: useMemo! thanks for taking it one step further :)

I like the async state idea but agree not worth the time. We can always implement it later if we come across a case where aborting an IPC request will have a significant performance impact (perhaps transcription???)

ghost (approved):
Very nice. I'm happy with it. I added a tiny linting issue fix. Could you please double check it?

PR #156, opened 2022-08-03
Bespoke selection logic

### LINKED ISSUE #163: Transcribing a video with no speaking (no transcription found) breaks program.
image

chloebrett:
Improves the selection logic to support
- clicking words to select them
- cmd/ctrl + clicking words to add/remove from the current selection
- shift + clicking words to expand the selection

Also improves some other things
- words are now their own component
- highlighting is a bit smarter
- clipboard.ts now exclusively deals with clipboard related things
- lots of tests

Things that need fixing

- [x] write tests
- [x] fix existing tests that are failing
- [x] separate between cmd and ctrl depending on OS
- [x] move modifier checks outside of the word object itself

Caveats
- I haven't tested it on windows or linux yet
- dragging to make a selection is not supported yet, so this counts as a temporarily 'lost feature' given it was primarily how selection was done before. It will come back in a future PR, it's not really in scope for this one but the existing selection logic was incompatible so it had to be removed
- shift + click selection logic works fine for expanding the selection outwards, but behaviour for clicking inside the existing selection has more ambiguous requirements; for now I just implemented a 'paint fill' type algorithm. More complex options are available but require keeping track of selection order (possible but maybe not a priority)

Disclaimer

- again, this is a massive PR
- however, of the ~900 line additions, ~400 of them are from tests

Video

https://user-images.githubusercontent.com/6735055/182984423-2ea4ef71-7fd6-48ba-8cb4-4fb3d70d9250.mov

Resolves #163


src/renderer/components/MediaDisplayOnImport.tsx
```diff
@@ -33,9 +33,10 @@ const MediaDisplayOnImport = ({ fileName, removeMediaFromImport }: Props) => {
           </Typography>
         </Stack>
         <IconButton
+          onClick={removeMediaFromImport}
           sx={{ color: colors.yellow[500], fontSize: 24, margin: '5px' }}
         >
-          <DeleteIcon onClick={removeMediaFromImport} />
+          <DeleteIcon />
         </IconButton>
```
MutazAshhab, line 40: Why did you move the `onClick` back to the IconButton after approving #161.

Did it make the clickable section smaller because of the margin?

MutazAshhab:
Now, i was about to test this but then I realised that copying, cutting, pasting are not working for me (both this branch and develop).

These shortcuts are working for me:

1. Undo - ctrl+z
2. Delete - backspace
3. Delete2 -Delete

Notes:
- Replacing `CommandOrControl` with `Ctrl` did not work

Could you run the app on your linux machine?

chloebrett:
I'll give it a try

src/renderer/components/MediaDisplayOnImport.tsx
MutazAshhab: Oh my bad, i got it mixed up… ignore this

chloebrett:
@Taz17 tested on my linux machine, this is really weird. Clicking the cut/copy/paste items in the _menu_ works, but using the keyboard shortcuts doesn't. Which is really weird because (a) the shortcuts are marked as ctrl+X, etc in the menu, and (b) it's the exact same approach as used by undo/redo.

I tested adding a log every time the 'cut' selector fires - it works when clicked, doesn't work when the keyboard shortcut is used. Maybe X, C and V are reserved? That would at least explain why delete/backspace are working...

Can you please test the functionality by clicking the menu items instead? Will look further into the keyboard shortcuts thing...

MutazAshhab:
@chloebrett I just tested clicking the menu items and it worked (using the shortcuts still does not work).

Could you publish a branch with any changes you made? I'll poke around and see what you tried when I have extra time

chloebrett:
All I tried was adding a console.log to the cut's click event. I guess another thing to try would be testing a different keyboard shortcut, maybe Ctrl+U or something for cut might work. If so then it's definitely the OS's problem

src/renderer/store/selection/reducer.ts
```diff
@@ -0,0 +1,71 @@
...
+  if (action.type === SELECTION_RANGE_ADDED) {
+    const { startIndex, endIndex } = action.payload as IndexRange;
+
+    /**
+     * Selection is stored in redux as an array because redux
+     * is only supposed to store primitives. So cast it to a set to manipulate then
+     * convert it back. O(n) with respect to length of transcription which is the same as
+     * the degenerate "select everything" case anyway
+     */
+    const selectionSet = new Set(selection);
```
ghost, line 25: Love your work.

Is it possible to put this functionality in a function? That function could take a modification function as a parameter (eg: `add()` or `delete()`) or maybe an enumerator value (eg: `'ADD'` for `add()`). Could be put in a helpers file or something.

Just want to make it a bit easier to read by making it more compact. :)

src/renderer/util.ts
```diff
@@ -162,17 +163,32 @@ export const removeExtension: (fileName: string) => string = (fileName) => {
...
+
+  return listNew;
+};
+
+/**
+ * By default the Array.sort() function sorts alphabetically, e.g. 10 comes before 2.
+ * This function properly sorts number arrays.
+ * It mutates the input array.
+ */
+export const sortNumerical: (list: number[]) => void = (list) => {
```
ghost, line 192: Nice :)

src/renderer/store/selection/reducer.ts
chloebrett: Put what in a function specifically? The entire reducer for SELECTION_RANGE_ADDED? It's pretty short, I'm not sure what the benefit would be vs keeping the selection manipulation reducers in one place

chloebrett: Maybe - what about if it was

```
enum SelectionModifier {
  ADD,
  REMOVE,
  TOGGLE
}

const updateSelection: (selection: number[], modifier: SelectionModifier) => number[];
```

And then just within the function,

```
const mapper: Record<SelectionModifier, (wasSelected: boolean) => boolean> = {
  SelectionModifier.ADD: () => true,
  SelectionModifier.REMOVE: () => false,
  SelectionModifier.TOGGLE: (wasSelected) => !wasSelected,
}
```

Oops I just 3/4 implemented it in this comment 🙈

ghost: To clarify, I meant the logic from put selection in set, for loop and turn back to array in a separate function (with the for loop modifier as a parameter). Just an idea.

All good if you don't think its worth it.

ghost (approved):

src/renderer/store/selection/reducer.ts
ghost: Sorry, confused by comments. If it reduces the re-used lines then yes please! :)

chloebrett: Yep that’s what the proposed impl does :)

rileykeane:
Has the ability to highlight words with a cursor to make selections (like in a word doc) been removed? It's not working for me. 

**If this is intended:** I think it should be brought back. The original spec for the UX was to be as close to Descript as possible - which has a similar workflow to editing a word doc. I'm not saying these new selection options should be removed (they are quite useful and something Descript does't have) however the original UX should still exist. 

**If this is not intended:** it's either not working on my local machine (M1 MacBook) or just me not knowing what I'm doing. 

@Irmy  for PM visibility.

rileykeane (changes requested):
See my comment above.

rileykeane:
> Has the ability to highlight words with a cursor to make selections (like in a word doc) been removed? It's not working for me.
> 
> **If this is intended:** I think it should be brought back. The original spec for the UX was to be as close to Descript as possible - which has a similar workflow to editing a word doc. I'm not saying these new selection options should be removed (they are quite useful and something Descript does't have) however the original UX should still exist.
> 
> **If this is not intended:** it's either not working on my local machine (M1 MacBook) or just me not knowing what I'm doing.
> 
> @Irmy for PM visibility.

Ignore most of this - just reread the caveats.

rileykeane:
> Ignore most of this - just reread the caveats.

However, I do think drag to select should be a higher priority and will need to get confirmation from Zac on how crucial that feature is to him.

chloebrett:
> > Ignore most of this - just reread the caveats.
> 
> However, I do think drag to select should be a much higher priority and will need to get confirmation from Zac on how crucial that feature is to him.

It’s right next up on things I’m going to work on, this PR is already 900 line additions so it’s separate - will be done by week 3

chloebrett:
Shift+click does the same thing as well, so it’s not like any tasks that could be done before now can’t, the UX is just different

Irmy:
I will check with Zac regarding which selecting feature is most priority for him since intuitively drag with cursor is definitely what I went to use first. 

Shortcuts on my end also didnt work ( Microsoft OS) - only one that did work was Ctl Z , delete and backspace.  Using the menu options did though

chloebrett:
> I will check with Zac regarding which selecting feature is most priority for him since intuitively drag with cursor is definitely what I went to use first.
> 
> Shortcuts on my end also didnt work ( Microsoft OS) - only one that did work was Ctl Z , delete and backspace. Using the menu options did though

Don’t worry about checking with Zac, as I said I will literally have drag select done by the end of the week, well before the end of milestone 3. It’s not like it’s been de scoped. It just had to not be part of this PR because it’s massive enough as is.

chloebrett:
Sounds like same problem as Linux. Will investigate further

rileykeane (approved):

MutazAshhab (approved):
Sorry this took a long time to review!

PR #157, opened 2022-08-04
Returning to home checks for file dirtiness before prompting

### LINKED ISSUE #95: Return to home needs to check for file dirtiness
Return to home PR: https://github.com/patrickbrett/mlvet/pull/81

File representation PR which implements dirtiness: https://github.com/patrickbrett/mlvet/pull/72

Also needs to sync up with the warning dialog as we declare it in two places

This was all introduced because these two were being worked on in parallel => hence duplicated work

fwang1703:
## Summary

This PR fixes [issue #95](https://github.com/chloebrett/mlvet/issues/95), where originally returning to home would only prompt you to save when saving has never occurred. Now returning to home will prompt users to save if their file has been updated since the last save. 

<hr/>

This change is necessary because we want to ensure a user is given warning about unsaved changes before leaving the main project page. 

<hr/>

### What did you change?

- Updated the `returnToHomeHandler` to check if file has been edited
- Extracted out the save changes dialog to its own separate component (removing duplicated code in both `saveUtils` and `promptToSaveWork`

<hr/>

### OS

Need reviewers from all OS' to test (hopefully this is covered in my reviewers choice)

- [x] Linux
- [x] MacOS
- [ ] Windows


src/main/handlers/misc/saveChangesDialog.ts
```diff
@@ -0,0 +1,27 @@
+import { BrowserWindow, dialog } from 'electron';
+
+type SaveChangesDialog = (
+  mainWindow: BrowserWindow,
+  projectFileName?: string
+) => number;
+
+export const SAVE_SELECTED = 0;
```
chloebrett: If you make these options of an enum, then you can set the return type of SaveChangesDialog to that enum

src/main/handlers/misc/saveChangesDialog.ts
```diff
@@ -0,0 +1,27 @@
+import { BrowserWindow, dialog } from 'electron';
```
chloebrett: This handler should probably be under `handlers/file` instead of `handlers/misc`

src/main/handlers/helpers/saveUtils.ts
```diff
@@ -54,19 +55,15 @@ export const confirmSave: (
...
-  });
+  const confirmSaveDialogResponse = saveChangesDialog(
+    mainWindow,
+    proposedFileName
+  );
 
   // if cancel button is selected
-  if (confirmSaveDialogResponse.response === cancelled) {
+  if (confirmSaveDialogResponse === SAVE_CANCELLED) {
     throw new Error('Dialog cancelled');
```
chloebrett: I know you didn't write this but I think throwing an error on dialog cancel isn't the right behaviour - can we just change this to console.log that it was cancelled

chloebrett (commented):
Tested the following
- attempt to go home from project that doesn't have any edits
- attempt to go home from project with edits
- use all three buttons in the dialog

All work as expected, just a few structure comments. Nice work refactoring!

fwang1703:
@chloebrett made updates from your review, could you please have another look thanks!

MutazAshhab:
I tested the same cases that Chloe did and they all worked on Linux as well.

> Tested the following
> 
> * attempt to go home from project that doesn't have any edits
> * attempt to go home from project with edits
> * use all three buttons in the dialog

src/main/promptToSaveWork.ts
```diff
@@ -17,17 +21,9 @@ const promptToSaveWork: (
...
-    buttons: ['Save', `Don't Save`, 'Cancel'],
-    type: 'warning',
-  });
-
-  if (buttonIndex === 0) {
+  const buttonIndex = saveChangesDialog(mainWindow, fileName);
+
+  if (buttonIndex === SaveDialogSelections.SAVE_SELECTED) {
     // Save, and tell the app to close after it's saved
     mainWindow.webContents.send('initiate-save-project', closeAfter);
```
MutazAshhab, line 28: Now that you implemented the SaveDialogSelections enum, Could you update line 34 to use the the enum as well rather than checking if it equal to 1?

```javascript
  if (buttonIndex === 1) { // <-- this line, im not sure if github will show this on the review
    // Don't save
    return true;
  }
```

MutazAshhab (changes requested):

src/main/handlers/file/saveChangesDialog.ts
```diff
@@ -0,0 +1,29 @@
+import { BrowserWindow, dialog } from 'electron';
+
+type SaveChangesDialog = (
+  mainWindow: BrowserWindow,
+  projectFileName?: string
+) => SaveDialogSelections;
+
+export enum SaveDialogSelections {
```
chloebrett, line 8: Nice 👍🏻

chloebrett (approved):

src/main/promptToSaveWork.ts
fwang1703: Good pick up completely missed that one! Just fixed it

MutazAshhab (approved):

PR #158, opened 2022-08-04
Bugfix/transcribe button delay

### LINKED ISSUE #106: "Transcribe" button has delay
When the "transcribe" button is pressed on the import page, we do some background processing (audio and/or thumbnail extract, I can't remember). This causes a 500-2000 ms delay which is not a great user experience. The button also doesn't disable or show that anything is loading.

Instead, we should either:

1. Do this processing on the transcription page instead (best)

or

2. Make the button disabled / say "loading" (less ideal but still better than current situation)

exquisitebin:
## Transcription Button Delay is Removed

https://user-images.githubusercontent.com/58348527/182779831-39525fce-a221-4d96-8185-256062ae0f9c.mp4

<hr/>

Pressing the Transcribe button caused a few seconds long delay in the ImportMediaView. MLVET used to perform a series of metadata functions for the project file and extraction of audio before moving to the next modal view.

<hr/>

### What did you change?
The media file functions are moved to the RunTranscriptionView.ts from the ImportMediaView.ts so that the processing can take place during the loading screen.

To do this mediaFilePath is saved to the currentProject on ImportMediaView but all other metadata (including thumbnail extraction) takes place on RunTranscriptionView before the transcription starts.

### OS

- [ ] Linux
- [x] MacOS
- [x] Windows


src/renderer/components/NewProjectModal/RunTranscriptionView.tsx
```diff
@@ -41,7 +52,7 @@ interface Props {
 const RunTranscriptionView = ({ closeModal, nextView }: Props) => {
   const [asyncState, setAsyncState] = useState<AsyncState>(AsyncState.READY);
   const [mediaFileName, setMediaFileName] = useState<string | null>(null);
-
+  const [mediaFilePath, setMediaFilePath] = useState<string | null>(null);
```
chloebrett: Not needed, see below

src/renderer/components/NewProjectModal/RunTranscriptionView.tsx
```diff
@@ -61,21 +72,55 @@ const RunTranscriptionView = ({ closeModal, nextView }: Props) => {
     ) {
       return;
     }
+    setMediaFilePath(currentProject.mediaFilePath);
```
chloebrett: Since you only ever set this to be `currentProject.mediaFilePath`, you can just replace all references to `mediaFilePath` with `currentProject.mediaFilePath`, or destructure it at the point where you set it if you prefer i.e. replace

```
setMediaFilePath(currentProject.mediaFilePath)
```

with just

```
const { mediaFilePath } = currentProject;
```

src/renderer/components/NewProjectModal/ImportMediaView.tsx
```diff
@@ -52,51 +46,35 @@ const ImportMediaView = ({ prevView, closeModal, nextView }: Props) => {
...
-  const setCurrentProject = (project: RuntimeProject) =>
-    dispatch(projectCreated(project));
-
   const handleTranscribe = async () => {
-    if (mediaFilePath === null) {
-      return;
-    }
-
-    const projectWithMedia = await updateProjectWithMedia(
+    const projectWithMedia = await updateProjectMediaFilePath(
```
chloebrett: This whole `updateProjectMediaFilePath` util probably isn't necessary, you can just do

```
const projectWithMedia = {
   ...currentProject,
   mediaFilePath,
};
```

And then check if mediaFilePath was null for the early exit

src/renderer/components/NewProjectModal/RunTranscriptionView.tsx
```diff
@@ -61,21 +72,55 @@ const RunTranscriptionView = ({ closeModal, nextView }: Props) => {
...
 
     setAsyncState(AsyncState.LOADING);
 
-    requestTranscription(currentProject)
-      .then((transcription) => {
-        if (transcription === null) {
+    const startProcessing = async () => {
+      await updateProjectWithMedia(currentProject, mediaFilePath);
+      await extractProjectAudio();
+      requestTranscription(currentProject)
```
chloebrett: If we're awaiting everything else, may as well refactor the `requestTranscription` call to be async as well (with a try/catch for the error state)

chloebrett (changes requested):
Tested on my machine and it fixes the bug, nice job - just a few style comments

emmapainter (approved):

exquisitebin:
Made the changes requested. Not sure if I refactored the async requestTranscription correctly.

src/renderer/components/NewProjectModal/RunTranscriptionView.tsx
```diff
@@ -41,7 +55,6 @@ interface Props {
 const RunTranscriptionView = ({ closeModal, nextView }: Props) => {
   const [asyncState, setAsyncState] = useState<AsyncState>(AsyncState.READY);
   const [mediaFileName, setMediaFileName] = useState<string | null>(null);
-
```
chloebrett: Keep this newline

src/renderer/components/NewProjectModal/RunTranscriptionView.tsx
```diff
@@ -61,21 +74,54 @@ const RunTranscriptionView = ({ closeModal, nextView }: Props) => {
...
     setAsyncState(AsyncState.LOADING);
 
-    requestTranscription(currentProject)
-      .then((transcription) => {
-        if (transcription === null) {
+    const startProcessing = async () => {
+      await updateProjectWithMedia(currentProject, mediaFilePath);
+      await extractProjectAudio();
+      await processTranscription(currentProject)
+        .then((transcription) => {
```
chloebrett: The idea with await is that the contents of the `then` can be extracted, i.e.

```
const transcription = await processTranscription(currentProject);
... then do things with transcription
```

This should be in a try/catch block as well so that we can set the async state to ERROR if it fails

src/renderer/components/NewProjectModal/RunTranscriptionView.tsx
```diff
@@ -61,21 +75,54 @@ const RunTranscriptionView = ({ closeModal, nextView }: Props) => {
...
         setAsyncState(AsyncState.DONE);
         setTranscription(transcription);
-        return null;
-      })
-      .catch(() => setAsyncState(AsyncState.ERROR));
-  }, [currentProject, setAsyncState, nextView, setTranscription, asyncState]);
+      } catch {
+        setAsyncState(AsyncState.ERROR);
+      }
+      return null;
```
chloebrett: I don't think this is needed

chloebrett (approved):
Looking good, just drop the `return null` on line 114 of RunTranscriptionView if it doesn't make it throw a lint error

PR #159, opened 2022-08-05
Bugfix/drag and drop accepts any file type

### LINKED ISSUE #149: Drag and drop accepts any file type
Should only accept video files (+ audio files if we're going to do that, still not sure if that's on the roadmap?)

Irmy:
## Summary

<hr/>

 Constraints were present when importing via dialog, but not through the drag feature. 
<hr/>

### What did you change?
Branch adds in drag and drop feature only accedting the file types specified in the getMediaType function in utils. 
<hr/>

Electron 18/ chrome update doesnt allow files to be detected when file is still currently being dragged. So had to add constraint after the file has been dropped.  (https://stackoverflow.com/questions/11065803/determine-what-is-being-dragged-from-dragenter-dragover-events)
<hr/>

<hr/>

### OS

- [ ] Linux
- [ ] MacOS
- [x] Windows

### DIFF
--- a/src/renderer/util.ts
+++ b/src/renderer/util.ts
@@ -24,7 +24,7 @@ export const extractFileExtension: (filePath: string) => string | null = (
 export const getMediaType: (extension: string) => 'audio' | 'video' | null = (
   extension
 ) => {
-  const audioFileExtensions = ['mp3'];
+  const audioFileExtensions = ['mp3', 'mpeg'];
   const videoFileExtensions = ['mp4'];
 
   if (audioFileExtensions.includes(extension)) {
--- a/src/renderer/components/SelectMediaBlock.tsx
+++ b/src/renderer/components/SelectMediaBlock.tsx
@@ -90,14 +90,19 @@ const SelectMediaBlock = ({
 
     if (draggedFiles.length > 0) {
       /* Currently only supporting single file import - taking first file */
-      const fileType = draggedFiles[0].type.split('/')[1];
-      if (getMediaType(fileType) !== null) {
-        const fileName = draggedFiles[0].name;
-        const filePath = draggedFiles[0].path;
+      const fileType = draggedFiles[0].type.split('/')[0];
+      console.log(fileType);
+      const fileTypeExtension = draggedFiles[0].type.split('/')[1];
+      console.log(fileTypeExtension);
+      if (fileType !== null && (fileType === 'audio' || fileType === 'video')) {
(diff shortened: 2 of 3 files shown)


Irmy:
I will add in a toast message if they put in the wrong file type - just have to figure out how notifications on electron work first  - will do this in a different branch/ PR

exquisitebin (changes requested):
Drag and drop worked for MP4 but not MP3. Opening the same MP3 through browse still works fine. 

https://user-images.githubusercontent.com/58348527/182982210-02096609-5e9b-4cdb-9962-c20da7260dd4.mp4




Also I'd recommend uppercasing the MP4 and MP3 in the text if it is going to be changed from "Drag and drop file here".

Irmy:
Thanks @exquisitebin !! Updated Utils to fix it - drag was detecting audio as mpeg rather than mp4.

exquisitebin (approved):
The MP3 drags in now, all good to go!!
Also sorry, I just had to update the text from "Mp3" to "MP3" it was killing me XD

MutazAshhab (approved):

src/renderer/util.ts
```diff
@@ -24,7 +24,7 @@ export const extractFileExtension: (filePath: string) => string | null = (
 export const getMediaType: (extension: string) => 'audio' | 'video' | null = (
   extension
 ) => {
-  const audioFileExtensions = ['mp3'];
+  const audioFileExtensions = ['mp3', 'mpeg'];
```
chloebrett, line 27: `.mpeg` is a video file format, it is not? https://www.howtogeek.com/367560/what-is-an-mpeg-file-and-how-do-i-open-one/#:~:text=A%20file%20with%20the%20.,than%20other%20popular%20video%20formats.

To be fair I don't think I've ever seen a file with this extension, any reason we want to support it?

Irmy: because thats how the drop handler detects audio files - not as mp3 which the file is, it just categories it all as mpeg - http://www.differencebetween.net/technology/protocols-formats/difference-between-mpeg-and-mp3/#:~:text=Actually%2C%20an%20MP3%20is%20just,compression%20of%20the%20audio%20component. According to this mp3 is a subset of mpeg

Irmy: bugfix drag

Irmy: @chloebrett  see new logic - narrows it further if that helps

Issue #101, opened 2022-05-28
Clicking outside the modal when on import or transcription page, resets all progress to that point


chloebrett:
E.g. you are here:

<img width="1440" alt="Screen Shot 2022-05-28 at 3 20 42 pm" src="https://user-images.githubusercontent.com/6735055/170811359-0826449f-f16a-4238-a643-e359251aac9e.png">

You then click anywhere outside of the modal. You are returned to the home screen and have to start again (naming project, importing media, etc).

Should instead, either:

1. Show a pop up warning you that you will lose progress if you go back to the home screen, do you want to anyway?

or

2. Just not do anything, the user can click cancel / the X in the corner instead

emmapainter:
This seems pretty standard for modal behaviour, probably needs a warning though.

PR #166, opened 2022-08-05
project name stays in text field when you go back

### LINKED ISSUE #160: Navigating back from import media to new project page removes project name
1. New project
2. Enter title
3. Continue
4. Back
5. Currently the project name text field is empty - it should save the title

emmapainter:
Elevated the state so that the project name can remain filled when the user continues to the next modal then navigates back to the new project view

## Summary

Lifted the project name state from the new project view into the modal container so that the project name persists when you continue to the import media view then navigate back again.

<hr/>

Currently the project name clears when the user navigates back to the project name view, which would be annoying if they only wanted to fix a simple typo.

<hr/>

### OS

- [x] Linux
- [x] MacOS
- [x] Windows

### DIFF
--- a/src/renderer/components/NewProjectModal/NewProjectView.tsx
+++ b/src/renderer/components/NewProjectModal/NewProjectView.tsx
@@ -12,6 +12,8 @@ import { PrimaryButton, SecondaryButton } from '../Blocks/Buttons';
 interface Props {
   closeModal: () => void;
   nextView: () => void;
+  projectName: string;
+  setProjectName: (projectName: string) => void;
 }
 
 const CustomStack = styled(Stack)({
@@ -36,8 +38,12 @@ const Container = styled(Box)({
   height: '200px',
 });
 
-const NewProjectView = ({ closeModal, nextView }: Props) => {
-  const [projectName, setProjectName] = useState<string>('');
+const NewProjectView = ({
+  closeModal,
+  nextView,
+  projectName,
+  setProjectName,
+}: Props) => {
   const [isAwaitingProjectName, setIsAwaitingProjectName] =
     useState<boolean>(true);
 
--- a/src/renderer/components/NewProjectModal/NewProjectView.tsx
+++ b/src/renderer/components/NewProjectModal/NewProjectView.tsx
@@ -1,7 +1,7 @@
 import { styled, Stack, Box, TextField, Typography } from '@mui/material';
 import CloseIcon from '@mui/icons-material/Close';
 import IconButton from '@mui/material/IconButton';
-import { ChangeEvent, useState } from 'react';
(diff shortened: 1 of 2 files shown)


ghost (approved):

exquisitebin:
I'm currently working on Issue #101, in which the idea is that the user is prompted with a dialog asking if they want to abandon the new project when they click outside of the Modal. Where they can choose to continue and not close the Modal, or close the Modal and start a new project again.

Not sure if remembering the previous project name clashes with this idea. Happy to hear thoughts on this.

exquisitebin (changes requested):
I'm currently working on Issue https://github.com/chloebrett/mlvet/issues/101 in branch [bugfix/click-outside-of-modal](https://github.com/chloebrett/mlvet/tree/bugfix/click-outside-of-modal). The idea is that the user is prompted with a dialog asking if they want to abandon the new project when they click outside of the Modal. Where they can choose to continue and not close the Modal, or close the Modal and start a new project again.

Not sure if remembering the previous project name clashes with this idea. Happy to hear thoughts on this.

MutazAshhab:
@exquisitebin If you will be showing a dialog and the user confirms they want to exit then you should use the `setProjectName` hook to re-set the `projectName` back to an empty string.

This PR has more to do with how the app behaves when you go back a step in the existing modal rather than what happens when you click off

MutazAshhab (approved):

exquisitebin (changes requested):
Mutaz clarified my query, and I agree it doesn't clash with the issue I'm working on.

I just have a bug that when you press back to name the new project the continue button is disabled until you change the text (which ideally you wouldn't have to).

image

fwang1703 (approved):
Nice!

emmapainter:
> Mutaz clarified my query, and I agree it doesn't clash with the issue I'm working on.
> 
> I just have a bug that when you press back to name the new project the continue button is disabled until you change the text (which ideally you wouldn't have to).
> 
> image

@exquisitebin this should be fixed now

exquisitebin (approved):
Looks good, awesome!

src/renderer/components/NewProjectModal/NewProjectView.tsx
```diff
@@ -60,17 +63,12 @@ const NewProjectView = ({ closeModal, nextView }: Props) => {
...
-    } else {
-      setIsAwaitingProjectName(true);
-    }
   };
 
   const continueButton = (
     <PrimaryButton
       onClick={handleContinue}
-      disabled={isAwaitingProjectName}
+      disabled={!projectName.trim()}
```
rileykeane, line 71: 👀

PR #168, opened 2022-08-06
Feature/drag text to move

chloebrett:
~~Depends on #156 and #169~~

Functionality:
- Holding alt/option lets you drag words around
- Potential drop locations (i.e. spaces between words) show themselves if the mouse is near them
- Some improvements to delete and paste ops to make them a bit more streamlined (they now handle selections themselves as part of the undo stack, so selected words work as expected between undos and redos)
- Works with undo/redo as expected

Known issues:
- Only supports dragging a single word at a time, for now
- ~~Dragging a word outside the transcription box produces undefined behaviour~~ - fixed
- Rendering is potentially inefficient during a drag (dragged word updates itself at 60fps, and all non-dragged words in the transcript at 30fps - can throttle further but this gets a little choppy). One option is to avoid sending mouse actions to words that are deemed to be a long way from the mouse pointer, but this is a little complex. Let's just play around with it for now, and see if the performance is acceptable or not - performance is not affected when not dragging so the rest of the app won't be affected
- Behaviour regarding the video preview current word indicator is undefined (this is going to be changed to a cursor soon anyway so I didn't spend too much time on it)
- ~~Sometimes makes a horizontal scrollbar appear in the transcription box, this is a small style fix but I haven't got around to it yet~~ - fixed

Demo video: (note - the 'instant edits' are when I use undo/redo)

https://user-images.githubusercontent.com/6735055/183241056-9d76f500-275b-40c1-817c-eada877875e0.mov


chloebrett:
Tested on my own linux machine, didn't notice any issues other than the known ones listed above

src/main/config.ts
```diff
@@ -2,4 +2,4 @@
 
 // Whether to use the sample transcription instead of actually transcribing.
 // Useful for testing.
-export const USE_DUMMY = false;
+export const USE_DUMMY = true;
```
jhea0008: This should probs be reset

src/renderer/components/WordDragManager.tsx
```diff
@@ -0,0 +1,149 @@
...
+) => JSX.Element;
+
+export type DragState = null | {
+  offset: Point;
+  wordIndex: number;
+};
+
+interface Props {
+  renderTranscription: RenderTranscription;
+  containerRef: RefObject<HTMLDivElement>;
```
Mhudlark, line 46: Works well and code looks good!

Just curious about this containerRef getting prop drilled right through the entire component hierarchy. Is it possible to use `document.body` or `document.getElementBy` with a use effect (inside the DragManager) instead?

rileykeane (approved):

src/renderer/components/WordDragManager.tsx
chloebrett: I agree the prop drilling isn’t ideal, I guess it would be possible to do native document methods though a bit ugly. Maybe react context instead? I know we already have redux but I think this is a pretty good use case for it

PR #169, opened 2022-08-06
Refactor ops to support multiple actions within a single undo/redo

chloebrett:
Small refactor to the undo stack to support each op containing an array of 'do' actions and an array of 'undo' actions.

This is a helpful abstraction for modelling certain actions as a combination of others. Namely, 'move' actions as a combination of 'cut' (without the actual copy-to-clipboard part, which isn't part of the undo stack anyway), and 'paste' to a new location. We may find other use cases for this in 'restore delete', etc. This allows us to simplify certain reducers.

This change is necessary because currently, if you dispatch multiple ops, each gets a discrete position on the undo stack which means that e.g. 'move' actions would take two undo/redo steps if modelled as a combination of cut + paste.


chloebrett:
Converted to draft as I've been using this for the move logic and realised undo actions should be in undo order, not reversed

ghost (approved):
Will the tests need to be updated for this? Reducer, etc?

chloebrett:
> Will the tests need to be updated for this? Reducer, etc?

No, because the outcome of running an op with a single element in the do array is the same as running an op before :)

PR #170, opened 2022-08-06
Chore/extract UI components

### LINKED ISSUE #119: Refactor: extract UI components
All UI components (in the format of styled\`...\`) should be separated from components with logic in them, and shared. Otherwise we keep reinventing the wheel doing things like this:

Screen Shot 2022-05-28 at 3 59 35 pm

(this is in a logic component file)

ghost:
## Summary

Replaces the styled component definitions with objects instead of giant css strings. 
Also replaced reused button code with reusable primary and secondary button components.

<hr/>

Makes the styling codebase more reliable. Replace reused code with components.

<hr/>


src/renderer/components/Blocks/Buttons.ts
```diff
@@ -0,0 +1,11 @@
...
+
+const StyledContainedButton = styled(Button)({
+  filter: 'drop-shadow(0px 8px 16px rgba(0, 0, 0, 0.8))',
+});
+
+export const PrimaryButton = styled(StyledContainedButton)``;
+PrimaryButton.defaultProps = { color: 'primary' };
+
+export const SecondaryButton = styled(StyledContainedButton)``;
+SecondaryButton.defaultProps = { color: 'secondary' };
```
fwang1703, line 11: nice!

fwang1703 (approved):
Love your work!

PR #173, opened 2022-08-09
Bugfix/video preview resize

### LINKED ISSUE #172: Video Preview Resize Page Resize Update
Basically the max width for the video preview resize is set on page mount.
If the user resizes the page, then the max width does not get updated, so the video preview can't get very big.

ghost:
## Summary

Fixed the bug where the video preview min/max width does not get updated when the page resizes.

Page resize fix includes:
- Min/Max width gets updated when page resizes
- If page is bigger than before, can then increase video preview to a larger size
- If page is smaller than before, then video preview should resize automatically to fit the new size (up to a point)

Also:
- Bug that existed before where dragging past max or past min caused problems with next resize should be fixed (`ondragstart` fix)

<hr/>

Fixes [this issue](https://github.com/chloebrett/mlvet/issues/172)


Irmy (commented):
I love the slider bar in the middle - The min size of the video works really well when dragging the slider to the left to make it smaller . I think a max size of the video should maybe be in place just so it doesnt overlap with the play/ skip elements - or do we want this? 

The slider bar made it super easy to resize but did we want the video to resize automatically when the window size is changed or was it intentional for the user to just resize both the window and video preview? 

https://user-images.githubusercontent.com/55219708/184560651-8b47d81d-3f82-41f0-b233-538a4963118f.mp4



Note: i was doing action in small bursts so zoom can record properly - everything worked super smooth.

Mhudlark:
@Irmy will look into this. There is a max set and it works on mine. There must be a bug. I'll let you know when I've fixed it.

Mhudlark:
Updated. Please have another look :)

rileykeane (approved):

Irmy (changes requested):
The max height works really well now!! 
I found a bug though sorry! - When you drag the window from the right ( horizontally) all the way closed and then all the way large again the video reappears as it should. However if you drag from the bottom ( vertically) all the way to a tiny window and then large again - the video completely disappears and the slider bar doesnt work to readjust. 

I refreshed the page entirely and went back into the project to get the video preview back - zoom recording is a bit janky

https://user-images.githubusercontent.com/55219708/184733899-b70bd656-1681-4be6-a991-a157a4748745.mp4

rileykeane (changes requested):
Same issue as Irmy. 

Video example [here](https://drive.google.com/file/d/1swbaTPlyx7i-fwTvJ3m9dHelhJYI4Ean/view?usp=sharing) (file is too big for GitHub)

ghost:
Ok I have now fixed this issue. Apologies for all the bugs. I am not creative enough to be a true tester. :)

Irmy (approved):
Look great! Works perfectly for horizontal video

rileykeane (approved):

PR #175, opened 2022-08-11
Add the ability to navigate using the Enter/NumpadEnter key

### LINKED ISSUE #47: Can't use enter/return to confirm project name
On the new project dialogue, when typing the project name, if enter/return is pressed, nothing happens. Instead it is required to press continue with the mouse. Keyboard key Enter/return should be able to be used as a replacement for pressing the continue button

MutazAshhab:
## Summary

You can now navigate through the new project modal using the "Enter" key (NumpadEnter) also works

Fixes: #47

<hr/>

Usability

<hr/>

### Notes?
* I re-introduced `isAwaitingProjectName` rather than using `projectName.trim()`
* I am using eventListeners which are set in the `useEffect` hook. (Hugh and I discussed wrapping the element with a `<form>` element [but that did not 100% work](https://github.com/chloebrett/mlvet/pull/49#issuecomment-1123114330))
<hr/>

### OS

- [x] Linux
- [x] MacOS
- [x] Windows


src/renderer/components/NewProjectModal/NewProjectView.tsx
```diff
@@ -45,30 +45,57 @@ const NewProjectView = ({
...
     const project = await makeProjectWithoutMedia(projectName);
     if (project === null) {
       return;
     }
     setProjectInStore(project);
     nextView();
-  };
+  }, [nextView, projectName, setProjectInStore]);
+
+  useEffect(() => {
```
chloebrett: This code is duplicated between the two pages - can we bring it up a level? Or maybe refactor into a custom hook

chloebrett (changes requested):

MutazAshhab:
Hey @chloebrett I extracted the repeated code into a custom hook... Now before you look at the code, I have to warn you, I dont know of an appropriate name for the hook, as well as the file name (should we put all hooks into a single file since we are not likely to have a lot of custom hooks or should we have 1 file per custom hook).

Tag me when you are done and I'll make the relevant changes.

ghost:
Not in `develop` yet but I have a hooks file coming in. Can copy and whoever is last can just merge them.
image

src/renderer/hooks/customHooks.ts
```diff
@@ -0,0 +1,25 @@
+import { useEffect } from 'react';
+
+function useCustomHook(
```
ghost: Nice work! Welcome to the truly wonderful world of react custom hooks where doing things that feel dumb makes YOU feel smart :)

I reckon you name the custom hook `useKeypress` and then instead of manually/statically checking 'Enter' and 'NumpadEnter', pass in an array of strings corresponding to keyboard press codes. Then in function check the array parameter (maybe `keycodes`) to see if `event.code` is in it.
Also, instead of `handleContinue` param, maybe rename it to something like `eventHandler` or if real basic, just `callback`.
This will make the hook more reusable without a huge amount of extra code.

Also, not sure if necessary for `keypress` event but if you're feeling up to it. Have a look at [my PR](https://github.com/chloebrett/mlvet/pull/173) where I use a debouncer (basically just spaces out function calls based on user input to stop app crashing) to stop the UI lagging/crashing on continuous input (not sure if a `keypress` event is a once off but if it isn't then the user holding down the key could cause lag).

src/renderer/components/NewProjectModal/NewProjectView.tsx
```diff
@@ -45,30 +46,42 @@ const NewProjectView = ({
...
+    } else {
+      setIsAwaitingProjectName(true);
+    }
   };
 
   const continueButton = (
     <PrimaryButton
       onClick={handleContinue}
-      disabled={!projectName.trim()}
+      disabled={isAwaitingProjectName}
```
emmapainter: This change brought back the bug mentioned in the comments of [this PR](https://github.com/chloebrett/mlvet/pull/166) - if you go back to the new project modal from the import media modal the continue button is disabled even though the project title field has text in it.

emmapainter (changes requested):

src/renderer/hooks/customHooks.ts
```diff
@@ -0,0 +1,25 @@
...
+  handleContinue: () => Promise<void>,
+  condition: boolean
```
emmapainter: Would love a more descriptive name on this - maybe keypressEnabled or something similar?

emmapainter:
Nice work!! I've been waiting for this one 😄 I think return should also work on the final modal for the "Get Started" button

src/renderer/components/NewProjectModal/NewProjectView.tsx
MutazAshhab: Mhmm I see, thank you for bringing that up

MutazAshhab: See this commit: [a7282be](https://github.com/chloebrett/mlvet/pull/175/commits/a7282be8840afa4db73aba0df4af6ca860454241)

src/renderer/hooks/customHooks.ts
MutazAshhab: Went with `useKeypress` in [2cfe8cd](https://github.com/chloebrett/mlvet/pull/175/commits/2cfe8cdf2d2f5cadb584f5f938d5cbaa3a43f3f3)

src/renderer/hooks/customHooks.ts
MutazAshhab: I did not end up implementing `useDebounce`. Here is the commit with the changes: [2cfe8cd](https://github.com/chloebrett/mlvet/pull/175/commits/2cfe8cdf2d2f5cadb584f5f938d5cbaa3a43f3f3).

I can already see how the code can become more reusable. Nice one Hugh

MutazAshhab:
@emmapainter I've added support for `Enter`/`NumpadEnter` on the final modal here: [983d495](https://github.com/chloebrett/mlvet/pull/175/commits/983d495423a45d214faf86a13e4de2d2bbdfa1a0)

I tested it out and it works. But i realised that by the time that the transcription is over your hands are off the keyboard and hitting `Enter` did not feel very natural. Test it out and let me know

MutazAshhab:
> Not in `develop` yet but I have a hooks file coming in. Can copy and whoever is last can just merge them. image

@hhen0005 here is commit with the refactoring of `customhooks` to `utils/hooks.ts` : [ff4bb96](https://github.com/chloebrett/mlvet/pull/175/commits/ff4bb9665a3142d634d407df566864d99f022df6)

src/renderer/utils/hooks.ts
```diff
@@ -0,0 +1,24 @@
...
+import { FunctionReturnVoid, FunctionReturnPromiseVoid } from 'sharedTypes';
+
+function useKeypress(
+  eventHandler: FunctionReturnPromiseVoid | FunctionReturnVoid,
+  condition: boolean,
+  keypressCodes: string[]
+) {
+  useEffect(() => {
+    const handleKeypress = async (event: KeyboardEvent) => {
+      if (keypressCodes.includes(event.code) && !condition) {
```
ghost: The hook looks much better. Nice!

One suggestion is change the input and usage of `condition`. It looks like the event handler only runs when `condition` is false, which seems a little confusing for other developers.

src/renderer/components/NewProjectModal/RunTranscriptionView.tsx
```diff
@@ -120,6 +121,12 @@ const RunTranscriptionView = ({ closeModal, nextView }: Props) => {
     dispatch,
   ]);
 
+  // we use !== instead of === since the useKeypress hook negates the condition that is given to it.
```
ghost: If you have to explain it... 😄

src/renderer/components/NewProjectModal/NewProjectView.tsx
```diff
@@ -45,24 +46,36 @@ const NewProjectView = ({
...
+  }, [nextView, projectName, setProjectInStore]);
+
+  useKeypress(handleContinue, isAwaitingProjectName, ['Enter', 'NumpadEnter']);
 
   const handleProjectNameInput = (
     event: ChangeEvent<HTMLInputElement | HTMLTextAreaElement>
   ) => {
-    setProjectName(event.target.value);
+    setProjectName(event.target.value.trim());
+    if (event.target.value !== '') {
```
ghost: Can this be simplified to something like `setIsAwaitingProjectName(event.target.value === '')` ?

src/renderer/components/NewProjectModal/NewProjectView.tsx
```diff
@@ -45,24 +46,36 @@ const NewProjectView = ({
   setProjectName,
 }: Props) => {
   const dispatch = useDispatch();
+  const [isAwaitingProjectName, setIsAwaitingProjectName] =
+    useState<boolean>(true);
```
ghost: Do we need an explicit type here?

src/renderer/components/NewProjectModal/ImportMediaView.tsx
```diff
@@ -66,9 +64,17 @@ const ImportMediaView = ({ prevView, closeModal, nextView }: Props) => {
...
-  };
+  }, [currentProject, mediaFilePath, nextView, setCurrentProject]);
+
+  useKeypress(handleTranscribe, isAwaitingMedia, ['Enter', 'NumpadEnter']);
+
+  if (currentProject === null) {
+    return null;
+  }
+
+  const projectName = currentProject.name;
```
ghost: Having a look at the code, this variable (which gets redefined on every render) is only used once. I reckon it might be better for readability to just change the single usage from `projectName` to `currentProject.name` and delete the variable.

src/renderer/components/NewProjectModal/RunTranscriptionView.tsx
MutazAshhab: i know... but if I fix the negation, then i'll have to explain two other usages of the `useKeypress` hook

src/renderer/components/NewProjectModal/NewProjectView.tsx
MutazAshhab: Nice catch: [c3a0de8](https://github.com/chloebrett/mlvet/pull/175/commits/c3a0de8b8c96366a77027e4b8d0f6624ec3897da)

src/renderer/components/NewProjectModal/RunTranscriptionView.tsx
ghost: maybe the conditional could be renamed to something like `shouldEventHandlerRun` (I think @emmapainter mentioned this but I can't find it anywhere). Then no need to explain usage anywhere.

Regardless, I reckon developers should always be able to presume conditional is positive. If that means explaining it twice...

src/renderer/components/NewProjectModal/ImportMediaView.tsx
MutazAshhab: Change made here: [a925e32](https://github.com/chloebrett/mlvet/pull/175/commits/a925e3217cf3cc53933cfd6ec287d1e8814190b2)

src/renderer/components/NewProjectModal/NewProjectView.tsx
MutazAshhab: As in using `useState(true)` instead of `useState<boolean>(true)`? If both work then it we don't need it. I thought you always had to have the type.

I'll change it soon

MutazAshhab: Here it is: [8ba7171](https://github.com/chloebrett/mlvet/pull/175/commits/8ba7171f6275fac423d921762e10a711e54fb490)

src/renderer/components/NewProjectModal/RunTranscriptionView.tsx
MutazAshhab: I thought about it more and I changed my mind, here is the change: [56ed2a8](https://github.com/chloebrett/mlvet/pull/175/commits/56ed2a81f3a3e2dce6254eed2d987f8738609f8f)

src/renderer/utils/hooks.ts
MutazAshhab: Here is the commit with the change: [56ed2a8](https://github.com/chloebrett/mlvet/pull/175/commits/56ed2a81f3a3e2dce6254eed2d987f8738609f8f)

src/renderer/hooks/customHooks.ts
ghost: I think @emmapainter was talking about renaming the `conditional`. I like that idea. `isKeypressEnabled` 😄

ghost (approved):
Nice work. Thanks for bearing with us through all the comments. 😃

src/renderer/hooks/customHooks.ts
MutazAshhab: Ahh i see, I updated it here: [e947eb3](https://github.com/chloebrett/mlvet/pull/175/commits/e947eb3ff17d018bc491c0d2526cf921bac915aa)

emmapainter:
> @emmapainter I've added support for `Enter`/`NumpadEnter` on the final modal here: [983d495](https://github.com/chloebrett/mlvet/pull/175/commits/983d495423a45d214faf86a13e4de2d2bbdfa1a0)
> 
> I tested it out and it works. But i realised that by the time that the transcription is over your hands are off the keyboard and hitting `Enter` did not feel very natural. Test it out and let me know

Yeah you might not use it in practice but I think it makes sense to have the option 🙂

src/renderer/components/NewProjectModal/NewProjectView.tsx
```diff
@@ -45,24 +46,31 @@ const NewProjectView = ({
...
     const project = await makeProjectWithoutMedia(projectName);
     if (project === null) {
       return;
     }
     setProjectInStore(project);
     nextView();
-  };
+  }, [nextView, projectName, setProjectInStore]);
+
+  useKeypress(handleContinue, !isAwaitingProjectName, ['Enter', 'NumpadEnter']);
```
emmapainter: You'll also need to use !projectName.trim() here, because at the moment when you go forward then back to this page it doesn't let you use the enter key until you've changed the value in the project name field

MutazAshhab: Hey Emma here is the change: [9033c84](https://github.com/chloebrett/mlvet/pull/175/commits/9033c84494d7d467e54bf2019bb9095aa396cb7c)

I was not able to use !projectName.trim() since the useKeypress hook no longer negates the useKeypressEnabled condition.

And I had to add !== '' since typescript was not happy when I passed a string instead of boolean variable.

src/renderer/utils/hooks.ts
```diff
@@ -1,5 +1,6 @@
 import { useDebounce } from '@react-hook/debounce';
 import { useEffect } from 'react';
+import { FunctionReturnVoid, FunctionReturnPromiseVoid } from 'sharedTypes';
```
chloebrett: What's the point of these? 

`() => void` and `() => Promise<void>` are concise and used all over the place already

chloebrett (approved):
Looks good and works well when tested locally. Just a bit iffy on the helper types, they don't seem that useful

src/renderer/utils/hooks.ts
MutazAshhab: You're right we mainly use `() => Promise<void>` throughout the project, however, the `nextView()` function is of type `() => void`

chloebrett: That’s fine, what I mean is I don’t feel the helper types add much value

chloebrett: For example you wouldn’t make a helper type

`type ArrayOfStrings = string[]`

because the name is more verbose than the actual type it’s describing. But a helper type is useful for more complex things, e.g. `MapCallback` is a good example.

I feel that FunctionReturningPromiseVoid and the other one are closer to the first category

chloebrett: Notable exception to this rule: helper types that serve to clarify a specific use of a general type.

For example a selection is currently represented as a list of word indices i.e. number[] but it could be helpful to make a helper `type Selection = number[]` because otherwise number[] doesn’t really explain what it is used for

MutazAshhab: I misread your first comment, I added these helper types after `eslint` got mad and suggested refactoring the parameter types into a helper type.

Im going to remove these types and just ignore the eslint rule.

MutazAshhab: I've looked at it again it was a ts rule not a eslint rule ***

chloebrett: Oh weird yeah, makes sense. Your solution looks fine

emmapainter (approved):
Love your work, super impressed as always by how well you take feedback <3

PR #177, opened 2022-08-12
Support dragging to select words

chloebrett:
Depends on #168 

Demo:

https://user-images.githubusercontent.com/6735055/184303355-b85b874a-6b19-4c60-ad6f-27d4e41482a5.mov

Generally performant below a few hundred words, once you start copying a lot and end up with 500+ words it gets a bit slow (though the whole app is not really optimised for this in general right now). I tried profiling to get some quick wins but to be honest it's more of a death by a thousand cuts situation (e.g. each render of a word might only take 0.1ms but when there are 500 of them we slow down to 20fps). There are a number of solutions to this but none are particularly trivial so I'd say leave it for now

### DIFF
--- a/src/main/config.ts
+++ b/src/main/config.ts
@@ -2,4 +2,4 @@
 
 // Whether to use the sample transcription instead of actually transcribing.
 // Useful for testing.
-export const USE_DUMMY = false;
+export const USE_DUMMY = true;
--- a/src/main/config.ts
+++ b/src/main/config.ts
@@ -2,4 +2,4 @@
 
 // Whether to use the sample transcription instead of actually transcribing.
 // Useful for testing.
-export const USE_DUMMY = true;
+export const USE_DUMMY = false;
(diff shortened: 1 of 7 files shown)


chloebrett:
Riley says: can we make the selection visually continuous

Made an issue for this, I'd like to do this after this PR is merged https://github.com/chloebrett/mlvet/issues/185

emmapainter:
Love that we have custom highlighting now! Is it possible to get rid of the native blue highlighting in the transcription?

chloebrett:
> Love that we have custom highlighting now! Is it possible to get rid of the native blue highlighting in the transcription?

It's not actually native, this was just a drop in replacement to show the currently active word in the video preview. I believe the plan here is to eventually change it to a cursor, but not part of this PR

emmapainter:
> > Love that we have custom highlighting now! Is it possible to get rid of the native blue highlighting in the transcription?
> 
> It's not actually native, this was just a drop in replacement to show the currently active word in the video preview. I believe the plan here is to eventually change it to a cursor, but not part of this PR

Oh I see it now 🤯

chloebrett:
Lmao what

emmapainter (approved):

PR #178, opened 2022-08-12
Feature/correct words

chloebrett:
Depends on #208

Features

- Double-click a word to open an edit box
- Type in the edit box then hit enter to save the edited word (marking a 'corrected' word)
- Supports undo/redo as expected
- Works as expected with other features

Known issues (mostly fixed)

- [x] Prevent multiple words being edited at once
- [ ] Support Ctrl+A to select all text in edit box -> this is actually ridiculously difficult with our architecture; left/right arrow keys and shift/ctrl selection in the textbox work but we've overridden Ctrl+A for another purpose and getting it to also work for inside-text-box selection is far from easy (I think we'd have to inject the ipc receivers into the react context and then use imperative handles or something crazy) - if you can think of an easier way please let me know
- [x] Prevent usual dragging, etc while edit is open
- [x] Cancel edit when clicking away
- [x] Modify input styles to look better / non default
- [ ] Possible memory leak on double-click timeout -> I don't think this is an issue

Demo with the ticked-off issues fixed, showing the feature in context with other features (merge, split, select, undo, redo, etc):

https://user-images.githubusercontent.com/6735055/185909918-f1accbe5-46a2-4be4-9e3b-f9904ceb723b.mov


src/renderer/components/Editor/TranscriptionBlock.tsx
```diff
@@ -45,14 +48,36 @@ const TranscriptionBlock = ({
...
+    () => (editWord === null ? new Set(selectionArray) : new Set()),
+    [selectionArray, editWord]
+  );
 
   const dispatch = useDispatch();
 
+  const submitWordEdit: () => void = () => {
+    if (
+      editWord !== null &&
+      editWord.text !== '' &&
```
fwang1703: Not sure if this is expected behaviour (so correct me if I'm wrong), but currently when a word is double clicked to be edited, and then the entire word is deleted via backspace button, pressing enter brings back the entire word. I would assume if a user manually deletes the word (even though we have deleting as a feature), they don't want the word to be there.

src/renderer/utils/ui.ts
```diff
@@ -33,3 +33,60 @@ export const getAspectRatio: (
   const elemSize = getElementSize(elem);
   return elemSize ? elemSize.width / elemSize.height : defaultAspectRatio;
 };
+
+/**
+ * Measure text logic adapted from
+ * https://stackoverflow.com/questions/118241/calculate-text-width-with-javascript/21015393#21015393
+ */
+
```
fwang1703, line 41: Nice! Love the work being done below

src/renderer/utils/ui.ts
```diff
@@ -33,3 +33,60 @@ export const getAspectRatio: (
...
+  }
+
+  const fontWeight = getCssStyle(element, 'font-weight') || 'normal';
+  const fontSize = getCssStyle(element, 'font-size') || '16px';
+  const fontFamily = getCssStyle(element, 'font-family') || 'Times New Roman';
+
+  return `${fontWeight} ${fontSize} ${fontFamily}`;
+};
+
+export const INPUT_FONT = 'bold Rubik 16px';
```
fwang1703: Should the alternative font weight/size/family correspond with the `INPUT_FONT`?

fwang1703:
Nice work! Just some small comments. Also not sure if this is the same addressed bug or not from your known issues list (and wasn't sure where to put it in the code review), but when a word has been edited but still in the 'edit' mode, doing a drag selection anywhere will just save the corrected word but not create the drag selection.

chloebrett:
> Nice work! Just some small comments. Also not sure if this is the same addressed bug or not from your known issues list (and wasn't sure where to put it in the code review), but when a word has been edited but still in the 'edit' mode, doing a drag selection anywhere will just save the corrected word but not create the drag selection.

I think this behaviour makes sense - basically any clicks outside of the edit field should close it before doing their primary action, otherwise you end up with weird state like a bunch of things selected while the edit is still open. Thoughts?

chloebrett:
Also I don’t think I can see your comments on the code itself if there are any?

fwang1703 (commented):
Oops forgot to submit my review comments that's why, should be able to see them now

src/renderer/components/Editor/TranscriptionBlock.tsx
chloebrett: It’s expected, because the default behaviour without this safety check is that the word’s text just becomes “” and the word takes up space without being able to be deleted, selected or edited.

i like your idea better though - why not just let this be an alternative way of deleting the word! I’ll implement that

src/renderer/utils/ui.ts
chloebrett: Good idea, will do

Irmy (commented):
Love this feature! Works so well and didnt have any of the issues you listed which is awesome. Ctrl + A works within the editing text field for me which was good. Just waiting on approval from Zac

emmapainter:
Just a couple of comments on the functionality 

- if you're editing a word and you click outside of the text box but not inside the transcription, it defocuses the text box but doesn't save the changes 
- if you have an edit box open and you try to highlight text, the highlight appears after you release the mouse. I'm actually not sure what makes the most sense here though - not highlighting and only closing the edit box, the highlight appearing on mouse up (current state), or highlight working the same as it usually would... @Irmy do you have thoughts on this?

Irmy:
Good pick up Emma - for the first dot point - this should behave the same as if you clicked elsewhere in transcription - so the changes should be saved and editing box should disappear. 

For the second dot point - definitely an interesting cross over of functions. If someone is editing a word and then goes to highlight elsewhere, which I am interpretting them as them abandoning that action, any word changes should be abandoned, editing word field should be exited and highlight action should be prioritised. Not sure if the rendering order of elements is going to be an issue when switching between actions. 

If this is an issue @chloebrett  is there a way to block highlighting/ other editing actions while in word correction state? That way they just have to click elsewhere in the UI  and then continue highlighting once it is no longer in word correction state.

chloebrett:
Good pickups Emma, I know exactly what is causing them and can fix accordingly.

For #1, will set click-outside-to-cancel receiver to be the root ref instead of the transcription box.

For #2, will clear the selection when the edit box closes so that highlights behave like other actions - see comment way above

> any clicks outside of the edit field should close it before doing their primary action, otherwise you end up with weird state like a bunch of things selected while the edit is still open

In both cases, UX may still be imperfect but I think they are both improvements on the issues you've found

chloebrett:
Remaining issues check off

- [x] simplify logic for determining default font
- [x] delete word if its text is emptied
- [x] allow clicking anywhere on screen to cancel edit
- [x] cancel highlight when edit finishes

Irmy (approved):
Thanks for fixing those up! Love it

PR #187, opened 2022-08-17
Feature/edit marker

fwang1703:
## Summary

This PR is adding an EditMarker component which is rendered when a word is deleted in the transcription. This is so that users will know where original transcription has been deleted or moved.

This change is needed to implement the feature which allows users to restore text when it has been deleted. The marker acts as a visual aid to users to know where text has been deleted/moved. 

### What did you change?

- Created the `EditMarker` component
- Added logic to the `TranscriptionBlock` to determine when to render an EditMarker

https://user-images.githubusercontent.com/40136329/185019784-df9b174e-5d02-4f3d-9dc8-1022e959fbe6.mov

Steps in video
- Delete a word and see EditMarker
- Delete words before and after the current EditMarker (they will combine into 1 EditMarker)
- Show EditMarkers in different spots of the transcription (as long as there is 1 word in between deleted words)
- Copying a word somewhere else and deleting it. Does not show EditMarker as it is not the original spot
- Show that EditMarker does not disappear if transcription is edited
- EditMarker disappears if original word returned to its place (via copy and pasting)

### Design Docs

https://www.figma.com/file/vPh49OC7hJpMD7cQMAnBUo/UI-Design?node-id=1034%3A2147 

### OS

- [ ] Linux
- [x] MacOS
- [ ] Windows


emmapainter (approved):

src/renderer/components/EditMarker.tsx
```diff
@@ -0,0 +1,23 @@
+const EditMarker = () => {
+  return (
+    <svg
```
chloebrett: Nice 🌟

src/renderer/components/TranscriptionBlock.tsx
```diff
@@ -2,12 +2,13 @@ import styled from '@emotion/styled';
 import { Box } from '@mui/material';
 import { Fragment, RefObject, useMemo } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
-import { Transcription } from 'sharedTypes';
+import { Transcription, Word as WordType } from 'sharedTypes';
```
chloebrett: I like the idea of differentiating between the `Word` react component and the `Word` type/interface - maybe we should do this more generally, rather than in this file, because I can foresee this being an issue for other changes as well. Perhaps the easiest way would be to rename the react component to be something like `WordComponent` instead though, as there will be fewer places to update. What do you think?

src/renderer/components/TranscriptionBlock.tsx
```diff
@@ -79,6 +80,31 @@ const TranscriptionBlock = ({
       []
     );
 
+  const renderEditMarker = (
```
chloebrett: I think this is probably a good candidate for its own react component, to avoid cluttering up the TranscriptionBlock file. Thoughts?

src/renderer/components/TranscriptionBlock.tsx
```diff
@@ -79,6 +80,31 @@ const TranscriptionBlock = ({
...
+    // word index has changed but still in the same relative position
+    const hasNotMoved =
+      index !== 0
+        ? transcription.words[index - 1].originalIndex ===
+          word.originalIndex - 1
+        : transcription.words[index + 1].originalIndex ===
+          word.originalIndex + 1;
+
+    // preventing two markers from being next to each other
+    const hasNoNeighbourMarker =
```
chloebrett: There is a small bug here - to replicate:

- delete the first word
- delete the second word

This creates no marker, but it should create one marker.

chloebrett (changes requested):
Looks pretty good, just a few stylistic suggestions and one bug fix.

There is also another bug. To replicate:
- Delete one word
- Cut a range from the word before it to the word after it (including the delete marker)
- This will leave the delete marker there
- Then paste to another place in the text
- This will create a second delete marker

When this happens, you end up with two delete markers - the preferred behaviour is that the one in the original location should be there, but the one in the pasted text shouldn't be.

I'm happy for this PR to be merged without that bug being fixed, as I think it will take a lot more thinking and some changes to the way cut/paste/delete work. (i.e. not pasting deleted words, marking words as deleted when cut instead of actually removing them, and maybe some other changes)

rileykeane (changes requested):
Looks good, just a few little things in addition to Chloe's comments. 

- It would be nice if the edit marker was bottom aligned with the word
Screen Shot 2022-08-19 at 11 27 59 AM

- The edit markers will need a key since we can have multiple rendered simultaneously. React is currently throwing an warning about this.

src/renderer/components/TranscriptionBlock.tsx
fwang1703: Updated! React component has been refactored to `WordComponent`

src/renderer/components/TranscriptionBlock.tsx
fwang1703: Moved to EditMarker react component file

src/renderer/components/TranscriptionBlock.tsx
fwang1703: Did a fix - seems to fix this issue but haven't tested thoroughly enough to know if it has created any other bugs

fwang1703:
@rileykeane Is this alignment better?
Screen Shot 2022-08-22 at 1 56 48 pm

Also for the EditMarkers, I've added a key for now, but may need to work with yourself and @emmapainter to change it for the restore text component

src/renderer/components/Editor/EditMarker.tsx
```diff
@@ -0,0 +1,32 @@
+import { Transcription, Word } from 'sharedTypes';
+import EditMarkerSvg from '../../assets/EditMarkerSvg';
+
+interface Props {
+  transcription: Transcription;
```
chloebrett, line 5: I believe passing this as a prop means that if the transcription changes at all, every single edit marker will re-render.

Perhaps instead pass `prevWord: Word | null` and `nextWord: Word | null` - that way, they will only re-render when relevant changes happen.

chloebrett (approved):
Looks good aside from the one comment I left

rileykeane (approved):

PR #188, opened 2022-08-17
Feature/video scrubber

rileykeane:
## Summary
Added a video scrubber
- Can be used to move the video forwards and backwards and gives the user finite control over playack
- Displays the current playback time and total video duration
- Clicking the left-most timestamp toggles between total duration and time remaining

**Note:** There is a bug when the project first loads; if you drag the scrubber to a position other than the start and hit play, the scrubber (and clock) resets back to 0, but the video continues to play from the new position. This bug existed before the scrubber was implemented (try the same thing by clicking a word on a different branch) and will need to be fixed in a separate PR. 

<hr/>

Zac wanted it. 

<hr/>

### What did you change?
- Created a new scrubber bar component (this uses the unstyled slider from MUI) 
- Added the scrubber bar to the project page under the video preview
- Updated the transcript interface to include a new field, `outputDuration`. This is the current duration of the edited video and is updated after every edit to the transcript
- Refactored `updateOutputStartTimes` to also update `outputDuration`. I also renamed both the function and file to `updateOutputTimes` as I thought it was a more fitting name given it now updates both the `outputStartTimes` and `outputDuration`
-  Refactored the transcript reducers to update `outputDuration` on any changes to the transcript 
- Various other small updates to account for the new field in the transcript interface 
- Added new test cases and updated some existing ones 
<hr/>

### Video demo

https://user-images.githubusercontent.com/27614346/185095050-95180ebc-179e-4171-9fbf-08c309efcf7f.mov

<hr/>

### OS

- [ ] Linux
- [x] MacOS
- [x] Windows


ghost (approved):
There is one bug that I noticed, not related to this PR. If the user deletes words before the current time (eg: user selects lots of words at start and deletes them after the playback has progressed past there), then the clock (and hence, the scrubber) does not update.

I have created [an issue](https://github.com/chloebrett/mlvet/issues/189) for this bug and assigned myself to fix.

src/renderer/components/Scrubber.tsx
```diff
@@ -0,0 +1,139 @@
...
+const Scrubber = ({
+  totalDuration,
+  currentTimeSeconds,
+  onScrubberChange,
+}: ScrubberProps) => {
+  const [showRemainingTime, setShowRemainingTime] = useState(false);
+
+  const sliderValue = useMemo(() => {
+    // Prevent the slider from going past the end of the video
+    const currentTime =
```
chloebrett: `Math.min(currentTimeSeconds, totalDuration)`

src/renderer/components/Scrubber.tsx
```diff
@@ -0,0 +1,139 @@
...
+      currentTimeSeconds > totalDuration ? totalDuration : currentTimeSeconds;
+
+    // Round to two dp to avoid precision errors
+    const percentageCompleted =
```
chloebrett: Make a util for this. Also, what's epsilon needed for?

src/renderer/components/Scrubber.tsx
```diff
@@ -0,0 +1,139 @@
...
+    const remainingSeconds = duration - currentTime;
+
+    // Sometimes the actual video playback timer (currentTime) is ever so slightly more then the
+    // transcript total duration. This causes the remaining time to be negative. In this case we
+    // just return 0.
+    if (remainingSeconds < 0) {
+      return `-${secondToTimestampUI(0, false)}`;
+    }
+
+    return `-${secondToTimestampUI(remainingSeconds, false)}`;
```
chloebrett: Just make the whole call

`secondToTimestampUI(Math.max(0, remainingSeconds), false)`

Then no need for conditional

src/renderer/pages/Project.tsx
```diff
@@ -167,6 +168,11 @@ const ProjectPage = ({ containerRef }: Props) => {
     }
   };
 
+  const onScrubberChange = (newTime: number) => {
+    const nSeconds = newTime - time;
+    videoPreviewControllerRef.current?.seekNSeconds(nSeconds);
```
chloebrett: Why do we need `seekNSeconds`? Can't you just `setPlaybackTime`? Otherwise you're making an absolute time relative, and then making that relative time absolute again which is a bit redundant

src/transcriptProcessing/updateOutputTimes.ts
```diff
@@ -43,10 +43,30 @@ export const calculateTime: (
   return roundToMs(prevWord.outputStartTime + bufferedWordDuration(prevWord));
 };
 
+const updateOutputVideoDuration: (words: Word[]) => number = (words) => {
+  const currentTranscriptWords = words.filter((word) => word.deleted === false);
+  const lastWord = currentTranscriptWords.pop();
```
chloebrett, line 48: `pop` mutates, which we want to avoid in redux. Instead check the length and if length > 0 then read the final word

src/transcriptProcessing/updateOutputTimes.ts
```diff
@@ -59,5 +79,8 @@ export const updateOutputStartTimes: (words: Word[]) => Word[] = (words) => {
     });
   });
 
-  return newWords;
+  return {
```
chloebrett, line 77: Nice 👍🏻

chloebrett (changes requested):
Looking really good, just a few comments. Functionality works really well!

src/renderer/components/Scrubber.tsx
rileykeane: done.

src/renderer/components/Scrubber.tsx
rileykeane: Moved into util and the epsilon has been removed.

src/renderer/components/Scrubber.tsx
rileykeane: done.

src/renderer/pages/Project.tsx
rileykeane: Good catch. This now just calls setPlaybackTime with the new time returned by the scrubber.

Irmy (changes requested):
Scrubber bar doesnt seem to update according to how far through the video is. For example its 3secs into a 7 second video - I thought scrubber circle should then be approx 50% of the way along the bar. 
Is this the bug you were talking about? The top timestamp and the scrubber timestamp is accurate and the video correctly displays along with the timestamp but the bar doesnt allow me to drag along it or see a progression. 

https://user-images.githubusercontent.com/55219708/185556887-4e3df9f6-151f-4ce5-89ac-f6cf6413bfdc.mp4

Windows OS

rileykeane:
> Scrubber bar doesnt seem to update according to how far through the video is. For example its 3secs into a 7 second video - I thought scrubber circle should then be approx 50% of the way along the bar. Is this the bug you were talking about? The top timestamp and the scrubber timestamp is accurate and the video correctly displays along with the timestamp but the bar doesnt allow me to drag along it or see a progression.
> 
>  video1949141754.mp4 
> Windows OS

@Irmy Is this with a new project or opened from recents? I've added a new field to the transcription interface so previously created projects will have an issue with the scrubber bar.

src/transcriptProcessing/updateOutputTimes.ts
chloebrett: Resolving as we had a convo about this

src/transcriptProcessing/updateOutputTimes.ts
```diff
@@ -43,10 +43,30 @@ export const calculateTime: (
...
 };
 
+const updateOutputVideoDuration: (words: Word[]) => number = (words) => {
+  const currentTranscriptWords = words.filter((word) => word.deleted === false);
+  const lastWord = currentTranscriptWords.pop();
+
+  if (lastWord) {
+    return roundToMs(
+      lastWord.outputStartTime +
+        lastWord.duration +
```
chloebrett: There's a helper that will make this simpler:

```
return roundToMs(lastWord.outputStartTime + bufferedWordDuration(lastWord));
```

chloebrett (approved):
Looks good! There have been a few refactor style changes on develop so you'll probably get a few conflicts - should be mostly fine aside from the transcription reducer test file. Let me know if you need a hand fixing the conflicts

src/transcriptProcessing/updateOutputTimes.ts
rileykeane: done. Thanks for the handy helper!

chloebrett (approved):

Irmy (approved):
Oh beautiful - All works perfectly with a new project!

PR #197, opened 2022-08-19
Sort UI components into folders

chloebrett:

PR #198, opened 2022-08-19
Refactor playback and resize management out of project page

chloebrett:
Depends on #197

Render props time 😎 

No functional changes, just a refactor
- New component PlaybackManager creates, stores and handles video playback state (rather than the Project page itself)
- New component ResizeManager does the same for video resize and aspect ratio


ghost (approved):
Thanks for refactoring this. I should've done it. Render props, very sexy 😍

PR #205, opened 2022-08-22
Chore/ci publish packaged app

ghost:
## Summary

This PR adds in a GitHub workflow (CI script) that builds the app and publishes a release containing a windows, mac and linux version.

To actually properly test this PR, not just on a forked repo like I did, we will need to merge `develop` into `uat` (this should require a pull request)

<hr/>

We need to have easily accessible versions for user during the UAT process

<hr/>

#### This workflow should only run when someone pushes to the `uat` branch. Pushing directly to `uat` should be blocked (we should do that if it isn't) so in fact this workflow will only run when a PR to merge `develop` into `uat` has been approved and merged.

<hr/>

### Note: 
There is some hacky pre-package code in there to take the repo package.json version and pass it in to electron-builder as metadata. This is because for some reason the repo package.json version prop will not override the core electron-react-boilerplate package.json version prop like every other prop (annoying....).

<hr/>

### OS

- [x] Linux
- [x] MacOS
- [x] Windows

<hr/>

Please ignore my commit names 😃


.github/workflows/publish.yml
```diff
@@ -0,0 +1,114 @@
...
+      - uses: actions/checkout@master
+        with:
+          ref: ${{ github.ref }}
+
+      # Installing node.js
+      - uses: actions/setup-node@master
+        with:
+          node-version: 16
+
+      # Installing dependencies with higher timeout in case of slow network
```
chloebrett, line 96: Why would the network be slow? If it's in a github actions runner it will have a 10 gigabit connection right? Guess it's not really an issue though, just curious

package.json
```diff
@@ -1,21 +1,26 @@
 {
   "name": "mlvet",
   "description": "Machine Learning Video Editor Toolkit",
+  "version": "0.0.1",
```
chloebrett, line 4: Check Irmy's PR about changing the name etc, might be some conflicts here

chloebrett (commented):
Code looks fine, haven't tested the build locally yet

MutazAshhab:
I've tested it both locally and and on github. Both worked flawlessly.

Here is my github test: https://github.com/Taz17/mlvet-temp/runs/7946346645?check_suite_focus=true

Here is my local test: 
image

PR #208, opened 2022-08-22
Refactor ops out of undo stack

chloebrett:
Bit of a cleanup, ops file getting big and will be even bigger when correct words op is done. Also removed some accidentally duplicated logic


ghost (approved):
I like it. More organisation 😃

PR #211, opened 2022-08-25
Feature/keyboard shortcuts modal

fwang1703:
## Summary

Creating a keyboard shortcuts modal. This can be accessed via the menu bar.
image

Provides all the keyboard shortcut information, which will be useful for the live demo. 

### What did you change?

* New shortcuts dialog created `KeyboardShortcutsDialog`
* Added an IPC receiver to call the modal from the backend 

https://user-images.githubusercontent.com/40136329/186603850-40e0fb01-a155-4023-8e47-f76ee68baf97.mov

### Links to any design documents which you made/used in order to make the decisions that you made.

https://www.figma.com/file/vPh49OC7hJpMD7cQMAnBUo/MLVET-UI-Design?node-id=1355%3A3448

### OS

- [ ] Linux
- [x] MacOS
- [ ] Windows

I need to also have this tested in Windows and Linux too!


chloebrett:
Can you make it show the correct shortcut _based on_ OS, rather than both? Can use `osQuery` handler to retrieve this - Windows and Linux both use ctrl, Mac uses cmd

src/renderer/pages/Home.tsx
```diff
@@ -1,17 +1,31 @@
...
 import { ApplicationStore } from '../store/sharedHelpers';
 
 const HomePage = () => {
+  const dispatch = useDispatch();
+
   const hasRecentProjects = useSelector(
     (store: ApplicationStore) => store.recentProjects.length > 0
   );
 
+  const hasOpenedShortcuts = useSelector(
```
chloebrett: Rather than duplicating this logic per page, why not bring it up a level (not sure where exactly, probably App.tsx?)

If you do this be aware the redux consumer must be a separate component from the one that contains the store provider.

src/main/menu.ts
```diff
@@ -344,6 +344,46 @@ export default class MenuBuilder {
...
+          label: 'Keyboard Shortcuts',
+          click: () => {
+            this.mainWindow.webContents.send('open-shortcuts');
+          },
+        },
+        { type: 'separator' },
+        {
+          label: 'Learn More',
+          click: () => {
+            shell.openExternal('https://electronjs.org');
```
chloebrett: Maybe remove the links to electron stuff (I know not strictly related to your PR)

src/main/menu.ts
```diff
@@ -428,29 +469,36 @@ export default class MenuBuilder {
         id: 'help',
         label: 'Help',
         submenu: [
+          {
```
chloebrett: If you just make the subMenuHelp thing a regular `MenuItemConstructorOptions`, then you can use that here instead, and save on duplication

src/renderer/components/KeyboardShortcutsDialog.tsx
```diff
@@ -0,0 +1,175 @@
...
+            paddingTop="10px"
+            paddingBottom="10px"
+            color={colors.white}
+          >
+            Editing
+          </Typography>
+          <CustomRow
+            wordsLeft={['Cut', 'Copy', 'Paste', 'Undo', 'Redo']}
+            shortcutsLeft={[
+              'Ctrl + X/⌘ + X',
```
chloebrett: If you split the shortcuts based on OS you could just do it as something like (pseudocode below)

`const CTRL_OR_CMD = isMacOs ? '⌘' : 'Ctrl';`

Then just interpolate strings like \`${CTRL_OR_CMD} + X\`

chloebrett (changes requested):
Looks pretty good just a few comments. UI is nice!!

Irmy (approved):
Absolute king - thanks so much for picking this up. Only thing - are we able to get rid of the white rounding on the corners of the modal?  Or change it to a deeper grey?

src/renderer/pages/Home.tsx
fwang1703: I've chucked the dialog into `AppContents` in App.tsx, hope this is okay

src/main/menu.ts
fwang1703: Good pickup don't know why I didn't do this! Updated now

src/renderer/components/KeyboardShortcutsDialog.tsx
fwang1703: Updated! Thank you!!

src/renderer/navigation/openShortcuts.tsx
```diff
@@ -0,0 +1,10 @@
+import { toggleShortcuts } from 'renderer/store/shortcuts/actions';
+import store from 'renderer/store/store';
+
+const { dispatch } = store;
+
+const openShortcuts: () => any = async () => {
```
chloebrett: Return void not any

src/renderer/components/KeyboardShortcutsDialog.tsx
```diff
@@ -0,0 +1,205 @@
...
+            </Typography>
+          </CustomListItem>
+        ))}
+      </List>
+    </CustomColumnStack>
+  );
+};
+
+const KeyboardShortcutsDialog = ({ open, onClose }: Props) => {
+  const [currentOs, setCurrentOs] = useState<OperatingSystems | null>(null);
```
chloebrett, line 105: OS is not something that ever changes, so I’d almost be tempted to say make this a global var that loads outside the react context - this is fine for now though

src/renderer/components/KeyboardShortcutsDialog.tsx
```diff
@@ -0,0 +1,205 @@
...
+              paddingTop="0px"
+              color={colors.grey[300]}
+            >
+              {word}
+            </Typography>
+          </CustomListItem>
+        ))}
+      </List>
+      <List>
+        {shortcutsRight.map((shortcut: string) => (
```
chloebrett: There’s a bit of repetition of the rendering here - e.g. the two word lists render exactly the same as each other, and so do the shortcut lists. Plus there are only very minor differences between the word and shortcut list rendering. Can you try to abstract it a little - should be pretty quick and easy. Apologies for not catching this before

chloebrett (approved):
I’m going to approve since the suggestions are really minor - looks great!

ghost (approved):
Love it

PR #214, opened 2022-08-25
Cloud config view

MutazAshhab:
Frame 10 [from the figma mock ups](https://www.figma.com/file/vPh49OC7hJpMD7cQMAnBUo/MLVET-UI-Design?node-id=1125%3A2863)

What has been done:
1. UI for that component.
5. Ability to navigate with enter for that component
6. Ability to open the URL in an external browser
7. Ability to pass text to the component to display in the case that we want to show a different message (handy when Manikas part comes in)
8. Store API key and client secret in a file on disk.

Latest screenshot: 
image

Note: Screenshots below are outdated. Refer to this ^ one as shows the most current version of this feature


chloebrett:
> We still need to figure out whether we will save the API key/client secret or ask the user for those details every time. Which will be addressed in [25th August SA standup](https://docs.google.com/document/d/1AoJfLn5sqjlIaW6e-FC2SFtx4RT50_pvFmQt79Q39Gc/edit#heading=h.wuiglbgijs8y)

Save them. It will be a PITA for the user to enter them every time, and every app with auth keys since the dawn of time saves them in plaintext on the user's computer. What's the issue?

src/main/handlers/file/openLinkInExternalWindow.ts
```diff
@@ -0,0 +1,9 @@
+import { shell } from 'electron';
+
+type OpenExternalLink = (url: string) => Promise<void>;
```
chloebrett, line 3: 👍🏻

src/renderer/components/ProjectCreation/CloudConfigView.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+const CloudConfigView = ({
+  prevView,
+  closeModal,
+  nextView,
+  projectName,
+  textToDisplay,
+}: Props) => {
+  const [isAwaitingAPIKey, setAwaitingAPIKey] = useState<boolean>(true);
+  const [isAwaitingClientSecret, setAwaitingClientSecret] =
```
chloebrett: Do these actually need to be their own state? It looks like they're pure functions of the apiKey and clientSecret values

src/renderer/components/ProjectCreation/CloudConfigView.tsx
```diff
@@ -0,0 +1,168 @@
...
+    useState<boolean>(true);
+  const [APIKey, setAPIKey] = useState<string>('');
```
chloebrett: `apiKey` and `setApiKey` - treating acronyms as lowercase words is the generally accepted way to use them in variable names (see https://stackoverflow.com/questions/15526107/acronyms-in-camelcase)

src/renderer/components/ProjectCreation/CloudConfigView.tsx
```diff
@@ -0,0 +1,168 @@
...
+              value={APIKey}
+              onChange={(event) => handleAPIKeyInput(event)}
+              autoFocus
+            />
+          </CustomStack>
+          <CustomStack>
+            <TextField
+              label="Client Secret"
+              value={clientSecret}
+              onChange={(event) => handleClientSecretInput(event)}
```
chloebrett: Since the event being passed in exactly matches your function signature you can make it point-free

`onChange={handleClientSecretInput}`

That said, a more common pattern is to extract the value from the event before passing it to the handler, since the handler isn't doing anything else with the event object. i.e.

`onChange={(event) => handleClientSecretInput(event.target.value)}`

is something you tend to see a lot in react form code

Obviously, same goes for the other one too

src/renderer/components/ProjectCreation/ModalContainer.tsx
```diff
@@ -92,6 +94,16 @@ const ModalContainer = ({ isOpen, closeModal }: Props) => {
...
           />
         );
+      case CloudConfigView:
+        return (
+          <CloudConfigView
+            prevView={prevView}
+            closeModal={showCancelProject}
+            nextView={nextView}
+            projectName={projectName}
+            textToDisplay={null}
```
chloebrett, line 122: You can just make this an optional prop instead of explicitly passing null

src/renderer/components/ProjectCreation/CloudConfigView.tsx
```diff
@@ -0,0 +1,168 @@
...
+  );
+
+  const cancelButton = (
+    <SecondaryButton onClick={prevView} fullWidth>
+      Back
+    </SecondaryButton>
+  );
+
+  const handleHelpClick: () => void = () => {
+    openExternalLink('https://mlvet.app');
```
chloebrett: Can we make this be declared somewhere else so we have all marketing / landing page related stuff in one place in case we change it (e.g. help page might go to mlvet.app/help in future)

src/renderer/components/ProjectCreation/CloudConfigView.tsx
```diff
@@ -0,0 +1,168 @@
...
+  };
+
+  const defaultText =
+    "This is your first time using the cloud transcription method. To get started you'll need to provide an API key and client secret for Google Cloud Speech-to-Text";
+
+  const text = textToDisplay ?? defaultText;
+
+  useKeypress(
```
chloebrett: YAYAYYAY I love this

src/renderer/components/ProjectCreation/CloudConfigView.tsx
```diff
@@ -0,0 +1,168 @@
...
+        <IconButton
+          sx={{ color: colors.yellow[500], fontSize: 20 }}
+          onClick={closeModal}
+        >
+          <CloseIcon />
+        </IconButton>
+      </CustomRowStack>
+      <CustomColumnStack
+        justifyContent="space-between"
+        sx={{ height: '90.5%' }}
```
chloebrett: Bit of a magic number / hacky - why so specific?

chloebrett (changes requested):

rileykeane:
UI is off to a good start but still needs some work. Here are the things that need fixing up. Just small CSS styling issues. 
- Margins/padding needs to be fixed. Everything is just a little too close together. The Figma designs should have all the correct values. More info on how to view them [here](https://help.figma.com/hc/en-us/articles/360039956974-Measure-distances-between-layers)
- When hovering over the link, it should have a pointer, not a text cursor
- The link text should be slightly smaller; see Figma
- This one is on me, but I'd like to change the copy slightly: 
    "_This is your first time using cloud transcription. To get started, you’ll need to provide an API key and client secret for 
      Google Cloud Speech-to-Text_"
- Change the link to an article from Google Cloud Platform on how to get a key (assuming GCP is the solution we will be launching the beta with) - I doubt we will have the time to write our own.

src/renderer/components/ProjectCreation/CloudConfigView.tsx
MutazAshhab: ME TOOO, how nice is it!

src/renderer/components/ProjectCreation/ModalContainer.tsx
MutazAshhab: I tried 

```typescript
interface Props {
  prevView: () => void;
  closeModal: () => void;
  nextView: () => void;
  projectName: string;
  textToDisplay?: string | null;
}
```

But it errored with
```
propType "textToDisplay" is not required, but has no corresponding defaultProps declaration
```

[I didnt quite understand what the documentation for this error was trying to say](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/require-default-props.md)/I couldnt find a way around it

src/renderer/components/ProjectCreation/CloudConfigView.tsx
MutazAshhab: For now, im going to go with Rileys suggestion and just point [this tutorial](http://nerdvittles.com/creating-an-api-key-for-google-speech-recognition/).

src/renderer/components/ProjectCreation/CloudConfigView.tsx
MutazAshhab: Yeah it is hacky, I couldnt get it right

MutazAshhab: Im going to re-visit this when I re-adjust the padding, etc. that Riley mentioned

src/renderer/components/ProjectCreation/CloudConfigView.tsx
MutazAshhab: Fixed here: [1ae157b](https://github.com/chloebrett/mlvet/pull/214/commits/1ae157b9d7d63093906cd6042cd4d3f382817931)

src/renderer/components/ProjectCreation/CloudConfigView.tsx
MutazAshhab: Nice catch, I fixed it here: [57dc015](https://github.com/chloebrett/mlvet/pull/214/commits/57dc0156edd66b6fa1d27f5349dcf4314d0dc1d1)

src/renderer/components/ProjectCreation/CloudConfigView.tsx
MutazAshhab: Done here: [0fbdc31](https://github.com/chloebrett/mlvet/pull/214/commits/0fbdc31db562590379be0dcb71febd9a5527f979)

MutazAshhab:
By the way, for some reason, when you enter the project name and you navigate to the `CloudConfigView`, it focuses on the client secret instead of the API key... or instead of not focusing on anything?

MutazAshhab:
Now whats left to fix are the:
1. Margins and padding and seperating everything from each other
2. https://github.com/chloebrett/mlvet/pull/214#discussion_r954827749

MutazAshhab:
@rileykeane Here is what it looks like after i've made the fixes you asked for.

image

MutazAshhab:
Update on the UI: 

Since we are going with Assembly AI (#221) we will only need input for an api key. 

It looks something like this now (Text is yet to be updated)


image

MutazAshhab:
`storeCloudCredentials` in [this commit](https://github.com/chloebrett/mlvet/pull/214/commits/57cdb376f7d78ed34ef1d6de971e3f65a8211972) purposely removes the file and re-creates it simply because we dont need what ever was stored in there before

MutazAshhab:
UI to match the [latest figma desgin](https://www.figma.com/file/vPh49OC7hJpMD7cQMAnBUo/MLVET-UI-Design?node-id=1125%3A2863) (frame 10)
image

src/renderer/components/ProjectCreation/ModalContainer.tsx
```diff
@@ -57,11 +61,28 @@ const ModalContainer = ({ isOpen, closeModal }: Props) => {
...
+        if (isConfigRequired) {
+          viewComponents.splice(1, 0, CloudConfigView);
+        }
+        // for linter
+        return isConfigRequired;
+      })
+      .catch((err) => {
+        console.log(err);
+      });
+  }, [viewComponents]);
```
MutazAshhab: @chloebrett Could you go over this?

It adds the `CloudConfigView` to the project creation modals when necessary. But when you create a project then go back to the home page (after creation and saving)... the `CloudConfigView` gets re-inserted to the array (or most likely is just not re-rendering thus the array is unchanged).

If you close the app completely and re-open it (with the `cloudConfig.txt` in `~/.config/Electron/mlvet/`) it wont add the `CloudConfigView` to the array...

How can we make it such that if the user sets an API key then creates a project then heads back to the home page and creates a second project it doesnt go through the cloud view again

src/renderer/components/ProjectCreation/ModalContainer.tsx
```diff
@@ -57,11 +61,28 @@ const ModalContainer = ({ isOpen, closeModal }: Props) => {
...
+    return [NewProjectView, ImportMediaView, RunTranscriptionView];
+  }, []);
+
+  useEffect(() => {
+    const fetchIfCloudConfigRequired = async () => {
+      const result = await requireCloudConfig();
+      return result;
+    };
+
+    fetchIfCloudConfigRequired()
```
chloebrett: Can't you just replace this function with `requireCloudConfig().then`

src/renderer/components/ProjectCreation/ModalContainer.tsx
```diff
@@ -57,11 +61,28 @@ const ModalContainer = ({ isOpen, closeModal }: Props) => {
...
+      .then((isConfigRequired) => {
+        if (isConfigRequired) {
+          viewComponents.splice(1, 0, CloudConfigView);
```
chloebrett: You're mutating a variable directly, I'm going to assume that's the problem

src/renderer/components/ProjectCreation/ModalContainer.tsx
```diff
@@ -57,11 +61,28 @@ const ModalContainer = ({ isOpen, closeModal }: Props) => {
...
+        }
+        // for linter
+        return isConfigRequired;
+      })
+      .catch((err) => {
```
chloebrett: `.catch(console.log)`

src/renderer/components/ProjectCreation/ModalContainer.tsx
chloebrett: I would do this

1. Make a state variable `isCloudConfigRequired`
2. In your useEffect, set that variable to true or false when the component loads based on the output of the API call
3. In the `useMemo` that builds the viewComponents, make it add the `CloudCOnfigView` if `isCloudConfigRequired` is true, and otherwise it just returns the other three

Just don't mutate variables directly, react won't know to re-render if you do

src/renderer/components/ProjectCreation/ModalContainer.tsx
chloebrett: The suggestion is to give it a default value, the page you linked described how. If you can't get your head around it don't worry, null is fine

src/renderer/components/ProjectCreation/NewProjectView.tsx
```diff
@@ -108,7 +107,7 @@ const NewProjectView = ({
           <TextField
             label="Project Name"
             value={projectName}
-            onChange={(event) => handleProjectNameInput(event)}
+            onChange={(event) => handleProjectNameInput(event.target.value)}
```
chloebrett, line 110: 👍🏻

chloebrett (changes requested):

rileykeane (changes requested):
Just need to update the copy and link to use AssemblyAI and this will be good to go. 

COPY:
_This is your first time using the cloud transcription method. To get started you’ll need to provide an API key and client secret for AssemblyAI_

LINK:
https://app.assemblyai.com/signup?_ga=2.64947567.1548607132.1661819143-2080070454.1661819143

src/renderer/components/ProjectCreation/ModalContainer.tsx
```diff
@@ -57,11 +62,26 @@ const ModalContainer = ({ isOpen, closeModal }: Props) => {
...
+        CloudConfigView,
+        ImportMediaView,
+        RunTranscriptionView,
+      ];
+    }
+    return [NewProjectView, ImportMediaView, RunTranscriptionView];
+  }, [isCloudConfigRequired]);
+
+  useEffect(() => {
+    const fetchIfCloudConfigRequired = async () => {
```
chloebrett, line 78: 👍🏻

src/renderer/components/ProjectCreation/ModalContainer.tsx
```diff
@@ -57,11 +62,26 @@ const ModalContainer = ({ isOpen, closeModal }: Props) => {
     }
   };
 
-  const viewComponents = [
-    NewProjectView,
-    ImportMediaView,
-    RunTranscriptionView,
-  ];
+  const viewComponents: any = useMemo(() => {
```
chloebrett: Why does this have to be `any`? There should be a type for this

chloebrett (commented):
Code looks fine now, I'll let someone else test locally

src/renderer/components/ProjectCreation/ModalContainer.tsx
MutazAshhab: Ahh, it was an issue when I had the previous approach implemented. I fixed it here: https://github.com/chloebrett/mlvet/pull/214/commits/1808dd506eb79470d46146ef473f260f6509e7d5

chloebrett (approved):

ghost (approved):
Tested on windows. Works for me. Nice job

rileykeane (approved):
Looks good!

PR #221, opened 2022-08-26
Implement assembly AI transcription

chloebrett:
So my API key is actually included since it didn't ask for a credit card or anything, obviously we extract into a config file later. Feel free to use the API key for small testing of this PR

Excited!! 🎉


src/main/handlers/helpers/transcriptionEngines/assemblyAiTranscribeFunction.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  const transcription = await pollTranscriptionResult(transcriptionId);
+
+  console.log(
+    `AssemblyAI transcription completed, ${transcription.words.length} words received`
+  );
+
+  return transcriptionAdaptor(transcription.words || []);
+};
+
```
MutazAshhab, line 146: I know this isnt entirely what this feature is about, but could you add functionality that handles the case when the API key is invalid?

It will help down the line!!!

MutazAshhab (approved):

src/main/handlers/helpers/transcriptionEngines/assemblyAiTranscribeFunction.ts
chloebrett: Done - it now throws a custom `ApiTokenError` if the error is specifically because of an invalid API token. You can wrap the transcription call in a try/catch and `instanceof` the error to check if the thrown error was an `ApiTokenError`, then handle appropriately from there

MutazAshhab: Yes, that will do. @rileykeane this is how we will check if the user has gave us a valid API key. Since you asked for this in class.

jhea0008 (approved):

PR #223, opened 2022-08-26
Show underlines on potentially incorrect words

chloebrett:
Since assembly AI gives us a confidence per word, we can use that to mark low confidence words for the user, indicating they might need correction.

Features:
- Show low-confidence words with a yellow underline
- Show very low confidence words with a red underline
- This functionality can be turned on/off from the edit menu (Ctrl+U)
- Correcting words (even if you leave them unchanged) removes the underline

Screenshot:

Screen Shot 2022-08-26 at 8 54 52 pm


Taz17:
@chloebrett Are you intending to merge this into `feature/assembly-ai-backend`?

chloebrett:
Nope just depends on that PR, will merge this after that one

LiamTodd (approved):

src/main/ipc.ts
```diff
@@ -102,6 +103,16 @@ const initialiseIpcHandlers: (ipcContext: IpcContext) => void = (
...
+  ipcMain.handle(
+    'set-confidence-lines-enabled',
+    async (_event, menuItemEnabled, confidenceLinesShown) =>
+      setConfidenceLinesEnabled(
+        ipcContext,
+        menuItemEnabled,
+        confidenceLinesShown
+      )
+  );
+
```
MutazAshhab: `setConfidenceLinesEnabled` takes 2 arguments but here its given 3. Looks like you just need to rerun `yarn gen`

MutazAshhab (approved):

PR #224, opened 2022-08-26
Feature/menu update transcription api key

manikag-12:
## Summary
Added a menu bar option - "Update Transcription API Key". Brings up the Cloud Config View pop-up to allow the user to update the cloud transcription API Key

<hr/>

To allow the user to update the cloud transcription API key from any page

<hr/>

### What did you change?
Added an IPC receiver to update transcription API key
Wrapped the CloudConfigView with a dialog to enable open and close from the menu option

<hr/>

Yes, @Taz17's [Cloud Config View PR](https://github.com/chloebrett/mlvet/pull/214)
However, merging into it instead of develop so should be good
<hr/>

image

image

<hr/>

### OS

- [ ] Linux
- [ ] MacOS
- [x] Windows

### DIFF
--- a/src/main/menu.ts
+++ b/src/main/menu.ts
@@ -222,6 +222,14 @@ export default class MenuBuilder {
           this.mainWindow.webContents.send('initiate-export-project');
         },
       },
+      {
+        id: 'updateTranscriptionAPIKey',
+        label: 'Update Transcription API Key',
+        accelerator: 'CommandOrControl+U',
+        click: () => {
+          this.mainWindow.webContents.send('open-update-transcription-API-key');
+        },
+      },
     ];
   }
 
--- a/src/main/menu.ts
+++ b/src/main/menu.ts
@@ -227,7 +227,7 @@ export default class MenuBuilder {
         label: 'Update Transcription API Key',
         accelerator: 'CommandOrControl+K',
         click: () => {
-          this.mainWindow.webContents.send('open-update-transcription-API-key');
+          this.mainWindow.webContents.send('open-update-transcription-api-key');
         },
       },
     ];
--- a/src/main/menu.ts
+++ b/src/main/menu.ts
@@ -225,7 +225,7 @@ export default class MenuBuilder {
       {
         id: 'updateTranscriptionAPIKey',
         label: 'Update Transcription API Key',
-        accelerator: 'CommandOrControl+U',
+        accelerator: 'CommandOrControl+K',
         click: () => {
           this.mainWindow.webContents.send('open-update-transcription-API-key');
         },
--- a/src/renderer/App.tsx
+++ b/src/renderer/App.tsx
@@ -1,9 +1,10 @@
-import { Box, CssBaseline, styled, ThemeProvider } from '@mui/material';
+import { Box, CssBaseline, Modal, styled, ThemeProvider } from '@mui/material';
 import { ReactElement, useContext } from 'react';
 import { Provider, useDispatch, useSelector } from 'react-redux';
 import './App.css';
 import colors from './colors';
 import KeyboardShortcutsDialog from './components/KeyboardShortcutsDialog';
+import CloudConfigView from './components/ProjectCreation/CloudConfigView';
 import HomePage from './pages/Home';
 import ProjectPage from './pages/Project';
 import { ContainerRefContext, ContextStore } from './RootContainerContext';
@@ -11,6 +12,7 @@ import { ApplicationPage } from './store/currentPage/helpers';
 import { ApplicationStore } from './store/sharedHelpers';
 import { toggleShortcuts } from './store/shortcuts/actions';
 import applicationStore from './store/store';
+import { toggleUpdateTranscriptionAPIKey } from './store/updateTranscriptionAPIKey/actions';
 import StoreChangeObserver from './StoreChangeObserver';
 import theme from './theme';
 
@@ -18,6 +20,19 @@ const RootContainer = styled(Box)({
   margin: 0,
   background: colors.grey[900],
   height: '100vh',
+});
+
+const CustomModal = styled(Modal)({
+  display: 'flex',
+  alignItems: 'center',
+  justifyContent: 'center',
+});
+
+const CustomModalInner = styled(Box)({
+  background: colors.grey[700],
+  padding: '15px 30px 30px 30px',
+  borderRadius: '5px',
+  boxShadow: '0px 4px 4px rgba(0, 0, 0, 0.25)',
 });
 
 function Router() {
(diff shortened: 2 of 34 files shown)


src/main/menu.ts
```diff
@@ -212,6 +212,14 @@ export default class MenuBuilder {
           this.mainWindow.webContents.send('initiate-export-project');
         },
       },
+      {
+        id: 'updateTranscriptionAPIKey',
+        label: 'Update Transcription API Key',
+        accelerator: 'CommandOrControl+U',
```
chloebrett: I know it’s still a WIP but this shortcut would conflict with https://github.com/chloebrett/mlvet/pull/223

maybe Ctrl+K for “key”?

src/renderer/navigation/openUpdateTranscriptionAPIKey.ts
```diff
@@ -0,0 +1,10 @@
+import store from 'renderer/store/store';
+import { toggleUpdateTranscriptionAPIKey } from 'renderer/store/updateTranscriptionAPIKey/actions';
+
+const { dispatch } = store;
+
+const openUpdateTranscriptionAPIKey: () => any = async () => {
```
chloebrett: void not any

src/renderer/pages/Home.tsx
```diff
@@ -1,17 +1,60 @@
...
+const CustomModalInner = styled(Box)({
+  background: colors.grey[700],
+  padding: '15px 30px 30px 30px',
+  borderRadius: '5px',
+  boxShadow: '0px 4px 4px rgba(0, 0, 0, 0.25)',
+});
 
 const HomePage = () => {
+  const dispatch = useDispatch();
+  const dummyPrevView: () => void = () => {};
```
chloebrett: Can we make these inline and null instead - bit of a code smell

src/renderer/pages/Project.tsx
```diff
@@ -59,6 +73,14 @@ const ProjectPage = () => {
             videoResizeOptions
           ) => (
             <>
+              <CloudConfigView
```
chloebrett: Rather than repeating yourself for each page can you put this in AppContents instead

Faye did a similar thing for the keyboard shortcuts

chloebrett (commented):
I know this is a WIP but I added some comments, looking good so far

src/renderer/store/sharedHelpers.ts
```diff
@@ -15,6 +15,7 @@ export interface ApplicationStore {
   clipboard: Word[];
   // Array of numbers corresponding to indices of words within the transcription
   selection: number[];
+  updateTranscriptionAPIKeyOpened: boolean;
```
itsMoxMox: Not absolutely necessary, but having "update" as the first word could be misread.

Maybe do something like 
`isUpdateTranscriptAPIKeyOpened`

src/renderer/store/updateTranscriptionAPIKey/reducer.ts
```diff
@@ -0,0 +1,20 @@
...
+
+const updateTranscriptionAPIKeyReducer: Reducer<
+  ApplicationStore['updateTranscriptionAPIKeyOpened'],
+  Action<boolean>
+> = (
+  updateTranscriptionAPIKeyOpened = initialStore.updateTranscriptionAPIKeyOpened,
+  action
+) => {
+  if (action.type === UPDATE_TRANSCRIPTION_API_KEY_TOGGLED) {
+    return action.payload as boolean;
```
itsMoxMox, line 14: More of a styling thing, but I've seen a lot of casting done as

`<boolean>action.payload`

elsewhere in the project. I don't think it's worth a change but worth mentioning, as maybe we can standardise it or something idk

itsMoxMox (approved):
I agree with Chloe with having `Ctrl + K` or something similar for the shortcut, but  looks good otherwise

src/renderer/store/updateTranscriptionAPIKey/reducer.ts
chloebrett: `as` is the more common one in our project IMO, this is fine

src/main/menu.ts
```diff
@@ -212,6 +212,14 @@ export default class MenuBuilder {
           this.mainWindow.webContents.send('initiate-export-project');
         },
       },
+      {
+        id: 'updateTranscriptionAPIKey',
+        label: 'Update Transcription API Key',
+        accelerator: 'CommandOrControl+K',
+        click: () => {
+          this.mainWindow.webContents.send('open-update-transcription-API-key');
```
chloebrett: lowercase for `api`

src/renderer/components/ProjectCreation/CloudConfigView.tsx
```diff
@@ -15,11 +15,10 @@ import { PrimaryButton, SecondaryButton } from '../Blocks/Buttons';
 import ipc from '../../ipc';
 
 const { openExternalLink, storeCloudCredentials } = ipc;
-
 interface Props {
-  prevView: () => void;
+  prevView: (() => void) | null;
```
chloebrett, line 19: Does this break the other callers of `prevView` / `nextView`

manikag-12: nah I tested it and works fine. There is a check for null before these are called

chloebrett: Ok great then

chloebrett (approved):

src/renderer/App.tsx
```diff
@@ -55,6 +77,20 @@ function AppContents() {
...
+          open={hasOpenedUpdateTranscriptionAPIKey}
+          onClose={closeUpdateTranscriptionAPIKey}
+        >
+          <CustomModalInner sx={{ width: { xs: 300, sm: 400, lg: 500 } }}>
+            <CloudConfigView
+              prevView={null}
+              closeModal={closeUpdateTranscriptionAPIKey}
+              nextView={null}
+              projectName=""
+              textToDisplay={null}
```
rileykeane, line 90: The text displayed in the modal should only be the default text if the user has never saved a key before (i.e. the first time using cloud). If they already have provided a key, the modal should look like this:

Screen Shot 2022-08-30 at 10 38 29 AM

rileykeane (changes requested):
Since the key is not persisting, I cannot test what the modal looks like when the user updates the key (and is not entering one for the first time). I couldn't find it in the code either, so I think it may have been missed. 

I left a comment on how the modal should look if the user has already provided a key. I believe Mutaz designed the component to account for the two states.

src/renderer/App.tsx
exquisitebin: I'm currently working on this, should be up later today

rileykeane (approved):

PR #227, opened 2022-08-27
Take manipulation fundamentals

chloebrett:
Depends on https://github.com/chloebrett/mlvet/pull/231 so please help me merge that first

Big one - not going to explain it here, basically just the fundamentals for take manipulation. Can be turned off by passing `shouldInjectTakeInfo = false` into `preProcessTranscript` and then the app behaves like it did before.

Includes basic UI (workable but far from good), redux, and modifying the underlying structures to work (e.g. video preview and edl export are fully supported).

Take detection logic is separate and that will be injected in when it's ready, for now it just makes some hard coded out of whatever words happen to be in certain ranges.

Reasonably well tested as far as major-features-implemented-six-days-before-the-demo go.

Credits to Hugh and Thien for initial implementation which I cleaned up and added to

Will do in next PR:
- matching UI to figma
- lots of other stuff in order to make that happen https://docs.google.com/document/d/1yK2w39clPdqt9wUtST_BBTTH4RA0k8GwmaMgFz-c-34/edit


tngu0041 (commented):
TranscriptionBlock.css looks like its currently unused. That CSS is using pseudo-elements to draw the horizontal lines for the square brackets '[' on the Takes.

Import this CSS file into the TakeComponent.tsx (and probably rename it appropriately) if you want to use it. At the moment, it doesn't detect whether a Take is at the start/middle/end of a TakeGroup so it draws the ::before and ::after on all cases regardless. But you can probably dynamically detect and assign the 'classNames=<>' of the div on TakeComponent to fix this so it doesn't double-draw the line in between successive Takes (i.e. making the line twice as thick)

Alternatively, if you can think of a better way to do this, then feel free to remove this CSS file altogether.

tngu0041 (approved):

PR #232, opened 2022-08-27
Improve take manipulation UI

chloebrett:
Depends on #227

Editing actions still not perfect but this should be pretty usable for the demo. Looks much better than before

Demo recording

https://user-images.githubusercontent.com/6735055/187029363-8457fb8e-6f95-4e0d-bc63-2d1d64ad3dcf.mov

Forgot to demo the video preview with this but it works more or less as expected


itsMoxMox (approved):
Looks good! Didn't find any issues in my testing either

tngu0041 (approved):

PR #233, opened 2022-08-28
Feature/take detection injectable kate

kateahyde:
## Summary

Implemented logic for take detection to extract sentences from a list of words and then group into take groups based on similarity (similarity not implemented in this PR)

<hr/>

### What did you change?

Added takeDetection (takeDetection.ts and sentenceSimilarity.ts for the similarity code to go into) as well as tests.

<hr/>

Similarity implementation required for it to actually work (Max or Zhaohui's upcoming PR)

<hr/>

### OS

- [ ] Linux
- [ ] MacOS
- [x] Windows

### DIFF
--- a/src/main/editDelete/injectTakeInfo.ts
+++ b/src/main/editDelete/injectTakeInfo.ts
@@ -1,4 +1,4 @@
-import { IndexRange, Word } from 'sharedTypes';
+import { IndexRange, TakeGroup, Word } from 'sharedTypes';
 import { mapInRanges } from '../../sharedUtils';
 
 export interface InjectableTake {
@@ -25,15 +25,17 @@ const mockTakeGroups: InjectableTakeGroup[] = [
   },
 ];
 
-const injectMockTakeInfo = (words: Word[]) => {
+const injectTakeInfo: (
+  words: Word[],
+  takeGroups?: InjectableTakeGroup[]
+) => { words: Word[]; takeGroups: TakeGroup[] } = (
+  words,
+  takeGroups = mockTakeGroups
+) => {
   let wordsCopy = [...words];
 
-  const takeGroups = mockTakeGroups.map((_, takeGroupIndex) => ({
-    id: takeGroupIndex,
-    activeTakeIndex: 0,
-  }));
-
-  mockTakeGroups.forEach((takeGroup, takeGroupIndex) => {
+  // Inject take info into words
+  takeGroups.forEach((takeGroup, takeGroupIndex) => {
     takeGroup.takes.forEach((take, takeIndex) => {
       const { wordRange } = take;
 
@@ -48,10 +50,18 @@ const injectMockTakeInfo = (words: Word[]) => {
     });
   });
 
+  // Convert a list of InjectableTakeGroup into a list of regular TakeGroup
+  const strippedDownTakeGroups: TakeGroup[] = takeGroups.map(
+    (_, takeGroupIndex) => ({
+      id: takeGroupIndex,
+      activeTakeIndex: 0,
+    })
+  );
+
   return {
     words: wordsCopy,
-    takeGroups,
+    takeGroups: strippedDownTakeGroups,
   };
 };
 
-export default injectMockTakeInfo;
+export default injectTakeInfo;
--- a/src/main/takeDetection/__tests__/takeDetection.test.ts
+++ b/src/main/takeDetection/__tests__/takeDetection.test.ts
@@ -23,7 +23,7 @@ function mockGetSentenceSimilarity(
   if (sentenceOne === 'assessment.' && sentenceTwo === 'exam') {
     return 0.8;
   }
-  throw AssertionError;
+  throw new AssertionError();
 }
 
 jest.mock('../sentenceSimilarity');
--- a/src/main/takeDetection/__tests__/takeDetection.test.ts
+++ b/src/main/takeDetection/__tests__/takeDetection.test.ts
@@ -1,0 +1,331 @@
+import { AssertionError } from 'assert';
+import { Word } from 'sharedTypes';
+import { InjectableTakeGroup } from 'main/editDelete/mockTakeInfo';
+import { findSentences, Sentence, findTakes } from '../takeDetection';
+import getSimilarityScore from '../sentenceSimilarity';
+
+function mockGetSentenceSimilarity(
+  sentenceOne: string,
+  sentenceTwo: string
+): number {
+  if (sentenceOne === 'Hello world!' && sentenceTwo === 'Hi there.') {
+    return 0.8;
+  }
+  if (sentenceOne === 'Hi there.' && sentenceTwo === 'All ok?') {
+    return 0.2;
+  }
+  if (sentenceOne === 'All ok?' && sentenceTwo === 'test.') {
+    return 0.4;
+  }
+  if (sentenceOne === 'test.' && sentenceTwo === 'assessment.') {
+    return 0.8;
+  }
+  if (sentenceOne === 'assessment.' && sentenceTwo === 'exam') {
+    return 0.8;
+  }
+  throw AssertionError;
+}
+
+jest.mock('../sentenceSimilarity');
+const mockedSentenceSim = getSimilarityScore as jest.Mocked<
+  typeof getSimilarityScore
+>;
+
+describe('findSentences should return a list of Sentence objects based on punctuation and findTakes should correctly identify different takes', () => {
+  it('should produce expected sentences when given a list of words', () => {
+    const wordList: Word[] = [
+      {
+        word: 'Hello',
+        duration: 1,
+        startTime: 0,
+        outputStartTime: 0,
+        originalIndex: 0,
+        pasteKey: 0,
+        bufferDurationBefore: 0,
+        bufferDurationAfter: 0,
+        deleted: false,
+        confidence: 1,
+        takeInfo: null,
+      },
+      {
+        word: 'world!',
+        duration: 1,
+        startTime: 0,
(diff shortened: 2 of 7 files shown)


itsMoxMox:
Code looks good, expect ZH's algo approach to have a PR soon

src/main/takeDetection/takeDetection.ts
```diff
@@ -0,0 +1,104 @@
+import { Word, IndexRange } from 'sharedTypes';
+import {
+  InjectableTake,
+  InjectableTakeGroup,
+} from 'main/editDelete/injectTakeInfo';
```
zwan0202, line 5: The import path seems wrong, it should be `../../sharedTypes` and `../../main/editDelete/injectTakeInfo`

src/main/editDelete/injectTakeInfo.ts
```diff
@@ -1,11 +1,11 @@
-import { IndexRange, Word } from 'sharedTypes';
+import { IndexRange, TakeGroup, Word } from 'sharedTypes';
```
zwan0202, line 1: Same issue here

chloebrett: Type imports are fine to do as absolute I think

src/main/takeDetection/__tests__/takeDetection.test.ts
```diff
@@ -0,0 +1,331 @@
...
+  if (sentenceOne === 'All ok?' && sentenceTwo === 'test.') {
+    return 0.4;
+  }
+  if (sentenceOne === 'test.' && sentenceTwo === 'assessment.') {
+    return 0.8;
+  }
+  if (sentenceOne === 'assessment.' && sentenceTwo === 'exam') {
+    return 0.8;
+  }
+  throw AssertionError;
```
chloebrett: 90% sure you have to instantiate the error to properly throw it - otherwise you're throwing a class

src/main/takeDetection/__tests__/takeDetection.test.ts
```diff
@@ -0,0 +1,331 @@
...
+
+jest.mock('../sentenceSimilarity');
+const mockedSentenceSim = getSimilarityScore as jest.Mocked<
+  typeof getSimilarityScore
+>;
+
+describe('findSentences should return a list of Sentence objects based on punctuation and findTakes should correctly identify different takes', () => {
+  it('should produce expected sentences when given a list of words', () => {
+    const wordList: Word[] = [
+      {
```
chloebrett, line 37: There's a `makeBasicWord` util for this where you can just override the things you like and the others will be sane defaults. E.g. you're not modifying the startTime or anything so it's redundant to keep defining it

src/main/takeDetection/sentenceSimilarity.ts
```diff
@@ -0,0 +1,44 @@
+/**
+ * https://github.com/stephenjjbrown/string-similarity-js
```
chloebrett, line 2: Have we modified this or is it just a one for one copy of the linked code? If the latter then why not just import as a library

src/main/takeDetection/takeDetection.ts
```diff
@@ -0,0 +1,104 @@
...
+};
+
+export function findSentences(words: Word[]): Sentence[] {
+  let currentSentence: Sentence = {
+    startIndex: 0,
+    endIndex: null,
+    sentenceString: '',
+  };
+  const sentences: Sentence[] = [];
+  words.forEach((word, idx) => {
```
chloebrett, line 22: Note - any refactor type stuff I'm going to say don't worry about for now, just get this merged - but I will suggest changes anyway and we can potentially return to them to clean this up after the demo is done - depends on if we deem it worthwhile or not.

Refactor suggestion - this loop would be much cleaner as a `reduce` - see `transcriptToCuts` for a very similar example. It can be tricky to get your head around if you haven't done functional / immutable programming before though, so probably leave it for now

src/main/takeDetection/takeDetection.ts
```diff
@@ -0,0 +1,104 @@
...
+      currentSentence.sentenceString = word.word;
+    } else {
+      currentSentence.sentenceString = currentSentence.sentenceString.concat(
+        ' ',
+        word.word.toString()
+      );
+    }
+
+    if (
+      word.word.includes('.') ||
```
chloebrett, line 33: Getting a little bit "clever" (in a bad way) but `['.', '?', '!'].some(word.word.includes)` is a bit neater

src/main/takeDetection/takeDetection.ts
```diff
@@ -0,0 +1,104 @@
...
+function newTakeGroup(
+  sentenceOne: Sentence,
+  sentenceTwo: Sentence
+): InjectableTakeGroup {
+  const newGroup: InjectableTakeGroup = { takes: [] };
+  let nextTakeRange: IndexRange = {
+    startIndex: sentenceOne.startIndex as number,
+    endIndex: sentenceOne.endIndex as number,
+  };
+  let nextTake: InjectableTake = { wordRange: nextTakeRange };
```
chloebrett, line 60: Another refactor to leave for now - `nextTake` should probably replaced with two separate consts, instead of being overwritten, since it's being reused for two different purposes. Even better would be `[sentenceOne, sentenceTwo].map(...)` to produce the takes and then make the take group out of the result

src/main/takeDetection/takeDetection.ts
```diff
@@ -0,0 +1,104 @@
...
+  nextTake = { wordRange: nextTakeRange };
+  newGroup.takes.push(nextTake);
+  return newGroup;
+}
+
+export function findTakes(words: Word[]): InjectableTakeGroup[] {
+  const takeGroups: InjectableTakeGroup[] = [];
+  const sentences: Sentence[] = findSentences(words);
+  for (let i = 0; i < sentences.length - 1; i += 1) {
+    const similarity: number = getSimilarityScore(
```
chloebrett, line 75: You don't need to put the `number` type here since it's implied by the return value of `getSimilarityScore`.

src/main/takeDetection/takeDetection.ts
```diff
@@ -0,0 +1,104 @@
...
+      sentences[i].sentenceString,
+      sentences[i + 1].sentenceString
+    );
+    const threshold = 0.65;
```
chloebrett: This should probably be declared in a consts file

src/main/takeDetection/takeDetection.ts
```diff
@@ -0,0 +1,104 @@
...
+    if (similarity > threshold) {
```
chloebrett: Since all the remaining logic only occurs if this condition is true, you can flip the condition and early exit to reduce nesting instead

src/main/takeDetection/takeDetection.ts
```diff
@@ -0,0 +1,104 @@
...
+      if (takeGroups.length !== 0) {
```
chloebrett, line 80: Same here... sort of. Flip the condition, then add a return to what is currently the `else`, that way the rest of the logic is less nested once again

src/main/takeDetection/takeDetection.ts
```diff
@@ -0,0 +1,104 @@
...
+        const lastTake: IndexRange = takeGroups.at(-1)?.takes.at(-1)
+          ?.wordRange as IndexRange;
+        if (
```
chloebrett, line 83: Same idea here

src/main/takeDetection/takeDetection.ts
```diff
@@ -0,0 +1,104 @@
...
+  newGroup.takes.push(nextTake);
+  return newGroup;
+}
+
+export function findTakes(words: Word[]): InjectableTakeGroup[] {
+  const takeGroups: InjectableTakeGroup[] = [];
+  const sentences: Sentence[] = findSentences(words);
+  for (let i = 0; i < sentences.length - 1; i += 1) {
+    const similarity: number = getSimilarityScore(
+      sentences[i].sentenceString,
```
chloebrett, line 76: You reference `sentences[i]` and `sentences[i+1]` a lot, give them meaningful names e.g. `currentSentence` and `nextSentence`. You can get `currentSentence` for free if you replace the `for i=0...` with a `sentences.forEach((currentSentence, i) => ...)`

src/main/takeDetection/takeDetection.ts
```diff
@@ -0,0 +1,104 @@
...
+        if (
+          lastTake.startIndex === sentences[i].startIndex &&
+          lastTake.endIndex === sentences[i].endIndex
+        ) {
+          const nextTakeRange: IndexRange = {
+            startIndex: sentences[i + 1].startIndex as number,
+            endIndex: sentences[i + 1].endIndex as number,
+          };
+          const nextTake: InjectableTake = { wordRange: nextTakeRange };
+          takeGroups.at(-1)?.takes.push(nextTake);
```
chloebrett, line 92: Because the end result of all branches is just pushing a take, you can extract this whole "find what take group to push" logic out into its own function. Once you've done that, you can then turn the whole thing into a "filter then map" pipeline where the filter condition is `similarity > threshold` and the map function is the function you just extracted.

chloebrett (commented):
Refactor comments don't need addressing for now but maybe address them in a follow up PR

chloebrett (approved):
I'm going to approve since it works and I want to get this in for the demo, just take note of the refactor stuff above

emmapainter (approved):

src/main/takeDetection/sentenceSimilarity.ts
kateahyde: @zwan0202

PR #235, opened 2022-08-28
Bugfix/fix ffmpeg ffprobe packaged app

ghost:
## Summary

Alright, this PR should fix the packaged app. I have checked on my windows PC and the packaged app correctly extracts the audio, transcribes and then displays the video preview.

<hr/>

Previously, when we (team mulan) thought we had fixed this issue. We did not. Now we are.

<hr/>

### What did you change?

I have edited the package.json so that at build time the ffmpeg and ffprobe binaries get copied over (adds about 20s to packaging) to the packaged app and then I have changed the code to set the binary paths to the new location.

Note: I also hardcoded the express port for the packaged app. Will figure out how to pass in environment variables to packaged app later.

<hr/>

### OS

@Taz17 tagging you for linux
@rileykeane @chloebrett @jhea0008 one of you please check on your snazzy arm macs
@emmapainter please check on your old ass intel mac

- [x] Linux
- [x] MacOS
- [x] Windows

### DIFF
--- a/src/main/ffUtils.ts
+++ b/src/main/ffUtils.ts
@@ -1,12 +1,13 @@
 import ffmpegStaticPath from 'ffmpeg-static';
 import ffprobeStatic from 'ffprobe-static';
+import path from 'path';
 
 export const ffmpegPath = ffmpegStaticPath.replace(
-  'app.asar\\dist\\main\\',
-  'ffmpeg-static\\'
+  path.join('app.asar', 'dist', 'main'),
+  path.join('ffmpeg-static')
 );
 
 export const ffprobePath = ffprobeStatic.path.replace(
-  'app.asar\\dist\\main\\',
-  'ffprobe-static\\'
+  path.join('app.asar', 'dist', 'main'),
+  path.join('ffprobe-static')
 );
--- a/src/main/ffUtils.ts
+++ b/src/main/ffUtils.ts
@@ -2,11 +2,11 @@ import ffmpegStaticPath from 'ffmpeg-static';
 import ffprobeStatic from 'ffprobe-static';
 
 export const ffmpegPath = ffmpegStaticPath.replace(
-  'app.asar',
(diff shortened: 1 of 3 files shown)


src/main/ffUtils.ts
```diff
@@ -2,11 +2,11 @@ import ffmpegStaticPath from 'ffmpeg-static';
...
-  'app.asar.unpacked'
+  'app.asar\\dist\\main\\',
+  'ffmpeg-static\\'
 );
 
 export const ffprobePath = ffprobeStatic.path.replace(
-  'app.asar',
-  'app.asar.unpacked'
+  'app.asar\\dist\\main\\',
+  'ffprobe-static\\'
```
MutazAshhab: It still spins forever on linux (and I suspect it will do the same for macos)

This is probably the cause. You are using `\\` which is only used by windows (wont work on linux/macos).

Could you try:

```javascript
import ffmpegStaticPath from 'ffmpeg-static';
import ffprobeStatic from 'ffprobe-static';
import path from 'path';

export const ffmpegPath = ffmpegStaticPath.replace(
  path.join('app.asar', 'dist', 'main'),
  path.join('ffmpeg-static')
);

export const ffprobePath = ffprobeStatic.path.replace(
  path.join('app.asar', 'dist', 'main'),
  path.join('ffprobe-static')
);
```

MutazAshhab (changes requested):

src/main/ffUtils.ts
MutazAshhab: My bad, even if you try it you wont be able to try linux/macos :laughing: 

I just tested that solution and it worked. No more infinite spinning.

MutazAshhab (approved):

ghost:
Thanks for adding those changes @Taz17 , I did not know that

chloebrett (approved):
Build is working on my machine

PR #236, opened 2022-08-28
Feature/take detection sentence sim

zwan0202:
## Summary
Implemented algorithm for sentence similarity calculation for take detection.

<hr/>

Add the sentence similarity algorithm. We decided not to use ML for now due to the packaging issue. 

<hr/>

Depends on https://github.com/chloebrett/mlvet/pull/233 and https://github.com/chloebrett/mlvet/pull/234

### DIFF
--- a/src/main/takeDetection/sentenceSimilarity.ts
+++ b/src/main/takeDetection/sentenceSimilarity.ts
@@ -2,6 +2,7 @@ export default function getSimilarityScore(
   sentenceOne: string,
   sentenceTwo: string
 ): number {
-  sentenceOne.concat(sentenceTwo);
-  return 0.5;
+  // For mock, just return 1 (100% similar) if the sentences start with the same letter, otherwise 0 (0% similar) - allows for some meaningful testing
+
+  return sentenceOne[0].toLowerCase() === sentenceTwo[0].toLowerCase() ? 1 : 0;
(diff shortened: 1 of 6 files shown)


src/main/takeDetection/sentenceSimilarity.ts
```diff
@@ -0,0 +1,48 @@
...
+ * Calculate similarity between two strings
+ * @param {string} str1 First string to match
+ * @param {string} str2 Second string to match
+ * @param {number} [substringLength=2] Optional. Length of substring to be used in calculating similarity. Default 2.
+ * @returns Number between 0 and 1, with 0 being a low match score.
+ */
+export const getSimilarityScore = (
+  str1: string,
+  str2: string,
+  substringLength = 2
```
kateahyde, line 12: What's this substringLength for/why does it need to be a param?

zwan0202: It indicates the length of the substring for each comparison, the higher the value the closer to the exact string match. I leave it here because It will be easier if we decide to use a tighter match later (guess not, I'll just move it to the inside function)

tngu0041 (approved):
Similarity detection using bigrams is very simple and surprisingly decent. Also O(n).

PR #239, opened 2022-08-30
Fix bug when returning home with word selected

### LINKED ISSUE #219: Bug: Return to home from transcription page causes crash if no changes were made to project
If an edit is made the " do you want to save" prompt comes up and all works well. However if no changes are made to the project this seems to trigger the big where the UI fails and the on refresh the Javascript error pops up as well ( refresh not shown in video)
https://user-images.githubusercontent.com/55219708/186790777-81077a31-4c96-4624-8ae5-c25e21eaf512.mp4

chloebrett:
Resolves #219


Irmy (approved):
Legendary

PR #240, opened 2022-08-30
Feature/restore text

rileykeane:
## Summary
Restore the original text of the transcription that has been removed. 

Currently, you click the edit marker to restore the text - this is not the intended behaviour. @emmapainter will be raising a PR shortly that will present a popover with the original text so the user can see it before it's restored. 

FYI - It looks like a lot of lines, but almost 70% are just test cases. 

<hr/>

Currently, if you delete a section of the text, you need to go back through the entire undo stack to bring it back. With this feature, you can always bring back original the original text. 
<hr/>

### What did you change?
- Created a function that finds the start and end index for an original deleted section of text 
- Created an op to restore a section of text that are also undo-able 
- Created a function that returns the original word objects for a deleted section of text (this will be used to display the text in the popover)
- Modified the edit marker to restore the original text when you click it. This will be changed to open the popover with the original text so the user can see it before its restored. (This will be included in a PR shortly)  
- Created test cases to test all the new functions
<hr/>

https://user-images.githubusercontent.com/27614346/187562760-984ac5d9-9096-4491-861b-1e9e20a1f12f.mov

<hr/>

### OS

- [ ] Linux
- [x] MacOS
- [ ] Windows


src/renderer/components/Editor/EditMarker.tsx
```diff
@@ -37,14 +26,19 @@ const EditMarker = ({ transcription, word, index, isSelected }: Props) => {
...
   const notPasted = word.pasteKey === 0;
 
-  const EditMarkerContainer = makeEditMarkerContainer(isSelected);
-
   return (isInOriginalPos || hasNotMoved) &&
     hasNoNeighbourMarker &&
     notPasted ? (
-    <EditMarkerContainer>
+    <Box
+      sx={{
```
chloebrett, line 33: Any reason you're inlining these styles now?

src/renderer/components/Editor/WordComponent.tsx
```diff
@@ -188,8 +188,6 @@ const WordComponent = ({
       return;
     }
 
-    console.log('click');
```
chloebrett, line 191: oops lol

src/renderer/editor/restore.ts
```diff
@@ -0,0 +1,66 @@
...
+
+      return (
+        !currWord.deleted ||
+        !nextWord.deleted ||
+        currWord.originalIndex + 1 !== nextWord.originalIndex
+      );
+    }
+  );
+
+  // If the number of words to restore is -1, then we are restoring
```
chloebrett: `If no last index is found, then...`

src/renderer/store/transcriptionWords/ops/restoreSection.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export type RestoreSectionOp = Op<
+  RestoreSectionPayload,
+  UndoRestoreSectionPayload
+>;
+
+export const makeRestoreSection: (ranges: IndexRange[]) => RestoreSectionOp = (
+  ranges
+) => ({
+  do: [sectionRestored(ranges)],
```
chloebrett: Can you make it select the restored words once the restore is complete? And likewise, when undoing a restore, just clear the selection

src/renderer/store/transcriptionWords/reducer.ts
```diff
@@ -132,6 +136,25 @@ const transcriptionWordsReducer: Reducer<Word[], Action<any>> = (
...
+      deleted: false,
+    });
+
+    return mapInRanges(words, markUndeleted, ranges);
+  }
+
+  if (action.type === UNDO_RESTORE_SECTION) {
+    const { ranges } = action.payload as UndoRestoreSectionPayload;
+
+    const markDeleted = (word: Word) => ({ ...word, deleted: true });
```
chloebrett: I feel like these are used enough they should be extracted as utils

chloebrett (commented):
Code looks super slick, testing functionality now

src/renderer/components/Editor/EditMarker.tsx
rileykeane: I could not get the onClick to work with the previous implementation. I think because the MUI component was converted to a React Component, the onClick that comes with a Box component was lost. I'd have to pass the onClick down through props, so I felt inline was the easiest and cleanest way.

chloebrett:
Tested locally with the following cases
- delete single word and restore
- delete multiple words and restore
- delete words then delete more words that wrap them and restore
- delete words then delete words next to them and restore
- delete the first word and restore
- delete the last word and restore
- delete the entire sequence and restore

All major use cases I could think of work just fine :)

There is one bug which may or may not be easy to fix given the current architecture. To replicate:
1. Delete a sequence of words e.g. three words
2. Delete another sequence that wraps the previous sequence
3. Hit undo

The entire five-word sequence will be restored. Hitting undo again does nothing

This also occurs if e.g.

A B C D E F G

1. delete C
2. delete D
3. delete E
4. delete B to F
5. undo

This restores the entire sequence B -> F

You might argue this is intended behaviour, but it means that undo isn't doing what it should - restoring the project state to what it was one action ago.

src/renderer/components/Editor/EditMarker.tsx
chloebrett: Interesting, no problem then

rileykeane:
> 
> There is one bug which may or may not be easy to fix given the current architecture. To replicate:
> 

I'll take a look, however I think you're right regarding architecture. I was discussing with @hhen0005 and think the edit marker and restore logic should be refactored. I also found some weird behaviour when copying text back into its original place and deleting it again - the edit marker doesn't seem to reappear and the text is unrecoverable. 

A redesign is not feasible right now, however something we can look into after the demo. From my testing, I believe it's stable enough for a demo and even the beta.

chloebrett:
I agree with everything you just said

src/renderer/store/transcriptionWords/ops/restoreSection.ts
rileykeane: Good idea. Done.

src/renderer/store/transcriptionWords/reducer.ts
rileykeane: Done.

rileykeane:
@chloebrett Just pushed some updates based on your feedback.

src/renderer/editor/restore.ts
rileykeane: Updated.

Irmy:
PM reviewed and accepted - the bug mentioned by Chloe isnt too far from expected behaviour as a user, however who knows what downstream unintented behaviour it may cause later. Totally agree with Riley that priorities for the demo are definitely met.

Irmy (approved):

PR #251, opened 2022-08-31
Cloud config cherry pick

chloebrett:
# Cloud Config File

## Description
Implements the ability to enter, store and retrieve API keys nad a default TranscriptionEngine. Configs are stored in a JSON file in the appdata mlvet folder as "cloudConfig.json". 

The currently stored key is displayed in the CloudConfigView text box upon opening.
The text for the CloudConfigView information changes depending on whether it is the user's first time entering the API key.

## Why is this change needed?
Allows users to enter their custom API key, use it to transcribe and have it saved for multiple sessions.

## OS

- [x] Windows
- [x] Mac
- [ ] Linux

### DIFF
--- a/src/main/config.ts
+++ b/src/main/config.ts
@@ -1,14 +1,7 @@
 /* eslint-disable import/prefer-default-export */
 
 import { TranscriptionEngine } from '../sharedTypes';
-import readDefaultEngineConfig from './handlers/file/readDefaultEngineConfig';
 
 // Which transcription engine to use for any transcription.
 // This will eventually be replaced with a config file allowing the user to choose which engine they want
 export const TRANSCRIPTION_ENGINE = TranscriptionEngine.ASSEMBLYAI;
-
-export const DEFAULT_ENGINE_CONFIG = readDefaultEngineConfig().then(
-  (engineConfig) => {
-    return engineConfig ?? '';
-  }
-);
--- a/src/main/config.ts
+++ b/src/main/config.ts
@@ -1,9 +1,14 @@
 /* eslint-disable import/prefer-default-export */
 
 import { TranscriptionEngine } from '../sharedTypes';
+import readDefaultEngineConfig from './handlers/file/readDefaultEngineConfig';
 
 // Which transcription engine to use for any transcription.
 // This will eventually be replaced with a config file allowing the user to choose which engine they want
 export const TRANSCRIPTION_ENGINE = TranscriptionEngine.ASSEMBLYAI;
 
-export const ASSEMBLYAI_API_KEY = '';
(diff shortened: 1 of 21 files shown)


exquisitebin (approved):
Awesome thanks for helping fix my branch! It seems to be all working.

manikag-12 (changes requested):
Look really good but when when there already is a key saved and I create a new project, it doesn't let me move forward without changing the transcription key.

The save button is disabled and that's the only way to move to the next view

image

MutazAshhab:
I've tested it out locally, it works and I found a couple of things.

1. If you add an api key, click save (go to the next view), then go back, it save button is then disabled. It should be enabled as the api key entered by the user is persisted.
2. The back and save button in the cloud config view and not longer on the same level as the next view. See this video


https://user-images.githubusercontent.com/86769131/187671070-0cb218a1-1ed9-421d-be07-cf837fabb8ec.mp4

exquisitebin:
> Look really good but when when there already is a key saved and I create a new project, it doesn't let me move forward without changing the transcription key.
> 
> The save button is disabled and that's the only way to move to the next view
> 
> image

New commit should fix this issue

rileykeane (changes requested):
@exquisitebin @Taz17 @chloebrett  It looks like you can't paste the key into the text field - which is a huge pain to have to type it out...

I believe it's because we have custom copy/paste actions. How difficult is this going to be to fix?

rileykeane:
Also - the screen to enter a key should not appear in the new project flow if the user has already provided a key.

exquisitebin:
> I've tested it out locally, it works and I found a couple of things.
> 
> 1. If you add an api key, click save (go to the next view), then go back, it save button is then disabled. It should be enabled as the api key entered by the user is persisted.
> 2. The back and save button in the cloud config view and not longer on the same level as the next view. See this video
> 
>  misalignment.mp4

New commits should fix both issues

exquisitebin:
> @exquisitebin @Taz17 @chloebrett It looks like you can't paste the key into the text field - which is a huge pain to have to type it out...
> 
> I believe it's because we have custom copy/paste actions. How difficult is this going to be to fix?

Chloe and I found it's a Mac problem, as it works fine on my Windows.



> Also - the screen to enter a key should not appear in the new project flow if the user has already provided a key.

The current issue is that its managed by the useEffect in the ModalContainer which only gets called once while you're at the home page. Therefore if you complete making a new project and go back to the homepage (or restart the app) it no longer appears.

rileykeane (approved):

manikag-12 (approved):

PR #263, opened 2022-09-03
Collaborative editing spike

chloebrett:
Things that aren't done yet
- media file is broken because it doesn't ask the user to find the equivalent file on their computer yet
- "join" and "host" buttons should open modals rather than just leaving free form text inputs all over the place


rileykeane:
Wow 😍

Just curious, what's the plan for video preview on remote connections? I imagine it's tricky since the video is local to the host.

chloebrett:
My thought is that when you join a session as a guest, you’re prompted to choose the same video file from your local machine, and then all collaborators have independent video previews (i.e. playback positions etc aren’t synced, just the transcription edits)

jhea0008 (approved):

jhea0008 (commented):
Probs want to have a bit more of a polished UI before we merge into dev

chloebrett:
I'll merge with the flag disabled so the UI doesn't show and then put the inputs behind buttons / a modal in a follow up

PR #265, opened 2022-09-09
Feature/multi sentence take detection

zwan0202:
## Summary

Updated single sentences take detection to multi-sentences.

<hr/>

### High-level explanation of the algorithm

```JavaScript
findTakes(words: Word[]): InjectableTakeGroup[]
  sentences: Sentence[] = findSentences(words)
  // initialise

  while (detectNotCompelet):
    // initialise

    // search for any similar sentence of current index
    while (not find similar sentence and not reach half point of remaining length from current index):
      getSimilarityScore(currentSentence, nextSentence)
      if found similar sentence:
        // record take length
        potentialTakeLen = nextSentenceIdx - currentSentenceIdx
      
      // next time skip that length and check for similarity
      // if not similar, break, detect takes similarity

    if found any potential takes:
      // start detection
      // compare sentences within the potential take length, start from the second sentence
      for i from 1 to potentialTakeLen - 1:
        // use the next sentence's index in the first take group
        // as reference to compare the rest of take groups

        // compare sentence in first potential take group with every other potential take group
        for j from 1 to len(potentialTakeStartIdxs) - 1:
          getSimilarityScore(currentSentence, nextSentence)

          if sentence in the take group is not similar
            discard all group to the right
          if only the first group has left
            update current index and start again 
          
      if at least two groups are similar:
        for takeGroup:
          new takeGroup = { 
            takeGroupStartSentence.startIndex,
            takeGroupEndSentence.endIndex
          }
          
        // update current index
        updatedCurrentSentenceIdx = potentialTakeLen * len(potentialTakeStartIdxs)
  
return takeGroups
```

The worst case occurs when there are no similar takes, it checks half of the remaining list each iteration, `O(n^2)`. Because each sentence is independent, and we can only compare two sentences each time. I didn't find a way to store previous similarities for future comparisons. 

If ML is working and can query by multiple sentences, we can optimize this using a pattern matching algorithm (maybe), and reduce the complexity to linear.

I haven't add enough test for it.

<hr/>

### OS

- [ ] Linux
- [ ] MacOS
- [ ] Windows


src/main/takeDetection/__tests__/takeDetection.test.ts
```diff
@@ -31,127 +37,45 @@ const mockedSentenceSim = getSimilarityScore as jest.Mocked<
...
-        pasteKey: 0,
-        bufferDurationBefore: 0,
-        bufferDurationAfter: 0,
-        deleted: false,
-        confidence: 1,
-        takeInfo: null,
-      },
-    ];
+    const wordList = makeWords([
+      'Hello',
```
chloebrett: You could just do this as a single string and then `.join(' ')`

src/main/takeDetection/__tests__/takeDetection.test.ts
```diff
@@ -188,130 +112,22 @@ describe('findSentences should return a list of Sentence objects based on punctu
...
+      'test.',
+      'assessment.',
+      'exam',
+    ]);
 
     (mockedSentenceSim as jest.Mock).mockImplementation(
       mockGetSentenceSimilarity
     );
-    const takes: InjectableTakeGroup[] = findTakes(wordList, 0.7);
+    const takes: InjectableTakeGroup[] = findTakes(wordList);
```
chloebrett: Keep the hard coded 0.7 here - the reason it was put here is because relying on a constant made the test break when we changed the constant - we want to test the behaviour of the findTakes function in general, without assuming the actual threshold used in the app will be a certain value

src/main/takeDetection/takeDetection.ts
```diff
@@ -48,56 +48,202 @@ export function findSentences(words: Word[]): Sentence[] {
...
-  let nextTake: InjectableTake = { wordRange: nextTakeRange };
-  newGroup.takes.push(nextTake);
-  nextTakeRange = {
-    startIndex: sentenceTwo.startIndex as number,
-    endIndex: sentenceTwo.endIndex as number,
-  };
-  nextTake = { wordRange: nextTakeRange };
-  newGroup.takes.push(nextTake);
+
+  potentialTakeStartIdxs.forEach((startIdx) => {
```
chloebrett: Since this is unconditionally pushing to an array you can replace the whole thing with a `.map`

i.e. the entire function body would become

```
return potentialTakeStartIdxs.map((startIdx) => {
  const takeStartSentence = sentences[startIdx];
  const takeEndSentence = sentences[startIdx + potentialTakeLen - 1];

  return {
    wordRange: {
      startIndex: takeStartSentence.startIndex as number,
      endIndex: takeEndSentence.endIndex as number,
    },
  } as InjectableTake;
});
```

Also I noticed take length is constant across an entire take group. What happens if two takes are similar but contain a different number of sentences - is that just not allowed? I can see it happening e.g.

take1: "Thanks for watching, everybody. Please don't forget to like and subscribe, and leave a comment below."
take2: "I appreciate every one of you watching this so please hit like and subscribe as well as leaving a comment."

Obviously a contrived example, but you get the point.

src/main/takeDetection/takeDetection.ts
```diff
@@ -48,56 +48,202 @@ export function findSentences(words: Word[]): Sentence[] {
...
-            startIndex: sentences[i + 1].startIndex as number,
-            endIndex: sentences[i + 1].endIndex as number,
-          };
-          const nextTake: InjectableTake = { wordRange: nextTakeRange };
-          takeGroups.at(-1)?.takes.push(nextTake);
-        } else {
-          takeGroups.push(newTakeGroup(sentences[i], sentences[i + 1]));
+
+  if (isSimilarTake)
+    takeGroups.push(
```
chloebrett: In general mutating the input to a function makes it hard to reason about what functions do - it's much easier if they are pure i.e. they take defined inputs and give defined outputs, without modifying the inputs they are given. Can you think of a way to apply that to this function?

(I'm aware the push method is asymptotically faster but I think in this case readability > speed as take detection should be pretty fast anyway)

src/main/takeDetection/takeDetection.ts
```diff
@@ -48,56 +48,202 @@ export function findSentences(words: Word[]): Sentence[] {
...
+
+export function findTakes(words: Word[]): InjectableTakeGroup[] {
+  const sentences: Sentence[] = findSentences(words);
+  const takeGroups: InjectableTakeGroup[] = [];
+
+  let currentSentenceIdx = 0;
+  let detectNotCompelet = true;
+  const maxSentenceIdx = sentences.length - 1;
+
+  while (detectNotCompelet) {
```
chloebrett: `complete`

src/main/takeDetection/takeDetection.ts
```diff
@@ -48,56 +48,202 @@ export function findSentences(words: Word[]): Sentence[] {
...
+      )
+        break;
+
+      const isSimilar =
+        getSimilarityScore(
+          sentences[currentSentenceIdx].sentenceString,
+          sentences[nextSentenceIdx].sentenceString
+        ) > THRESHOLD;
+
+      // if already found potentail take
```
chloebrett: `potential`

src/main/takeDetection/takeDetection.ts
```diff
@@ -48,56 +48,202 @@ export function findSentences(words: Word[]): Sentence[] {
...
+    } else {
+      currentSentenceIdx += 1;
+    }
+
+    if (
+      potentialTakeStartIdxs[potentialTakeStartIdxs.length - 1] +
+        potentialTakeLen ===
+      sentences.length
+    ) {
+      detectNotCompelet = false;
```
chloebrett: Since the only time you update this condition is at the end of the loop, it'd be cleaner as a `do ... while`

src/main/takeDetection/takeDetection.ts
```diff
@@ -48,56 +48,202 @@ export function findSentences(words: Word[]): Sentence[] {
...
+};
 
-export function findTakes(words: Word[], threshold = THRESHOLD): InjectableTakeGroup[] {
-  const takeGroups: InjectableTakeGroup[] = [];
-  const sentences: Sentence[] = findSentences(words);
-  for (let i = 0; i < sentences.length - 1; i += 1) {
-    const similarity: number = getSimilarityScore(
-      sentences[i].sentenceString,
-      sentences[i + 1].sentenceString
+const startDetection: (
```
chloebrett: Add a comment explaining what this function does, what the inputs and outputs are, any mutations etc

src/main/takeDetection/__tests__/takeDetection.test.ts
```diff
@@ -31,127 +37,45 @@ const mockedSentenceSim = getSimilarityScore as jest.Mocked<
   typeof getSimilarityScore
 >;
 
+const makeWords = (words: string[]): Word[] => {
+  const defaultWord = {
```
chloebrett: Can use `makeBasicWord` for this

chloebrett (changes requested):
Haven't had a chance to fully test or play around with the logic yet but here are some preliminary comments

chloebrett:
~~Btw is there a doc explaining the algorithm? Would be useful as it's a bit hard to follow purely based on the code~~

Edit: oops me dumb dumb didn't see description

src/main/takeDetection/takeDetection.ts
zwan0202: I think we decided similar takes should have the same length. 

Because currently, we can only compare two sentences each time. It is hard to detect one sentence that has a similar meaning to several sentences combined, unless we compute all combinations of the sentences

src/main/takeDetection/__tests__/takeDetection.test.ts
chloebrett: Did you resolve this without commenting?

src/main/takeDetection/takeDetection.ts
chloebrett: Okay, it does seem easier

You could do a multi sentence comparison using the geometric mean of all cross-similarities, though this would be O(n^2) with respect to the number of sentences compared

chloebrett (changes requested):
Code is looking cleaner, just add some tests that verify the multi sentence behaviour works as expected and it's good to go

src/main/takeDetection/__tests__/takeDetection.test.ts
zwan0202: Sorry, I resolved this together with use `makeBasicWord` one. 
Based on my understanding, `.join(' ')` returns an array as a string, but here we want to pass a list of strings to get a list of words.

chloebrett: Ah my bad meant `.split`. Up to you, I think the current approach is clean enough

MutazAshhab (approved):

ghost:
Hey, nice work. I'm Trying to better understand the algorithm and any test cases that might break it.
I'm wondering what would happen if the potential takes had a different number of sentences? I realise this is more logic and more computation time but maybe in a future PR we could compare paragraphs with `sentenceSimilarity` func rather than comparing the sentences within.
Sorry, If I have misunderstood the algo.

Also,re:Chloe, please add tests

MutazAshhab:
Hey @zwan0202, I thought more about this, I think some of the comparisons are redundant and there could be a way of decreasing the number of comparisons done.

One approach could be we first iterate over all the sentences and group all adjacent sentences of similar (or exact) lengths together. And measure lengths by number of words rather than length of the string.

We then run the string similarity algo on the groups that we have. That way we dont compare sentences of 8 words with a sentence of 20 words.

For example we have these sentences:

`[3, 6, 6, 20, 15, 20, 35, 6, 6, 7, 15]` (Here im just showing the number of words in the sentence.

Then group then like so

`[[6, 6], [6, 6, 7]]` First nested list contains elements setences[] with index 1 and 2. And the second nested list contains indexes 7 8 9

zwan0202:
> 

That's a good idea. But what if the take has multiple sentences?

We can't just group sentences with similar lengths. For example `[3, 6, 5, 7, 8, 6, 4, 7]` In this case, there's no take.

Maybe we can convert all sentences into word count and run a pattern matching algo first, and then run the string similarity algo.

zwan0202:
> Hey @zwan0202, I thought more about this, I think some of the comparisons are redundant and there could be a way of decreasing the number of comparisons done.
> 
> One approach could be we first iterate over all the sentences and group all adjacent sentences of similar (or exact) lengths together. And measure lengths by number of words rather than length of the string.
> 
> We then run the string similarity algo on the groups that we have. That way we dont compare sentences of 8 words with a sentence of 20 words.
> 
> For example we have these sentences:
> 
> `[3, 6, 6, 20, 15, 20, 35, 6, 6, 7, 15]` (Here im just showing the number of words in the sentence.
> 
> Then group then like so
> 
> `[[6, 6], [6, 6, 7]]` First nested list contains elements setences[] with index 1 and 2. And the second nested list contains indexes 7 8 9

That's a good idea. But what if the take has multiple sentences?

We can't just group sentences with similar lengths. For example `[3, 6, 5, 7, 8, 6, 4, 7]` In this case, there's no take.

Maybe we can convert all sentences into word count and run a pattern matching algo first, and then run the string similarity algo. But we need to decide what length differences can be counted as similar. I think `+-3` is enough.

MutazAshhab:
The differences can be calculated using the threshold. If threshold is 1 then you can only compare sentences of same length.

We can calculate it.

acceptable +- = `math.ceil(sentence length * (1/threshold))`

if theshold is 0.3, then the acceptable difference would be

`math.ceil(sentance length * 3.3333)`

if theshold is 1, then the acceptable difference would be

`math.ceil(sentance length * 1)`

zwan0202:
> Hey, nice work. I'm Trying to better understand the algorithm and any test cases that might break it. I'm wondering what would happen if the potential takes had a different number of sentences? I realise this is more logic and more computation time but maybe in a future PR we could compare paragraphs with `sentenceSimilarity` func rather than comparing the sentences within. Sorry, If I have misunderstood the algo.
> 
> Also,re:Chloe, please add tests

I think we decided similar takes should have the same length, due to the string similarity algo can only take two sentences. It's not like ML model. 
If we want to compare paragraphs with `sentenceSimilarity`, we have to pre-compute all sentence combinations with different lengths to form a paragraph.

ghost (approved):

chloebrett (commented):
Looks mostly fine, the one concern I have is that we're not testing that multi sentence take detection works with sentences longer than a single word, so add a test for that

zwan0202:
> Looks mostly fine, the one concern I have is that we're not testing that multi sentence take detection works with sentences longer than a single word, so add a test for that

The reason I only use a single word is because the sentence similarity algorithm only checks the occurrence of substrings, so the length of the sentence is not that important I think.
But I'll add one for it

chloebrett (approved):

itsMoxMox (approved):

PR #266, opened 2022-09-09
Support explicitly manipulating long pauses

chloebrett:
No UI design has been made yet so these are just underscores for now.

TODO

- [x] Prevent dragging pauses around
~~- [x] Prevent pauses being copied unless the text either side is being copied~~
- [x] Prevent splitting pauses
- [x] Ensure merged phrases containing pauses behave as expected
- [x] Prevent opening correction box for pauses


Irmy (changes requested):
"Prevent pauses being copied unless the text either side is being copied" - Unfortunately having this felt like a bug. If i copied a set of text "__TEXT__TEXT__"  when its pasted it should be the same rather than "TEXT__TEXT". 
Please have the long pauses be able to be moved individually as well. At the moment I can cut the silences "__", but cant paste them. The use case for this is where the creator has done a particular body language or facial expression that they may want to move or copy to multiple places in the video but would be represented as a silence in the transcription

src/main/editDelete/preProcess.ts
```diff
@@ -84,6 +85,103 @@ const calculateBuffers: (totalDuration: number) => MapCallback<Word, Word> =
...
+  bufferDurationBefore: 0,
+  bufferDurationAfter: 0,
+  deleted: false,
+  originalIndex: 0, // this gets fixed later
+  confidence: 1,
+  pasteKey: 0,
+  takeInfo: null,
+});
+
+const injectPauses: (
```
jhea0008, line 86: This decleration is very difficult to read. Could we could create types or something to better display it?

jhea0008 (commented):
Agree with Irmy that the copy/paste pauses thing feels like a bug. Other than that only one minor code issue

src/main/editDelete/preProcess.ts
chloebrett: Yeah it's a curried flatMap predicate 🍛 

Could just use the existing MapCallback type. It was originally a reduce then I changed it to a flatmap so the inputs used to be different

src/main/takeDetection/takeDetection.ts
```diff
@@ -101,4 +110,4 @@ export function findTakes(words: Word[], threshold = THRESHOLD): InjectableTakeG
   return takeGroups;
 }
 
-export default { getSimilarityScore };
+export default getSimilarityScore;
```
fwang1703: Wait why are we exporting `getSimilarityScore` from this file when we export it from `sentenceSimilarity.ts`?

src/renderer/components/Editor/WordDragManager.tsx
```diff
@@ -123,15 +123,22 @@ const WordDragManager = ({ clearSelection, children }: Props) => {
       }
 
       if (event.altKey) {
-        // If alt/option held, then start a drag-move action
+        // If alt/option held, then start a drag-move action.
+
+        // Pauses (where word text === null) cannot be dragged around
+        const isPause = (words ?? [])[wordIndex].word === null;
+        if (isPause) {
+          return;
```
fwang1703: Following on from Irmy and Justin's suggestion to be able to copy/paste pauses, should we also be able to drag pauses around if the pause is like an action or movement? But I guess that might depend on what the UI's gonna look like for pauses.

src/renderer/store/transcriptionWords/ops/correctWord.ts
```diff
@@ -21,7 +21,7 @@ export const makeCorrectWord: (
       selectionRangeAdded(rangeLengthOne(index)),
     ],
     undo: [
-      undoWordCorrected(index, prevText, prevConfidence ?? 1),
+      undoWordCorrected(index, prevText as string, prevConfidence ?? 1),
```
fwang1703, line 24: Is this just for clarity? Or was the typing not working properly?

fwang1703 (changes requested):
Most of it looks good! Just a few questions (some which aren't directly related to this PR sorry), and I agree with Irmy and Justin on the copy/paste thing + could we also potentially drag pauses?

src/renderer/store/transcriptionWords/ops/correctWord.ts
chloebrett: The action type assumes text to be a string, but with the addition of pauses text can be null. Hence the cast, as we know that if a word is being corrected, it must not have been a pause - since you can't correct pauses.

src/renderer/components/Editor/WordDragManager.tsx
chloebrett: Solid chance we'll be nuking drag as multi drag isn't supported (and would be a lot of work) and drag in general is not performance optimised

src/main/takeDetection/takeDetection.ts
chloebrett: Yeah weird dunno how that happened, removed

src/main/editDelete/preProcess.ts
chloebrett: Changed it

src/renderer/store/transcriptionWords/ops/correctWord.ts
fwang1703: Ah yep that makes sense!

fwang1703 (approved):
Looks good, thank you!

Irmy (approved):
Looks so good! Thank you!

PR #270, opened 2022-09-12
Fix/cancel project modal styling

tngu0041:
Before:
image

After:
image

Task #10 on the sheet Irmy posted: _"10. Cancel project screen should have continue highlighted + cancel project in grey"_


Irmy (approved):
Thanks so much Thien! And thanks for updating the read me too!

PR #271, opened 2022-09-12
Added handling for cancel button, so it no longer goes to HOME page

tngu0041:
Task 4 on the sheet Irmy posted: _"Cancel button on go home doesn't work"_

Clicking cancel now correctly leaves the user in the project page.
Before this change it would go to the HOME page and the user would lose all changes to their project.

I assume this doesn't need a written test for. But lmk if you disagree,


fwang1703 (approved):
Nice! I think there's a global enum called `SaveDialogSelections` which I reckon we could use to replace `saveChanges` and `cancel` to reduce any repeated code, but not super important.

PR #272, opened 2022-09-12
Fix/tag silence on end

tngu0041:
## Summary
Removing the silence buffer before words. Buffers are all at the end of a Word now. This is task 12 from the doc Irmy posted: _"12. Tag silence on end (don't split the silence between end of word A and start of word B"_

<hr/>

Currently, the first word still has the bufferDurationBefore. I left it like that so that it still works with the rest of the code without requiring any further changes. If anyone has any thoughts on this edge case, lmk. (Could also be an SA discussion if we want to remove the bufferDurationBefore property from Word objects completely).

<hr/>

### OS

- [ ] Linux
- [ ] MacOS
- [x] Windows


chloebrett (approved):
Looks good. Glad you caught the first word edge case

I think maybe a small buffer at the start of each word (e.g. max 0.2 seconds for example) could be good because sometimes the start time from the transcription is slightly off and so we miss the start of the word if there’s no buffer. This is a product problem though and we can return to it later. @rileykeane

rileykeane:
> think maybe a small buffer at the start of each word

That's not a bad idea. I'll check with Zac and get back to you.

tngu0041:
Just my 2 cents on this: It might be easier to add the buffer on the video-playback side. E.g. when a user clicks on a word, it'll jump to the startTime of that word minus 0.2s.

This could also mean if the user wants the buffer to exist in the exported EDL, we'll have to manually add in the buffer at every cut (probably; idk exactly how EDL's work)

chloebrett:
Your second sentence is the reason it makes more sense to do at a transcription level than playback level

also playback calculates the active word based on playback time so if we offset it highlights will be wrong - resolvable but not elegantly

rileykeane:
@tngu0041 Heard back from Zac, he's keen to see it with the 0.2s buffer at the start.

tngu0041:
Alright cool. I can add that in.

PR #273, opened 2022-09-13
FIX/EDL-Fix

itsMoxMox:
## Summary

**What is fixed**
- **EDLs are now fully usable.**
- Cut gap and frame gap problems fixed

**What you may encounter, which are intended**
- Some editing software still requires you to link the media.
- The video might not be the right scale when imported

**What problems exist that will not be fixed**
- "FCM: NON-DROP FRAME" vs "FCM: DROP FRAME" - doesn't affect functionality, and is too hard
- Some timestamps may be off by 1~2 frames compared to using _real_ editing software.

**Reviewer Selection**
I just added everyone I know who has some sort of editing software who can test if the EDLs work.

<hr/>

### OS

- [ ] Linux
- [ ] MacOS
- [X] Windows

### Video Editor
- [X] Premiere Pro CC 2018
- [X] Premiere Pro CS6 2016
- [X] Davinci Resolve 2021
- [ ] Final Cut Pro X
- [ ] Your favourite editor not on this list


moxchan:
Oops forgot to adjust a test case for the most recent change bear with me

src/main/export/export.ts
```diff
@@ -1,53 +1,71 @@
 /* eslint-disable no-plusplus */
 import { BrowserWindow } from 'electron';
-import { writeFileSync } from 'fs';
+import { existsSync, writeFileSync } from 'fs';
 import path, { join } from 'path';
-import { secondToTimestamp, padZeros } from '../timeUtils';
+import ffprobe from 'ffprobe';
```
chloebrett: Check if this breaks packaging, I'm pretty sure we were using the static version for this reason

src/main/export/export.ts
```diff
@@ -1,53 +1,71 @@
 /* eslint-disable no-plusplus */
 import { BrowserWindow } from 'electron';
-import { writeFileSync } from 'fs';
+import { existsSync, writeFileSync } from 'fs';
```
chloebrett: If the whole function is async anyway better to use `fs/promises` so that the calls are non blocking

src/main/export/export.ts
```diff
@@ -1,53 +1,71 @@
...
+    let fps = 30; // default to a safe 30fps.
+    if (existsSync(source)) {
+      const videoData = await ffprobe(source, { path: ffprobeStatic.path });
+      fps = fracFpsToDec(videoData.streams[0].avg_frame_rate);
+    }
 
-  const cuts = convertTranscriptToCuts(transcription);
-  const entries = cuts.length;
+    let output = `TITLE: ${title}\nFCM: NON-DROP FRAME\n\n`;
+    // this can be technically incorrect in some rare cases, but it doesn't affect functionality
```
chloebrett: What line does this comment apply to? If it's the above one then move it a line up

src/main/export/export.ts
```diff
@@ -1,53 +1,71 @@
...
+        );
+        /**
+         * Addressed the gap problem by rounding up 1 frame per cut, it will be overwritten in case there is no rounding error,
+         * so this will only work to fill in microgaps. LMK if a less jank solution is wanted.
+         */
 
-      const timelineStart = secondToTimestamp(timeline.start);
-      const timelineEnd = secondToTimestamp(timeline.end);
+        timeline.start = timeline.end;
+        timeline.end = timeline.start + cut.duration;
```
chloebrett: Since `timeline.start` is equal to `timeline.end` due to the line before, can just replace this with `timeline.end += cut.duration`

src/main/export/export.ts
```diff
@@ -65,15 +83,17 @@ export const exportEDL: (
...
-    );
+    // eslint-disable-next-line promise/catch-or-return
+    constructEDL(
+      project.name,
+      project.transcription,
+      project.mediaFilePath,
+      mainWindow
+    )
+      // eslint-disable-next-line promise/always-return
+      .then((edl) => {
```
chloebrett: Change this to async/await

src/main/handlers/helpers/exportUtils.ts
```diff
@@ -25,3 +25,8 @@ export const getExportFilePath: (
 
   return dialogResponse.filePath as string;
 };
+
+export const fracFpsToDec: (fracFps: string) => number = (fracFps) => {
+  const nums = fracFps.split('/').map((item: string) => +item); // get the numbers (separated by / ), and convert to numbers.
```
chloebrett: `Number(item)`

`+item` is a bit 'clever' whereas `Number(item)` is very explicit

src/main/timeUtils.ts
```diff
@@ -7,12 +7,24 @@ export const integerDivide: (a: number, b: number) => number = (a, b) => {
...
 
-  return new Date(num * 1000).toISOString().slice(11, 22).replace('.', ':');
+  const date = new Date(num * 1000);
+  const frame = Math.floor((edlFps * date.getMilliseconds()) / 1000);
+  /**
+   * For the frame gap problem, making timestamps round down 1 frame while having the
+   * input duration round 1 up frame fills in microgaps cleanly.
+   */
+
+  const timestamp = date.toISOString().slice(11, 20).replace('.', ':');
```
chloebrett: What is this doing? Looks a bit hacky and I feel like it might be fragile (e.g. if some component of the date takes up an extra digit in some edge case then it will be off by one)

chloebrett (changes requested):

chloebrett:
I'll test the actual changes tomorrow in class, just added some style comments

moxchan:
Stuff works (hopefully), and yeah I'll admit it reads like it was written on crack - I'll implement the style changes when I can

exquisitebin (commented):
I'm having some issues importing the EDL into Davinci Resolve. I can't seem to be able to link the video clip even if I want to manually. As you can see the clip is already sitting in the media pool.

Not sure what the underlying problem could be. It's possible there's a different way to import the EDL that I haven't tried.

https://user-images.githubusercontent.com/58348527/190061004-167bc5f9-6cb1-4275-b007-546ae25edac6.mp4

src/main/export/export.ts
itsMoxMox: Fixed!

src/main/export/export.ts
itsMoxMox: Implemented! It was originally like that for readability, but I agree that doing it with `+=` doesn't make it unclear

src/main/handlers/helpers/exportUtils.ts
itsMoxMox: Changed it to be something a little bit more declarative, 
`.map(Number)`
still a bit iffy in terms of keeping consistent with any other instances of a cast map in the codebase tho.

chloebrett: `.map(Number)` is fine :)

src/main/timeUtils.ts
itsMoxMox: Changed it to be clear that it converts to a date for ease of parsing hours, mins, sec, etc.

src/main/export/export.ts
itsMoxMox: Done!

src/main/export/export.ts
itsMoxMox: Changed to use `fluent-ffmpeg` instead of `ffprobe` directly, tested packaging and it is A-OKAY! (on my end (on Windows) )

itsMoxMox:
@exquisitebin 
Here's the Resolve process - I think it's because of the _auto set project settings_, double check if that makes it work

https://user-images.githubusercontent.com/55727741/190847214-5dcb03c8-0dea-4908-8429-d03c4872ebd5.mp4

package.json
```diff
@@ -47,7 +47,7 @@
     "copyright": "Copyright © 2022 ${author}",
     "extraMetadata": {
       "//": "--- Prop 'version' will be automatically set by electron-builder ---",
-      "version": "0.0.1"
```
chloebrett: 🤔

src/main/export/__tests__/secondToTimestamp.test.ts
```diff
@@ -1,10 +1,10 @@
-import { secondToTimestamp } from '../../timeUtils';
+import { secondToEDLTimestamp } from '../../timeUtils';
```
chloebrett: nitpick: acronyms in variable and function names should be treated as words i.e. `secondToEdlTimestamp`

src/main/export/export.ts
```diff
@@ -1,18 +1,43 @@
 /* eslint-disable no-plusplus */
 import { BrowserWindow } from 'electron';
-import { writeFileSync } from 'fs';
+import * as promises from 'fs/promises';
```
chloebrett: Just import as fs - `import fs from 'fs/promises`

src/main/handlers/helpers/exportUtils.ts
```diff
@@ -25,3 +25,8 @@ export const getExportFilePath: (
 
   return dialogResponse.filePath as string;
 };
+
+export const fracFpsToDec: (fracFps: string) => number = (fracFps) => {
+  const nums = fracFps.split('/').map(Number); // get the numbers (separated by / ), and convert to numbers.
```
chloebrett: You can array destructure to give the numbers actual meaning

`const [numerator, denominator] = fracFps.split ...`

Also this requires the input be a certain format or it'll crash, so verify that and/or add a docstring explaining

Finally, the name `fracFpsToDec` is not accurate since this always returns an int (semantically it does, even though JS has no concept of an actual int type). Maybe `fractionToInt` would be more accurate as this isn't necessarily FPS specific either

src/main/timeUtils.ts
```diff
@@ -7,12 +7,29 @@ export const integerDivide: (a: number, b: number) => number = (a, b) => {
...
-  return new Date(num * 1000).toISOString().slice(11, 22).replace('.', ':');
+  const date = new Date(num * 1000);
+
+  /**
+   * For the frame gap problem, making timestamps round down 1 frame while having the
+   * input duration round 1 up frame fills in microgaps cleanly.
+   */
+  const frame = Math.floor((edlFps * date.getMilliseconds()) / 1000);
+
+  const timestamp = `${padZeros(date.getUTCHours(), 2)}:${padZeros(
```
chloebrett: Can clean this up a bit. Rename `frame` to `frameNumber` and then

```
const [hours, minutes, seconds, frame] = [date.getUTCHours(), date.getMinutes(), date.getSeconds(), frameNumber].map(value => padZeros(value, 2));

return [hours, minutes, seconds, frame].join(':');
```

chloebrett:
Some more comments, sorry I didn't pick these up earlier, looking better overall

package.json
itsMoxMox: Ok yeah I have no idea how this turned up 💀

src/main/export/export.ts
itsMoxMox: Just imported the necessary functions `{ access, writeFile }`

src/main/handlers/helpers/exportUtils.ts
itsMoxMox: Implemented the changes - `ffprobe` seems to prefer always returning fps in this format, whether average or minimum fps. [Documentation here](https://trac.ffmpeg.org/wiki/FFprobeTips) not sure if we should put this somewhere.

rileykeane:
@moxchan @chloebrett is this good to go?

itsMoxMox:
It should be, at least in terms of functionality - I'll go ahead and resolve the comments

chloebrett (approved):
Looks good!

Irmy:
@moxchan Just checking on this - is this all good to be merged? :)

moxchan:
> @moxchan Just checking on this - is this all good to be merged? :)

I'll merge dev into it and merge it right now, mb

PR #274, opened 2022-09-13
Feature/play pause redux

v1ta1287:
## Summary
PR for storing play/pause state in redux. Work in progress PR, so will definitely need to change some things before merging to develop.

<hr/>

Play/pause state of the video should be in redux. Also users should be able to pause the video by pressing space.
<hr/>

### What did you change?
Added play/pause state to redux. Also added a isSpacePressed state to redux to change play/pause state when space bar is pressed, but I'm not sure if this is the best way to do things.
<hr/>

No
<hr/>

None
<hr/>

Right now how I am getting the video to pause when pressing spacebar is a useEffect in VideoController.tsx that calls the
onClickPlayPause() function whenever the state of spacePressed in redux store is changed. This state is changed when spacebar is pressed (see keyboardShortcutsRegistration.ts). This causes a bug of the video to play when a project is accessed, and its also generally not great code so I was hoping for some suggestions to improve this. 
<hr/>

None
<hr/>

Using spacebar to pause/play the video
<hr/>

### Links to any design documents which you made/used in order to make the decisions that you made.

<hr/>

### OS

- [ ] Linux
- [ ] MacOS
- [x] Windows


src/renderer/store/sharedHelpers.ts
```diff
@@ -33,6 +33,8 @@ export interface ApplicationStore {
   collab: CollabClientSessionState | CollabClientInitialState | null;
   // Op queue session state for pending actions when in a collab session
   opQueue: OpQueueItem[];
+  isVideoPaused: boolean;
```
chloebrett: Read the doc I wrote, this explains roughly what needs to be in the store https://docs.google.com/document/d/1Fowj3DKGy9lQAdRtpT7M4ktUvc7w3LnxWpERjfT38Ac/edit

- flip the boolean so that it's `isVideoPlaying` instead of `isVideoPaused` since pause is more of a default state
- remove the `isSpacePressed` since this shouldn't be in redux
- add `playbackStateLastUpdated: Date` and `playbackTime: number` to the store
- put these three related things (`isVideoPlaying`, `playbackStateLastUpdated`, `playbackTime`) in an interface and put that on the base of the store so they can be handled by a single reducer

src/renderer/keyboardShortcutsRegistration.ts
```diff
@@ -28,6 +30,11 @@ const registerKeyboardHandlers: () => Promise<void> = async () => {
   }
 
   window.onkeydown = (event: KeyboardEvent) => {
+    if (event.key === ' ') {
```
chloebrett: Rather than registering this here, register it as a menu item and use the normal strategy of dispatching an action via an IPC handler on the front end. See things like "select all" or "merge words" for an example

This keyboard shortcuts registration is only for the ones that are reserved by Win/Linux anyway like Ctrl+X for example, and I don't think space will be one of them (but it might)

Then, rather than dispatching an intermediary "space pressed" action, toggle the "is video playing" state. Add a useEffect to the video preview controller so that if the `isVideoPlaying` state changes in redux then it pauses or plays the video accordingly.

Also follow guidelines in the doc I linked regarding updating the `playbackStateLastUpdated` time, etc

chloebrett (changes requested):

v1ta1287:
Hey Chloe, deleted the spacePressed state from redux and did the menu/IPC stuff. Just wanna get everything double checked before I implement seek and forward/backwards.

src/main/menu.ts
```diff
@@ -176,6 +176,14 @@ export default class MenuBuilder {
         },
         enabled: false, // by default, gets updated when a project is entered
       },
+      {
```
chloebrett: Perfect 👍🏻

src/renderer/components/VideoPreview/VideoPreviewController.tsx
```diff
@@ -225,6 +226,15 @@ const VideoPreviewControllerBase = (
     );
   }, [currentProject?.mediaFilePath]);
 
+  useEffect(() => {
```
chloebrett, line 249: This is exactly what I was saying 👍🏻

src/renderer/store/root.ts
```diff
@@ -27,6 +28,7 @@ const rootReducer = combineReducers({
   isUpdateTranscriptionAPIKeyOpened: updateTranscriptionAPIKeyReducer,
   collab: collabReducer,
   opQueue: opQueueReducer,
+  isVideoPlaying: updatePlaybackPauseReducer,
```
chloebrett: I'd just call this `playbackReducer`

chloebrett (commented):
Right on track, nice work!

Irmy:
Just checking in here - looks great from my end - HUGE improvement for the user! I noticed that when the plauback runs to the end of the clip, I have to click space twice to get it to start playing again from the beginning - ideally if the video playback ended and I pressed play again, it should start back from the beginning. But very minor point. Also love that the hot key works regardless of where I click in the transcription page. 

Let me know when this PR is in the final state and I will re-review

v1ta1287:
Thanks for the feedback Irmy. I should be able to add a fix for replaying the video.

src/renderer/store/sharedHelpers.ts
v1ta1287: Hi Chloe, I'm not really too sure what you mean by putting the 3 states in an interface so they can be handled by a single reducer. Don't I need a separate reducer for each state? Otherwise, how would I use useEffect() to only activate on the change of one state.

chloebrett: You don’t need a separate reducer for each. When it comes to useEffect, you can get the variable of interest from useSelector and go as fine grained as you want.

v1ta1287: Is there any templates in the codebase on how I should implement the interface? I had a look at selection but the actions/reducer is really confusing.

chloebrett: If you’ve pulled from latest master selection is an okay example - although it does have the self/others thing. Perhaps a better example is the undoStack reducer - aside from being a bit meta it shows the benefit of updating multiple fields at once within the reducer.

Let me know if you want to pair on this

v1ta1287: Hey Chloe, just pushed a bunch of commits which implements skipping forward/backwards with left and right arrow keys. State seems to be updating correctly and the video preview is able to skip 10 seconds forward/backwards regardless of whether the video is paused or playing. I'm just not sure if the interface/reducer is in the structure you wanted so it'd be good to double check before I finish up this pr.

src/main/menu.ts
```diff
@@ -176,6 +176,30 @@ export default class MenuBuilder {
...
+        id: 'playPause',
+        label: 'Play Or Pause Video',
+        accelerator: 'Space',
+        click: () => {
+          this.mainWindow.webContents.send('toggle-play-pause');
+        },
+      },
+      {
+        id: 'skipForward',
+        label: 'Skip forwards n seconds',
```
chloebrett: Change `n` to the actual amount we skip forward

src/renderer/store/playback/helpers.ts
```diff
@@ -0,0 +1,5 @@
+export interface PlaybackState {
+  playbackPlaying: boolean;
```
chloebrett: This is good - can simplify the names a bit though because the `playback` bit is implied by the key of the object they're in

Something like

```
isPlaying: boolean;
playbackTime: number;
lastUpdated: Date;
```

or even change `playbackTime` to just `time`

src/renderer/store/playback/reducer.ts
```diff
@@ -0,0 +1,52 @@
+import { Reducer } from 'redux';
+import { Action } from '../action';
+import { ApplicationStore, initialStore } from '../sharedHelpers';
+import { VIDEO_PLAYING, VIDEO_SEEK, VIDEO_SKIP } from './actions';
+
+const playbackReducer: Reducer<ApplicationStore['playback'], Action<any>> = (
```
chloebrett, line 14: This is really good! Only comment is that reducers must be pure functions (i.e. the output depends only on the input and the current system state does not affect this, and also the functions do not do any side effects like making API calls, etc).

This has an implication here: it means we can't use `new Date()` inside the reducer itself. This is because running an action on a given state should have the same effect no matter what the external environment looks like (time, state of other parts of the store, etc).

Therefore, if you're going to use `new Date()`, which you should as it's part of the algorithm we need, you need to pass it in as an argument in the action. That way, once an action is made, it could theoretically be re-run as many times as you want on a given application state and the reducer would do the exact same thing every time.

Let me know if this doesn't make sense or you need any further clarification. Basically, what to do = extract out the `currentTime` as a parameter passed in the action payload and call `new Date()` in the action constructors instead.

src/renderer/store/playback/actions.ts
```diff
@@ -0,0 +1,24 @@
...
+export const VIDEO_SKIP = 'VIDEO_SKIP';
+
+export const videoPlaying: (playbackPlaying: boolean) => Action<boolean> = (
+  playbackPlaying
+) => ({
+  type: VIDEO_PLAYING,
+  payload: playbackPlaying,
+});
+
+export const videoSeek: (playbackTime: number) => Action<number> = (
```
chloebrett: regarding comment in the next file - `new Date()` would be called in this function, for example (as well as in `videoSkip` and `videoPlaying`)

chloebrett (changes requested):
Looking really good, you've implemented the algorithm well. Just a few comments about function purity

src/renderer/store/playback/actions.ts
v1ta1287: Hi Chloe, I made the changes requested. Let me know if you see anything else which needs to be changed or if I implemented anything incorrectly. Thanks.

src/renderer/components/Editor/WordComponent.tsx
```diff
@@ -225,6 +227,12 @@ const WordComponent = ({
       }, DOUBLE_CLICK_THRESHOLD);
 
       setPlaybackTime(outputStartTime + 0.01); // add a small amount so the correct word is selected
+      store.dispatch(
+        videoSeek({
+          time: outputStartTime + 0.01,
+          lastUpdated: new Date(),
+        } as UpdatedTimeSeek)
```
chloebrett: You can remove this cast, the type is implied by the argument type of the function. In fact casting can be dangerous and is best to avoid wherever possible because it clears some type errors (see my later comment)

src/renderer/components/Editor/WordComponent.tsx
```diff
@@ -225,6 +227,12 @@ const WordComponent = ({
       }, DOUBLE_CLICK_THRESHOLD);
 
       setPlaybackTime(outputStartTime + 0.01); // add a small amount so the correct word is selected
```
chloebrett, line 155: Do we still need this? I thought "setPlaybackTime" wasn't needed anymore since we will update the playback time via `useEffect` when the store changes

src/renderer/components/Scrubber.tsx
```diff
@@ -92,6 +94,12 @@ const Scrubber = ({
     // not. If it is an array, we take the first value. This does however have a side effect
     // of the value being null, in this case, we just return the current timestamp.
     if (typeof newSliderValue === 'number') {
+      store.dispatch(
+        videoSeek({
+          time: sliderValueToSeconds(newSliderValue),
+          lastUpdated: new Date(),
```
chloebrett, line 100: Optionally you could just make this part of the `videoSeek` function i.e. the `new Date()` bit is called whenever `videoSeek` is run. Either way is fine, the important thing is that the `new Date()` just isn't in the reducer

src/renderer/components/VideoPreview/VideoPreviewController.tsx
```diff
@@ -225,6 +233,29 @@ const VideoPreviewControllerBase = (
...
   }, [currentProject?.mediaFilePath]);
 
+  // If the state of play/pause is changed, play/pause the video preview
+  // if the video is paused, update the time in redux
+  useEffect(() => {
+    if (playState) {
+      play();
+    } else {
+      pause();
+      store.dispatch(
```
chloebrett: Good catch

src/renderer/components/VideoPreview/VideoPreviewController.tsx
```diff
@@ -225,6 +233,29 @@ const VideoPreviewControllerBase = (
...
+      );
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [playState]);
+
+  // if the state of playback time is changed, set the time accordingly for video preview
+  useEffect(() => {
+    setPlaybackTime(timeState);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [timeState]);
```
chloebrett: What dependency is missing?

src/renderer/components/VideoPreview/VideoPreviewController.tsx
```diff
@@ -225,6 +233,29 @@ const VideoPreviewControllerBase = (
...
+      pause();
+      store.dispatch(
+        videoSeek({
+          time: clockRef.current.time,
+          lastUpdated: new Date(),
+        } as UpdatedTimeSeek)
+      );
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [playState]);
```
chloebrett: What dependency is missing here?

src/renderer/editor/skipBackward.ts
```diff
@@ -0,0 +1,13 @@
+import { videoSkip, UpdatedTimeSkip } from 'renderer/store/playback/actions';
+import store from '../store/store';
+
+const skipBackward: () => void = () => {
+  const dispatchState = {
+    addtime: -10,
+    lastUpdated: new Date(),
+    maxDuration: store.getState().currentProject?.transcription?.duration,
+  } as UpdatedTimeSkip;
+  store.dispatch(videoSkip(dispatchState));
```
chloebrett: You can just inline the `dispatchState` if you like, rather than declaring it separately

src/renderer/editor/skipBackward.ts
```diff
@@ -0,0 +1,13 @@
+import { videoSkip, UpdatedTimeSkip } from 'renderer/store/playback/actions';
+import store from '../store/store';
+
+const skipBackward: () => void = () => {
+  const dispatchState = {
+    addtime: -10,
```
chloebrett: `addTime`

src/renderer/editor/skipForward.ts
```diff
@@ -0,0 +1,13 @@
+import { videoSkip, UpdatedTimeSkip } from 'renderer/store/playback/actions';
+import store from '../store/store';
+
+const skipForward: () => void = () => {
+  const dispatchState = {
+    addtime: 10,
+    lastUpdated: new Date(),
+    maxDuration: store.getState().currentProject?.transcription?.duration,
```
chloebrett: What happens if this is `undefined`? See the comment about casting above

src/renderer/editor/skipBackward.ts
```diff
@@ -0,0 +1,13 @@
...
+    lastUpdated: new Date(),
+    maxDuration: store.getState().currentProject?.transcription?.duration,
+  } as UpdatedTimeSkip;
```
chloebrett: There's a difference between casting and declaring a type. Casting says "I know this is an `UpdatedTimeSkip` so even if it doesn't quite match just ignore that". Declaring a type says "This should be an `UpdatedTimeSkip` so throw an error if the thing I assign it to doesn't match". We want to declare, not cast, so we get proper errors. To declare:

```
const dispatchState: UpdatedTimeSkip = { ... };
```

rather than

```
const dispatchState = { ... } as UpdatedTimeSkip;
```

Note that this now warns you about the type of `maxDuration` being incorrect:

Screen Shot 2022-09-24 at 10 56 32 am

src/renderer/editor/togglePlayPause.ts
```diff
@@ -0,0 +1,13 @@
+import { videoPlaying, UpdatedPlaying } from 'renderer/store/playback/actions';
+import store from '../store/store';
+
+const togglePlayPause: () => void = () => {
+  const currState = store.getState().playback.isPlaying;
+  const dispatchState = {
+    isPlaying: !currState,
+    lastUpdated: new Date(),
+  } as UpdatedPlaying;
```
chloebrett: same deal here

src/renderer/pages/Project.tsx
```diff
@@ -61,14 +61,7 @@ const ProjectPage = () => {
...
             <>
-              <VideoController
-                time={time}
-                isPlaying={isPlaying}
-                play={play}
-                pause={pause}
-                seekForward={seekForward}
-                seekBack={seekBack}
-              />
+              <VideoController time={time} isPlaying={isPlaying} />
```
chloebrett: 👍🏻

src/renderer/store/playback/actions.ts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export interface UpdatedTimeSkip {
+  addtime: number;
+  lastUpdated: Date;
+  maxDuration: number;
+}
+
+export const videoPlaying: (
+  playingState: UpdatedPlaying
```
chloebrett, line 24: There are two approaches to calling functions when you have multiple things you want to pass in:

1. Pass each in as positional arguments each with individual types
2. Pass a payload object containing all the data you want to pass in and give it an interface as a type

You've done the second approach which is generally better for large numbers of arguments - so feel free to keep it that way. Just be aware another option would be to pass in each of the parts of the payload as separate arguments - see `editWordStarted` for an example. This might make your call sites a bit simpler since you don't have to name every argument. It's up to you.

src/renderer/store/playback/reducer.ts
```diff
@@ -0,0 +1,77 @@
...
+import {
+  VIDEO_PLAYING,
+  VIDEO_SEEK,
+  VIDEO_SKIP,
+  UpdatedPlaying,
+  UpdatedTimeSeek,
+  UpdatedTimeSkip,
+} from './actions';
+
+const correctTime = (inputTime: number, maxDuration: number) => {
```
chloebrett: We already have a generic `clamp` function which does exactly this.

e.g. instead of calling `correctTime(time, maxDuration)` you would call `clamp(time, 0, maxDuration)`

src/renderer/store/playback/reducer.ts
```diff
@@ -0,0 +1,77 @@
...
+    if (playback.isPlaying) {
+      const timeDifference =
+        (new Date().getTime() - playback.lastUpdated.getTime()) / 1000;
+      return {
+        isPlaying: playback.isPlaying,
+        time: correctTime(
+          playback.time + timeState.addtime + timeDifference,
+          timeState.maxDuration
+        ),
+        lastUpdated: new Date(),
```
chloebrett: replace `new Date()` with `timeState.lastUpdated`

src/renderer/store/playback/reducer.ts
```diff
@@ -0,0 +1,77 @@
...
+          playback.time + timeState.addtime,
+          timeState.maxDuration
+        ),
+        lastUpdated: timeState.lastUpdated,
+      };
+    }
+
+    if (playback.isPlaying) {
+      const timeDifference =
+        (new Date().getTime() - playback.lastUpdated.getTime()) / 1000;
```
chloebrett: replace `new Date()` with `timeState.lastUpdated`

chloebrett (changes requested):
Getting there!

src/main/menu.ts
chloebrett: Also for case consistency: `Skip Forwards 10 Seconds`

src/main/menu.ts
```diff
@@ -176,6 +176,30 @@ export default class MenuBuilder {
...
+        accelerator: 'Right',
+        click: () => {
+          this.mainWindow.webContents.send('initiate-skip-forward');
+        },
+      },
+      {
+        id: 'skipBackward',
+        label: 'Skip backwards n seconds',
```
chloebrett: Same deal here

src/main/menu.ts
```diff
@@ -176,6 +176,30 @@ export default class MenuBuilder {
         },
         enabled: false, // by default, gets updated when a project is entered
       },
+      {
+        id: 'playPause',
+        label: 'Play Or Pause Video',
```
chloebrett: At the very least, `Play or Pause Video`

But better would be `Play/Pause Video`

src/main/menu.ts
```diff
@@ -176,6 +176,30 @@ export default class MenuBuilder {
         },
         enabled: false, // by default, gets updated when a project is entered
       },
+      {
```
chloebrett: I think these playback actions should have their own separator so they appear separate on mac - otherwise they're grouped with the word correction actions which is a bit weird.

Screen Shot 2022-09-24 at 11 06 48 am

chloebrett: Even better would be if they had their own `Playback` menu, I don't really think it makes sense for them to be in the `edit` menu since they don't actually modify the project at all

chloebrett (changes requested):
Just tested locally. The actual functionality is working really well, just some comments about the menu

src/renderer/components/Editor/WordComponent.tsx
v1ta1287: True, I did think about this but I didn't want to remove anything incase I broke the program.

src/renderer/components/VideoPreview/VideoPreviewController.tsx
v1ta1287: setPlaybackTime

v1ta1287: If I include the dependency, it creates an infinite loop.

src/renderer/store/playback/reducer.ts
v1ta1287: Okay noted

src/renderer/store/playback/reducer.ts
v1ta1287: forgot about this

src/main/menu.ts
v1ta1287: Yeah I remember you telling me to do this earlier, I just forgot to do it, cheers for the reminder

src/main/menu.ts
v1ta1287: Okay

src/main/menu.ts
v1ta1287: Okay no worries

src/renderer/store/playback/actions.ts
v1ta1287: Okay

src/renderer/components/Editor/WordComponent.tsx
v1ta1287: Okay

src/renderer/components/Scrubber.tsx
v1ta1287: Okay

v1ta1287:
Made some changes fixing up the interfaces/bugs and added a separate playback menu to the toolbar. Let me know what you think.

src/main/menu.ts
```diff
@@ -455,6 +455,36 @@ export default class MenuBuilder {
                 },
               ],
       },
+      {
```
chloebrett: This menu needs to be added on mac as well, currently it's only displayed on linux / windows

src/renderer/components/VideoPreview/VideoPreviewController.tsx
chloebrett: You can fix this by wrapping `setPlaybackTime` in a `useCallback`. Best to add all relevant dependencies to the useEffect because if you don't, it'll be calling an old version of the function which had a closure around some outdated state. Kind of hard to explain. General practice is just wrap functions in `useCallback` and variable assignments in `useMemo` so they don't change every re-render

Let me know if you need help with this

chloebrett: Let me know if you need help with this

src/renderer/store/playback/reducer.ts
```diff
@@ -0,0 +1,70 @@
...
+        (timeState.lastUpdated.getTime() - playback.lastUpdated.getTime()) /
+        1000;
+      return {
+        isPlaying: playback.isPlaying,
+        time: clamp(
+          playback.time + timeState.addtime + timeDifference,
+          0,
+          timeState.maxDuration
+        ),
+        lastUpdated: timeState.lastUpdated,
```
chloebrett, line 62: 👍🏻 nice to see this fixed

chloebrett (changes requested):
Getting so close. Just respond to the comments that are still open / fix the issues they mention, and merge from develop and fix conflicts, then this is good to go

src/main/menu.ts
v1ta1287: I couldn't find any documentation on how to add this to mac, and I'm not too sure how I can test it either since I'm on windows. Where should I look in the codebase for this?

src/renderer/store/playback/reducer.ts
v1ta1287: I tried wrapping setPlaybackTime in a useCallback, but this forced me to wrap pretty much every other function in a useCallback.

chloebrett: Doing that is fine - it means the functions don't get re-defined on every re-render. If you can't get that to work then leave it as is, but there could be undefined behaviour

src/main/menu.ts
chloebrett: See how the mac (sometimes called 'darwin') menu is constructed. You wont be able to test it but it should be pretty obvious if it's right or not. It's all in menu.ts

v1ta1287: Okay think I have added the menu item to mac, let me know if you are able to see it

src/renderer/store/playback/reducer.ts
v1ta1287: This is complete

v1ta1287:
Merged from develop and everything seems to be working okay.

src/renderer/store/playback/reducer.ts
v1ta1287: Merged from develop and everything seems to be okay

LiamTodd (approved):
Good stuff Andy, works fine on my machine and the code looks great

src/renderer/components/Editor/WordComponent.tsx
chloebrett: Try without, I think it will be double dispatching otherwise

chloebrett (changes requested):
Fix the casts then this is good to go

fwang1703 (approved):
Looks awesome! Love your work, and the Mac menu is working for me. (Only small thing would be to fix up the type casting like Chloe mentioned, but otherwise everything looks really good!)

v1ta1287:
Okay, I have removed casting.

Irmy (approved):
Looks good Andy! Awesome work!! Huge effort

chloebrett (approved):

PR #276, opened 2022-09-14
feature: fix take detection ui

MutazAshhab:
Depends on https://github.com/chloebrett/mlvet/pull/265

Updates the UI to match the figma mock ups

What it looks like:

https://user-images.githubusercontent.com/86769131/190136301-8d6c84ed-7431-4dda-99b7-47ca43c706b6.mp4

------
### OS

- [X] Linux
- [ ] MacOS
- [ ] Windows

### DIFF
--- a/src/renderer/components/Editor/SquareBracket.tsx
+++ b/src/renderer/components/Editor/SquareBracket.tsx
@@ -1,12 +1,14 @@
 import { Box } from '@mui/system';
 import { useState } from 'react';
+import colors from 'renderer/colors';
 import SquareBracketHover from './SquareBracketHover';
 
 interface Props {
   isLast: boolean;
+  isTakeGroupOpened: boolean;
 }
 
-const SquareBracket = ({ isLast }: Props) => {
+const SquareBracket = ({ isLast, isTakeGroupOpened }: Props) => {
   // const bottomWidth = isLast ? '2px' : '0px';
   const bottomWidth = '2px';
 
@@ -27,7 +29,7 @@ const SquareBracket = ({ isLast }: Props) => {
         height: '60px',
         width: '15px',
         borderStyle: 'solid',
-        borderColor: '#FFB355',
+        borderColor: isTakeGroupOpened ? '#FFB355' : colors.grey[500],
         borderWidth: '0px',
         borderLeftWidth: '2px',
         borderTopWidth: '2px',
--- a/src/renderer/components/Editor/SquareBracket.tsx
+++ b/src/renderer/components/Editor/SquareBracket.tsx
@@ -7,7 +7,8 @@ interface Props {
 }
 
 const SquareBracket = ({ isLast }: Props) => {
-  const bottomWidth = isLast ? '2px' : '0px';
+  // const bottomWidth = isLast ? '2px' : '0px';
+  const bottomWidth = '2px';
 
   const [isHovering, setIsHovering] = useState(false);
 
--- a/src/renderer/components/Editor/SquareBracket.tsx
+++ b/src/renderer/components/Editor/SquareBracket.tsx
@@ -1,0 +1,25 @@
+import { Box } from '@mui/system';
+
+interface Props {
+  isLast: boolean;
+}
+
+const SquareBracket = ({ isLast }: Props) => {
+  const bottomWidth = isLast ? '2px' : '0px';
+  return (
+    <Box
+      sx={{
+        height: '60px',
+        width: '15px',
+        borderStyle: 'solid',
+        borderColor: '#FFB355',
+        borderWidth: '0px',
+        borderLeftWidth: '2px',
+        borderTopWidth: '2px',
+        borderBottomWidth: bottomWidth,
+      }}
+    />
+  );
+};
+
+export default SquareBracket;
--- a/src/renderer/components/Editor/SquareBracket.tsx
+++ b/src/renderer/components/Editor/SquareBracket.tsx
@@ -6,15 +6,18 @@ interface Props {
 }
 
 const SquareBracketBackground = styled(Box)({
-  height: '100%',
-  width: '16px',
+  height: 'inherit',
+  width: '8px',
   opacity: 0,
+  borderRadius: '8px',
+  backgroundColor: '#FFB355',
+
+  // positioning
+  position: 'absolute',
+  left: '17px',
 
   '&:hover': {
     opacity: 0.5,
-    backgroundColor: '#FFB355',
-    borderRadius: '8px',
-    borderWidth: '10px',
   },
 });
 
--- a/src/renderer/components/Editor/SquareBracket.tsx
+++ b/src/renderer/components/Editor/SquareBracket.tsx
@@ -1,28 +1,24 @@
-import styled from '@emotion/styled';
 import { Box } from '@mui/system';
+import { useState } from 'react';
+import SquareBracketHover from './SquareBracketHover';
 
 interface Props {
   isLast: boolean;
 }
 
-const SquareBracketBackground = styled(Box)({
-  height: 'inherit',
-  width: '8px',
-  opacity: 0,
-  borderRadius: '8px',
-  backgroundColor: '#FFB355',
-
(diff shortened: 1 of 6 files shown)


emmapainter (approved):
UI looking good! This knocks off 6 tickets which is awesome. I'm only approving the functionality/UI as a PM - wait for an actual code review before merging :)

ghost:
For future reference, I believe the GIT practice is: 
When a PR is dependent on another open PR, rather than merge into develop, merge into that branch.
That way @zwan0202 's changes won't appear in your PR diff

src/renderer/components/Editor/SquareBracket.tsx
```diff
@@ -0,0 +1,45 @@
...
+  };
+
+  return (
+    <Box
+      id="squareBracket"
+      sx={{
+        height: '60px',
+        width: '15px',
+        borderStyle: 'solid',
+        borderColor: isTakeGroupOpened ? '#FFB355' : colors.grey[500],
```
ghost: Please define this colour with other colours if possible. For reuse

src/renderer/components/Editor/SquareBracketHover.tsx
```diff
@@ -0,0 +1,90 @@
...
+        opacity: opacityValue,
+        borderRadius: '8px',
+        backgroundColor: '#FFB355',
+
+        // positioning
+        position: 'relative',
+        left: '-5px',
+        top: '-8px',
+
+        '&:hover': {
```
ghost: Is this CSS logic needed as well as the JS logic used above?

src/renderer/components/Editor/WordComponent.tsx
```diff
@@ -31,7 +31,7 @@ const makeWordInner = (isDragActive: boolean, isInInactiveTake: boolean) =>
   styled('div')({
     display: 'inline-block',
     cursor: isInInactiveTake ? 'pointer' : 'text',
-    color: colors.white,
+    color: isInInactiveTake ? colors.grey[600] : colors.white,
```
ghost, line 34: is logic enabled all the time (for words that aren't in takes at all) or just for words in takes? (sorry I'm too blind to tell in the video)

src/renderer/components/Editor/TakeGroupComponent.tsx
```diff
@@ -91,38 +103,39 @@ const TakeGroupComponent = ({
...
-        onWordMouseMove={onWordMouseMove}
-        isWordBeingDragged={isWordBeingDragged}
-        mousePosition={mousePosition}
-        nowPlayingWordIndex={nowPlayingWordIndex}
-        transcription={transcription}
-        selectionSet={selectionSet}
-        transcriptionIndex={transcriptionIndex}
-        {...passThroughProps}
-      />
+      <CustomRowStack sx={{ justifyContent: 'flex-start' }}>
```
ghost: Just checking that this CustomRowStack is necessary? It only seems to have 1 child component

src/renderer/components/Editor/TakeComponent.tsx
```diff
@@ -71,99 +81,98 @@ const TakeComponent = ({
...
+          {isTakeGroupOpened || isActive ? (
+            <>
+              {!isFirstTimeOpen && isTakeGroupOpened && (
+                <Avatar
+                  onClick={onSelectTake}
+                  sx={{
+                    height: 22,
+                    width: 22,
+                    fontSize: 12,
+                    color: '#1D201F',
```
ghost: Please extract these colours to variables for reusability

ghost:
Love you work. Few tiny comments to peruse. 
.
As a token of my appreciation, here is your name with all letters replaced with sheep:
🐑🐑🐑🐑🐑        🐑🐑-🐑🐑🐑🐑🐑🐑

src/renderer/components/Editor/WordComponent.tsx
chloebrett: `isInInactiveTake` is false if a word is not in a take at all - for `isInInactiveTake` to be true, the word has to be (a) in a take and (b) that take has to be inactive

src/renderer/components/Editor/SquareBracket.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+interface Props {
+  isLast: boolean;
+  isTakeGroupOpened: boolean;
+}
+
+const SquareBracket = ({ isLast, isTakeGroupOpened }: Props) => {
+  const bottomWidth = isLast || !isTakeGroupOpened ? '2px' : '0px';
+
+  const [isHovering, setIsHovering] = useState(false);
```
chloebrett, line 14: `isHovering` seems like it's only being used to adjust the opacity of the children, and it's a fair bit of JS logic to accomplish something that CSS is better suited to. Instead, you could define this in CSS. Something along the lines of (syntax might not be correct but the idea is there)

```
&:hover {
  & .innerBracket {
    opacity: 0.5;  
  }
}
```

Which would work if you used `styled()` to define the "squareBracket" box as we have for other components, and give the children a classname

chloebrett (commented):
Agree with Hugh's comments, overall nice work

chloebrett:
I changed your branch base to point to the multi sentence take detection one - makes the diff only show your changes, and you don't want to be merging this to develop anyway until the multi sentence take detection has landed. It doesn't mean that you should actually merge this into the MSTD branch, but instead that you should wait until that one is merged before merging this

src/renderer/components/Editor/SquareBracket.tsx
MutazAshhab: I really liked your solution and I gave this a shot. The class names have to be unique so that it could only turn on the hover of SquareBracketHover component. If all the class names are the same then all the components will have the hover effect when you hover over 1.

I tried using a template literal so that I can have unique class names per SquareBracket component I got a syntax error.

Here is a screenshot: 
image


Im going to stash these changes and i'll show you them tomorrow

src/renderer/components/Editor/TakeGroupComponent.tsx
MutazAshhab: No, i just checked

With CustomRowStack:

image


Without CustomRowStack:

image

src/renderer/components/Editor/SquareBracketHover.tsx
MutazAshhab: Nope

src/renderer/components/Editor/SquareBracket.tsx
chloebrett: Very cool, you're so close 😅 

Because it's an object, if the key is a variable and not a string literal then you need to put it in square brackets. I'll give you a more generic example and you can apply to your specific situation

Example:

```javascript
const keyName = 'myKey';
const object = {
    [keyName]: 'someValue'
}
```

The above will produce this object: `{ "myKey": "someValue" }`
Whereas without the square brackets you get: `{ "keyName": "someValue" }`

This is because it's ambiguous otherwise whether you mean "I want my key to be the literal string 'keyName'" versus "I want my key to be whatever the value of the keyName variable is, which is 'myKey'"

MutazAshhab: The javascript portion of that worked, 

image


But it is not hovering, when i use the dev tools it shows nothing up. 

image


I pushed those changes on another branch, I made a PR so you can see the diff: https://github.com/chloebrett/mlvet/pull/279. Let me know if you can spot whats wrong.

Code looks like this:

```javascript
  const hoverCss = { '&:hover': { [componentSelector]: { opacity: 0.5 } } };

  console.log(hoverCss);

  const SquareBracketBox = styled(Box)({
    height: '60px',
    width: '15px',
    borderStyle: 'solid',
    borderColor: isTakeGroupOpened ? colors.yellow[500] : colors.grey[500],
    borderWidth: '0px',
    borderLeftWidth: '2px',
    borderTopWidth: '2px',
    borderBottomWidth: bottomWidth,

    ...hoverCss,
  });
```

PR #277, opened 2022-09-14
Feature/edit word shortcut e

jhea0008:
## Summary

Replaces the double click to edit word with a keyboard shortcut mapped to 'E'

<hr/>

Usability? idk tbh I prefer double clicking over this shortcut, but I think the best solution would be to add a right-click menu to the app and include it there. Might work on a right-click menu if I get time / PM approval
<hr/>

### What did you change?
the function to edit a word is now called from the edit drop down menu, or with the shortcut 'e'. This should only be able to be called when a single word is selected, and is not currently being edited

<hr/>

2 main new functions here. One to determine when the shortcut/menu option is available to be pressed/clicked ("setMenuButtonEnabled"), and one to actually perform the action when it is pressed/clicked. ("editWord")
<hr/>

Do I need tests for this? 
<hr/>

### OS

- [ ] Linux
- [X] MacOS
- [X] Windows

### DIFF
--- a/src/renderer/editor/editWord.ts
+++ b/src/renderer/editor/editWord.ts
@@ -1,0 +1,19 @@
+import store from '../store/store';
+import { editWordStarted } from '../store/editWord/actions';
+import { getSelectionRanges } from './selection';
+
+const { dispatch } = store;
+
+const editWord = () => {
+  const words = store.getState().currentProject?.transcription?.words;
+  if (words === undefined) {
+    return;
+  }
+  const ranges = getSelectionRanges();
+  // Assume that only one word is selected, thus there is only one range and start/end indexes are the same
+  const index = ranges[0].startIndex;
+  const { word } = words[index];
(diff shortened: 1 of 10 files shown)


Irmy (approved):
Love it! The E mapping was requested by Zac - carry over from you know who software :D  Love that you added it to the menu drop down too - thanks for that

src/renderer/editor/editWord.ts
```diff
@@ -0,0 +1,19 @@
...
+import { getSelectionRanges } from './selection';
+
+const { dispatch } = store;
+
+const editWord = () => {
+  const words = store.getState().currentProject?.transcription?.words;
+  if (words === undefined) {
+    return;
+  }
+  const ranges = getSelectionRanges();
```
chloebrett, line 12: Just to be on the safe side, early exit if `ranges.length === 0`

chloebrett (approved):
Code looks good just one comment

PR #282, opened 2022-09-16
Refactor selection to be a single range

chloebrett:
~~Depends on https://github.com/chloebrett/mlvet/pull/281~~


jhea0008 (approved):
New ticket raised in clickup for re-adding shift-selecting single ranges. Other than that, all good

PR #283, opened 2022-09-16
Per-character highlight

chloebrett:
https://user-images.githubusercontent.com/6735055/190565952-f1aa4c70-7c0c-44b8-b288-947c50b8dcea.mov

Glitches a little sometimes, showing the whole selection for one frame before it shows the partial selection.


Irmy (approved):
Looks great thankyou! Apologies for the delayed review

PR #284, opened 2022-09-17
feature: fix take detection ui

MutazAshhab:
~~Depends on https://github.com/chloebrett/mlvet/pull/265~~

## **Note:**

This is the second time I open a PR for this feature. Here is the first one https://github.com/chloebrett/mlvet/pull/276. The reason behind this is because that branch had `feature/multi-sentence-take-detection` as its base which was merged and deleted. I was not able to restore the branch nor re-open that PR and change the base branch. 

--------

Updates the UI to match the figma mock ups

What it looks like:

https://user-images.githubusercontent.com/86769131/190136301-8d6c84ed-7431-4dda-99b7-47ca43c706b6.mp4

------
### OS

- [X] Linux
- [ ] MacOS
- [ ] Windows


MutazAshhab:
The PR changes have been approved in #276 however i'd like someone to test it after i merged develop into the branch since it was a massive merge.

src/renderer/components/Editor/SquareBracket.tsx
```diff
@@ -0,0 +1,45 @@
...
+  isLast: boolean;
+  isTakeGroupOpened: boolean;
+}
+
+const SquareBracket = ({ isLast, isTakeGroupOpened }: Props) => {
+  const bottomWidth = isLast || !isTakeGroupOpened ? '2px' : '0px';
+
+  const [isHovering, setIsHovering] = useState(false);
+
+  const handleMouseOver = () => {
```
chloebrett: If you're not going to memoise them I think it's fine to just inline them since you only use them once

i.e. `onMouseOver={() => setIsHovering(true)}`

chloebrett:
When I open a project for the first time, all the takes show but the full take group UI isn't there (with the delete group button, etc)

Screen Shot 2022-09-18 at 11 55 03 am

May just by the project I'm using but see if you can replicate

MutazAshhab:
> When I open a project for the first time, all the takes show but the full take group UI isn't there (with the delete group button, etc)
> 
> Screen Shot 2022-09-18 at 11 55 03 am
> 
> May just by the project I'm using but see if you can replicate

That is actually the wanted behaviour, here are the [figma mockups](https://www.figma.com/file/vPh49OC7hJpMD7cQMAnBUo/MLVET-UI-Design?node-id=1034%3A2148)

And here is a screenshot of the figma mockups if you dont have access:
image

src/renderer/components/Editor/SquareBracket.tsx
MutazAshhab: I've done this in https://github.com/chloebrett/mlvet/pull/284/commits/a473d51cd03cedd7a5a0d82c563dabe7052d6deb

PR #286, opened 2022-09-18
FEATURE/Paste Replaces Highlighted Selection

itsMoxMox:
## Summary

Pasting now replaces the highlighted selection of words, unless there is only a single word highlighted, i.e. "Nothing is highlighted"

This is because if nothing is highlighted, it exists in code as _the current word is highlighted_.

It should be easy to adjust the parameters for which pasting replaces the selection.

<hr/>

### OS

- [ ] Linux
- [ ] MacOS
- [X] Windows

### DIFF
--- a/src/renderer/editor/clipboard.ts
+++ b/src/renderer/editor/clipboard.ts
@@ -49,7 +49,6 @@ export const pasteText: () => void = () => {
   if (clipboard.length > 0) {
     pasteWord(range, clipboard);
   }
-  console.log(range, clipboard.length, clipboard);
 
   // TODO(chloe): should also seek to the start of the pasted text.
 };
--- a/src/renderer/editor/clipboard.ts
+++ b/src/renderer/editor/clipboard.ts
@@ -1,4 +1,3 @@
-import { getLengthOfRange } from 'renderer/utils/range';
 import { Word } from '../../sharedTypes';
 import { clipboardUpdated } from '../store/clipboard/actions';
 import store from '../store/store';
@@ -10,9 +9,13 @@ const { dispatch } = store;
 
 const pasteWord = (afterWordIndex: number, clipboard: Word[]) => {
   const { currentProject } = store.getState();
+  const wordIndexRange = {
+    startIndex: afterWordIndex,
+    endIndex: afterWordIndex,
+  }; // Need to convert number to index, to conform to param type.
 
   if (currentProject && currentProject.transcription) {
-    dispatchOp(makePasteWord(afterWordIndex, clipboard));
+    dispatchOp(makePasteWord(wordIndexRange, clipboard));
   }
 };
 
--- a/src/renderer/editor/clipboard.ts
+++ b/src/renderer/editor/clipboard.ts
@@ -1,4 +1,4 @@
-import { Word } from '../../sharedTypes';
+import { IndexRange, Word } from '../../sharedTypes';
 import { clipboardUpdated } from '../store/clipboard/actions';
 import store from '../store/store';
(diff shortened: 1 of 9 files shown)


src/renderer/editor/clipboard.ts
```diff
@@ -46,7 +46,12 @@ export const pasteText: () => void = () => {
   const { clipboard } = store.getState();
 
   // Paste after the last word in the selection
-  const { endIndex } = range;
+  const { startIndex, endIndex } = range;
+
+  if (clipboard.length && endIndex - startIndex > 1) {
```
chloebrett: `if (clipboard.length > 0 && getLengthOfRange(range) > 1) {`

src/renderer/editor/clipboard.ts
```diff
@@ -46,7 +46,12 @@ export const pasteText: () => void = () => {
...
+    // only replace if you're highlighting more than one word.
+    dispatchOp(makeDeleteSelection(range));
```
chloebrett: The problem with dispatching two ops at once (this one plus the paste word one) is that it requires two separate undos. Instead, make a new op type that does both the delete selection and then the paste word action in a single op. (or, just modify the existing paste op)

chloebrett (changes requested):

src/renderer/editor/clipboard.ts
itsMoxMox: Edit: The two-undo issue is fixed, mbmb

I reimplemented it at the reducer level, since it was kind of annoying to do at the op level, check [commit 91f2edb](https://github.com/chloebrett/mlvet/pull/286/commits/91f2edb91a89ac186f480b3d88431408eb1d3c4e) if you care enough.

At the reducer level, I save having to specify the precise differences between a paste which deletes, and a paste which doesn't.

src/renderer/editor/clipboard.ts
itsMoxMox: Split into different sections, where they are each more relevant.

Don't initiate paste if you have nothing in your clipboard:
https://github.com/chloebrett/mlvet/blob/4fb7ad2dc3f83e9f97a2d7fc4f22df4cfda59a76/src/renderer/editor/clipboard.ts#L52-L54

and only perform a deleting paste if your selection is more than a word:
https://github.com/chloebrett/mlvet/blob/4fb7ad2dc3f83e9f97a2d7fc4f22df4cfda59a76/src/renderer/store/transcriptionWords/reducer.ts#L81-L88

chloebrett:
I was playing around this locally and it kept throwing errors, I made a little fix PR that I'm pretty sure fixes all the bugs I found. https://github.com/chloebrett/mlvet/pull/288

Apologies for doing it this way instead of just as comments, this was just the result of trying to fix it locally and thought it might be useful

Irmy (changes requested):
I can see the convo above already is in the midst of debugging things so won't add much from my end here - here is the error I got if it helps 
image

chloebrett (approved):
Barack Obama awarding himself a medal moment, approving now that the fixes I did are merged in

chloebrett:
@Irmy re-test

Irmy:
Still crashing for me :( Here is a video if it helps. In this it crashed when I was just trying to paste normally - so only 1 word was highlighted so I expected for it to paste normally after the word. ( I guess that brings up the question would a user ever want to paste over and replace just 1 word? How do we differentiate between normal paste and paste/override?)

Irmy (approved):
It works so well!!! Mega team effort @itsMoxMox and @chloebrett . It works so smooth with the edit/restore marker too.

PR #288, opened 2022-09-18
bugfixes on "paste replaces highlighted"

chloebrett:


src/renderer/utils/words.ts
```diff
@@ -15,10 +16,22 @@ export const isInOriginalOrder: (
       return true;
     }
 
+    // edge case, only one word
+    if (getLengthOfRange(range) === 1 || words.length === 1) {
+      return true;
+    }
+
+    const prevWordOriginalIndex =
```
chloebrett, line 24: These changes aren't strictly required but I realised there was a small bug in this function that was getting hit because of the bugs in the other stuff, so I fixed that as well

itsMoxMox (approved):

PR #293, opened 2022-09-20
Feature/select sentence hotkey

jhea0008:
## Summary

New hotkey: shift-command/control-a
This expands the current selection up to the next end-sentence marker on either side if not already a end-sentence marker. 
Thus, if you've selected a word, it will select the sentence that the word is part of. 
If you have selected across a sentence boundary, it will select both sentences that's a part of
The option is greyed out and can't be used when nothing is selected

<hr/>

cool new feature: https://app.clickup.com/t/2z4q96a

<hr/>

### Known Issues
Not sure if this is a bug, but the sentence select will select a delete marker that's at the beginning of a sentence. Not sure how this affects behaviour in terms of user interactions with it. 
<hr/>

Screen Shot 2022-09-20 at 19 02 44

<hr/>

### Explain things which the reviewers are likely to need explaining. 

Pulled out the sentence-detection logic in take detection into a shared utils file, and used this for sentence select. 

<hr/>

select-sentence logic is the key section to review for this one. Open to any changes to simplify logic or care for edge cases I haven't thought of
<hr/>

### OS

- [ ] Linux
- [X] MacOS
- [x] Windows


src/renderer/StoreChangeObserver.tsx
```diff
@@ -156,6 +156,11 @@ const StoreChangeObserver = () => {
     ipc.setEditWordEnabled(editWordEnabled);
   }, [selfSelection, editWordIndex]);
 
+  useEffect(() => {
+    const selectSentenceEnabled = getLengthOfRange(selfSelection) > 0;
```
chloebrett: Should we grey out the button if the current sentence is already selected? Since it would be idempotent in that case

src/renderer/editor/selection.ts
```diff
@@ -26,3 +27,30 @@ export const selectAllWords: () => void = () => {
...
 };
+
+export const selectSentence: () => void = () => {
+  const { currentProject } = store.getState();
+
+  if (currentProject === null || currentProject?.transcription === null) {
+    return;
+  }
+  let { startIndex, endIndex } = store.getState().selection.self;
+  endIndex -= 1; // Account for the fact that the selection is exclusive of the end index
```
chloebrett: nice copilot comment 😉

src/renderer/editor/selection.ts
```diff
@@ -26,3 +27,30 @@ export const selectAllWords: () => void = () => {
...
+  startIndex -= 1; // Move past the current word to account for the edge case of selecting a sentence end word
+  let startWord = currentProject.transcription.words[startIndex];
+  while (!(checkSentenceEnd(startWord) && startWord.deleted === false)) {
```
chloebrett: Clearer as `while (!checkSentenceEnd(startWord) || startWord.deleted) {`

src/sharedUtils.ts
```diff
@@ -132,3 +132,14 @@ export const isInInactiveTake: (
 
 export const sleep: (seconds: number) => Promise<void> = (seconds) =>
   new Promise((resolve) => setTimeout(resolve, seconds * 1000));
+
+export const checkSentenceEnd = (word: Word): boolean => {
```
chloebrett: Does this syntax work? I've seen it used for `function` declarations but surprised it works for const. If it works then all good but if it doesn't here's what we normally use

`export const checkSentenceEnd: (word: Word) => boolean = (word) => {`

chloebrett (commented):
Code looks pretty good

Irmy (changes requested):
Looks really good!! Also the 'bug' you mentioned with the delete marker - I have no issue with it since if i was to copy and paste elsewhere, the delete marker doesnt carry through so its all good.

I did have a bug though where it wasn't working JUST for the very first sentence of the transcript. I tried it on a few different projects and always the very first sentance the hot key wouldn't work. 

Also ill check with Zac re the cross sentance border logic - I can see a use for it for sure, but I can also see how a user can feel like it's unexpected behaviour

Irmy:
Got feedback from Zac - could we please change the cross-sentence border logic to only select 'forward' rather than the previous AND next sentence. 
So If i highlight "Hello. This is my....." only extend the highlight to the end of "this is my" sentence.

chloebrett:
> Got feedback from Zac - could we please change the cross-sentence border logic to only select 'forward' rather than the previous AND next sentence. So If i highlight "Hello. This is my....." only extend the highlight to the end of "this is my" sentence.

What if the sentences are "One two three. Four five six seven eight." and you've highlighted "three. Four five"? Surely we want to make sure the end state is a set of complete sentences rather than just hanging last words on the first sentence

src/renderer/editor/selection.ts
jhea0008: Hahaha I was pretty impressed it came up with this one

jhea0008:
> I did have a bug though where it wasn't working JUST for the very first sentence of the transcript. I tried it on a few different projects and always the very first sentance the hot key wouldn't work.

Good catch Irmy, I'll get that fixed

Irmy:
> What if the sentences are "One two three. Four five six seven eight." and you've highlighted "three. Four five"? Surely we want to make sure the end state is a set of complete sentences rather than just hanging last words on the first sentence
Yep agree! So in that case the jump highlight would extend to include "three. Four five six seven eight." making sre that "one two" are _not_ included in the hotkey extension

chloebrett:
@Irmy what you’ve suggested would contain incomplete sentences though. Do you get my drift

Irmy:
> 

Yeah I get you, thats fine. Because having the sentence select in both 'backwards' and 'forwards' direction feels unexpected. It feels more expected behaviour to select the next sentence - Zac's feedback was "just select the next sentence is good."  Both logics certainly are useful in different scenarios but the one Zac mentioned is more concerned with extending the right boundary.

src/renderer/editor/selection.ts
jhea0008: Should be fixed now

src/sharedUtils.ts
jhea0008: Idk, but I changed it to normal for consistency anyway

src/renderer/StoreChangeObserver.tsx
jhea0008: Yep, sounds good. Added this functionality

chloebrett (approved):

chloebrett:
Tested locally, looks good and works really well. I am strongly on team "the end state should be a set of fully selected sentences" - which is what the code changes do currently - if we want to make it select forwards only I think that should be a team decision based off user/client feedback. To me, the existing behaviour is exactly as expected. Nice work @jhea0008

exquisitebin (approved):
Looks great Justin, functionality works well. I think I agree with Chloe on the selection of both forward and backward of multiple sentences. The alternative also makes sense too if that's what the client wants.

Irmy (approved):

PR #294, opened 2022-09-20
Spike/vosk rewrite

ghost:
## Summary

A PR to add local transcription using Vosk models and a rewritten NodeJS vosk API.

Chunky I know but this contains no frontend code, just the changes needed to get vosk working.

Should work packaged too but as this PR has no frontend code and model has to be downloaded and inserted, can be a bit of a process.

**NOTE: Please read the requirements below to test this locally**

<hr/>

## Testing requirements

First `yarn`.

You will need to download 2 different assets from [this folder on the drive](https://drive.google.com/drive/u/1/folders/11iA_znZmAT8vhiCjTO7WHtOkkarB14WO)

1. The Vosk model (this is 2GB)
2. The Vosk DLLs

Each of them needs to be downloaded and then the contents of the zip file extracted to the assets folder.
Eg: `./assets/voskModel/vosk-model-en-us-0.22/...etc` and `./assets/voskDLLs/lib/...etc`

**---** 

**Additionally** you will need to change the frontend code to use `voskTranscribeFunction` instead of `assemblyAiTranscribeFunction`. To do this:

Change line 14 of `./src/main/handlers/media/transcribe.ts` to be:
`[TranscriptionEngine.ASSEMBLYAI]: voskTranscribeFunction,`
Note: you will also need to import `voskTranscribeFunction`

<hr/>

## Design Docs

https://docs.google.com/document/d/1lRisa9bl972S3EctbKoJ6kxBQqHPFJ3UIKwwLPrTlSI/edit

<hr/>

## OS

- [ ] Linux
- [x] MacOS ARM
- [x] MacOS Intel
- [x] Windows


LiamTodd:
This is some chad behaviour Hugh, awesome work.
It works almost perfectly on my machine with a one issue and a couple of noteworthy things:
1. Random underscores are inserted into the transcription.
3. The old punctuation function isn't being used, so it looks a bit janky. Quick and easy fix though, if we are sticking to the old punctuating code. I think Kate has since done some work on it, although I may be wrong.
4. I get 'MLVET (not responding)' at the top of the app window while vosk is transcribing. This is not something new - it was happening with the old (broken) implementation of vosk.

chloebrett:
The underscores are just the early UI for the long pause manipulation lol

ghost:
> This is some chad behaviour Hugh, awesome work. It works almost perfectly on my machine with a one issue and a couple of noteworthy things:
> 
> 1. Random underscores are inserted into the transcription.
> 2. The old punctuation function isn't being used, so it looks a bit janky. Quick and easy fix though, if we are sticking to the old punctuating code. I think Kate has since done some work on it, although I may be wrong.
> 3. I get 'MLVET (not responding)' at the top of the app window while vosk is transcribing. This is not something new - it was happening with the old (broken) implementation of vosk.

All expected behaviour.
Will try to add punctuation in a future PR.

src/main/handlers/helpers/transcriptionEngines/voskTranscribeFunction.ts
```diff
@@ -0,0 +1,55 @@
...
+};
+
+const transcriptionAdaptor: (wordList: VoskWord[]) => JSONTranscription = (
+  wordList
+) => ({
+  words: wordList.map((result) => ({
+    word: result.word,
+    startTime: result.start,
+    duration: result.end - result.start,
+    confidence: 0.6,
```
chloebrett: If vosk doesn't output confidence just set confidence to `1` so that there are no underlines regardless of the threshold

src/vosk/helpers.ts
```diff
@@ -0,0 +1,70 @@
...
+  // The partial sentence that have been detected until now
+  partial: string;
+}
+
+export interface Model {
+  handle: any;
+  free: () => void;
+}
+
+export interface SpeakerModel {
```
chloebrett: `export type SpeakerModel = Model`

it's the same type as `Model` so just shadow it

src/vosk/helpers.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export type Result = (SpeakerResults & RecognitionResults) | RecognitionResults;
+
+export interface PartialResults {
+  // The partial sentence that have been detected until now
+  partial: string;
+}
+
+export interface Model {
+  handle: any;
```
chloebrett, line 47: what's the actual type of this?

src/vosk/helpers.ts
```diff
@@ -0,0 +1,70 @@
...
+  free: () => void;
+}
+
+export interface SpeakerModel {
+  handle: any;
+  free: () => void;
+}
+
+export interface Recognizer {
```
chloebrett: `export interface Recognizer extends Model` then you don't need `handle` and `free`

src/vosk/index.ts
```diff
@@ -0,0 +1,34 @@
...
+
+const getVoskTranscript = async (modelPath: string, filePath: string) => {
+  // Read in the wav audio file as a buffer
+  const audioBuffer = fs.readFileSync(filePath);
+
+  // Confirm the audio file is in the correct format
+  checkWavFile(audioBuffer);
+
+  // Create the models
+  const vosky = vosk();
```
chloebrett: what is `vosky`?

src/vosk/util.ts
```diff
@@ -0,0 +1,46 @@
...
+      : '../../../assets/voskDLLs';
+
+  return path.join(__dirname, prodPath, 'lib');
+};
+
+export const getDLLDir = () => {
+  const baseDLLPath = getBaseDLLPath();
+
+  let dllDir;
+  if (os.platform() === PLATFORMS.WINDOWS) {
```
chloebrett: Could do a `Record<Platform, string>` for the paths

src/vosk/vosk.ts
```diff
@@ -0,0 +1,365 @@
...
+    /**
+     * Releases the model memory
+     *
+     * The model object is reference-counted so if some recognizer
+     * depends on this model, model might still stay alive. When
+     * last recognizer is released, model will be released too.
+     */
+    const free = () => freeRecognizer(handle);
+
+    /** Configures recognizer to output n-best results
```
chloebrett, line 203: You should definitely credit where you got this code from https://github.com/alphacep/vosk-api/blob/master/src/vosk_api.h

chloebrett (commented):
Haven't tested, just added some code comments

src/main/handlers/helpers/transcriptionEngines/voskTranscribeFunction.ts
ghost: Done! Thanks

src/vosk/helpers.ts
ghost: Good idea. Done!

src/vosk/helpers.ts
ghost: It's basically a pointer to the model used for the C++ code. As it is not referenced by the TS code, I would prefer to leave it as `any` type. I realise this is not great but the effort to open it up is not worth the code cleanliness reward.

ghost: I can add a comment describing this but I feel that looking at its usage makes it clear what it is used for

src/vosk/helpers.ts
ghost: Great catch. Done!

src/vosk/index.ts
ghost: Leftover testing code, my bad. Replaced with `voskManager`

src/vosk/util.ts
ghost: Good idea. My bad for not thinking of this. Done! (Also combined two objects into one for more cleanliness!)

src/vosk/vosk.ts
ghost: Very very good point. I added a reference following the requirements for modifications to code under an Apache 2.0 license. Please take a look.

src/vosk/util.ts
```diff
@@ -0,0 +1,55 @@
...
+      ? '../../assets/voskDLLs'
+      : '../../../assets/voskDLLs';
+
+  return path.join(__dirname, prodPath, 'lib');
+};
+
+/**
+ * Returns the path to the vosk DLL file for the current OS
+ */
+export const getDLLDir = () => {
```
chloebrett: `getDllDir` etc (no caps on acronyms)

src/vosk/util.ts
```diff
@@ -0,0 +1,55 @@
...
+import fs from 'fs';
+import { OperatingSystems } from '../sharedTypes';
+import { operatingSystemDllFilePaths } from './helpers';
+
+/**
+ * Returns the path to the vosk DLL library for all OS's
+ */
+export const getBaseDLLPath = () => {
+  // Path is different in dev than in production
+  const prodPath =
```
chloebrett: `dllPath`? 'prod' is only true if it's production but this also has a value in development env

src/vosk/util.ts
```diff
@@ -0,0 +1,55 @@
...
+ */
+export const appendPATHStr = (currentPATHStr: string, newPath: string) => {
+  return newPath + path.delimiter + currentPATHStr;
+};
+
+/**
+ * Adds the path to the vosk DLL file to the windows PATH environment variable
+ * Note: new paths take precedence over old paths
+ */
+export const updatePathWithDLLs = (dllFilePath: string) => {
```
chloebrett: and here

src/vosk/util.ts
```diff
@@ -0,0 +1,55 @@
...
+    throw new Error(`DLL could not be found at path '${dllDir}'`);
+  }
+
+  return dllDir;
+};
+
+/**
+ * Utility function for adding new paths/strings to the windows PATH environment variable
+ */
+export const appendPATHStr = (currentPATHStr: string, newPath: string) => {
```
chloebrett: `appendPathStr`, `currentPathStr`. I know there's a distinction between `path` and `PATH` but this is already explained in the docstring comment

src/vosk/util.ts
```diff
@@ -0,0 +1,55 @@
...
+ * Adds the path to the vosk DLL file to the windows PATH environment variable
+ * Note: new paths take precedence over old paths
+ */
+export const updatePathWithDLLs = (dllFilePath: string) => {
+  let currentPath = process.env.Path as string;
+
+  const dllDirectory = path.dirname(dllFilePath);
+  currentPath = appendPATHStr(currentPath, path.join(dllDirectory));
+
+  process.env.Path = currentPath;
```
chloebrett: Environment variables are uppercase by convention. Are you sure this is right?

Note: windows env vars are case insensitive. Linux and mac are case sensitive.

src/vosk/vosk.ts
```diff
@@ -0,0 +1,375 @@
...
+  );
+  const recognizerReset = libvosk.func('vosk_recognizer_reset', 'void', [
+    recognizerPointer,
+  ]);
+
+  /**
+   * Set log level for Kaldi messages
+   * @param {number} level The higher, the more verbose. 0 for infos and errors. Less than 0 for silence.
+   */
+  const setLogLevel = (level: number) => {
```
chloebrett, line 115: This function seems redundant

src/vosk/vosk.ts
chloebrett: Looks good!

chloebrett:
From the description:

> Change line 14 of ./src/main/handlers/media/transcribe.ts to be:
> [TranscriptionEngine.ASSEMBLYAI]: voskTranscribeFunction,
> Note: you will also need to import voskTranscribeFunction

Um, why not just add a new transcription engine?

chloebrett (changes requested):
Tested locally, here's what I found
- first had to allow `libvosk.dll` in the security and privacy settings on mac, this was a bit annoying but I realise there's no way around it
- transcription made the UI thread hang 🤔
- but it did transcribe very fast and it worked!!!

chloebrett:
Also need to gitignore the vosk DLLs

MutazAshhab:
It doesnt work on linux: 

https://user-images.githubusercontent.com/86769131/192077291-386cfa08-bd8d-43f7-b8aa-cd015b602464.mp4

src/vosk/vosk.ts
ghost: I understand, its not be used. It does appear to work on linux and will be useful when we try and debug linux. So, would like to leave for now.

src/vosk/util.ts
ghost: Yes, this is definitely right. I don't understand it fully (I agree I thought they were all UPPERCASE) but it correctly updates the windows PATH variable at runtime.

src/vosk/util.ts
ghost: I'm a bit confused. I understand its a bit hacky but basically just want one path for production and one for development.
This logic checks if the variable exists, which it wouldn't in a proper production app and checks if its in 'development'. I can change it to just check if the variable exists if you like.
Also,
This logic will hopefully be removed in future PR's when we download the model at runtime so it will be the same for prod and development. I am currently working on this atm.

src/vosk/util.ts
ghost: Fair point. I'll clean it up. Thanks

src/vosk/util.ts
ghost: Fixed!

src/vosk/util.ts
ghost: Fixed!

ghost:
> Also need to gitignore the vosk DLLs

I'm confused @chloebrett, they should already be ignored. They are on mine?...

ghost:
> Tested locally, here's what I found
> 
> * first had to allow `libvosk.dll` in the security and privacy settings on mac, this was a bit annoying but I realise there's no way around it
> * transcription made the UI thread hang 🤔
> * but it did transcribe very fast and it worked!!!

@chloebrett 
Hanging will be fixed in a later PR. I will add it to the clickup ticket as part of UI. Just don't want to make this too chunky. In regards to the security issue, maybe we can not download all of the different OS dlls when they get downloaded at runtime in the future.

ghost:
> From the description:
> 
> > Change line 14 of ./src/main/handlers/media/transcribe.ts to be:
> > [TranscriptionEngine.ASSEMBLYAI]: voskTranscribeFunction,
> > Note: you will also need to import voskTranscribeFunction
> 
> Um, why not just add a new transcription engine?

@chloebrett 
We are only doing this for testing to quickly and easily trick the UI without changing it in this PR. Have 100% added it as another engine in local transcription UI branch which will soon have a PR

src/vosk/vosk.ts
chloebrett: Not about it not being used, more about it exactly shadowing `setLogLevelDebug` lol

src/vosk/util.ts
chloebrett: That's because windows is case insensitive. Does it work with all caps?

src/vosk/util.ts
chloebrett: just suggested changing the variable name to `dllPath` since it's not a `prodPath` if it's not prod

chloebrett:
> Hanging will be fixed in a later PR. I will add it to the clickup ticket as part of UI. Just don't want to make this too chunky. In regards to the security issue, maybe we can not download all of the different OS dlls when they get downloaded at runtime in the future.

I don't think this would fix the problem, macOS is just a lil beatch about privacy

> We are only doing this for testing to quickly and easily trick the UI without changing it in this PR. Have 100% added it as another engine in local transcription UI branch which will soon have a PR

All good then

MutazAshhab:
Update: 

Hugh and I debugged the vosk transcription and we found out that the small model (40MB) on linux runs perfectly and can transcribe a video with no issue.

However, the big model (1.8 GB) crashes with exit code SIGTRAP.

src/vosk/util.ts
Mhudlark: Oh I see. Fixed!

src/vosk/vosk.ts
Mhudlark: Just using it to add typing and comments. I understand it is heavily shadowing

src/vosk/util.ts
Mhudlark: Yep it does. I agree, its easier to understand if its uppercase. Changed!

chloebrett (approved):

PR #295, opened 2022-09-22
Update gencode to handle dirs within handlers correctly

ghost:
## Summary

A little fix to `yarn gen` so that it successfully generates IPC code for handlers in dirs within `handlers` such as `handlers/file`.

<hr/>

Might not have been working on all OS's before due to different OS's handling path strings in different ways.

<hr/>

### What did you change?

Used `path` pkg a bit more as it handles paths across OS's

<hr/>

### OS

- [ ] Linux
- [ ] MacOS
- [x] Windows

### DIFF
--- a/tools/gencode.ts
+++ b/tools/gencode.ts
@@ -117,13 +117,15 @@ const extractHandlersMetadata: () => Promise<
   GencodeIpcHandler[]
 > = async () => {
   // Load list of file paths in handlers directory, recursively visiting subfolders
-  const fileNames = (await recursivelyReadDirectory(HANDLERS_DIRECTORY)).map(
-    (filePath: string) => filePath.split(`${HANDLERS_DIRECTORY}/`)[1]
+  const filePaths = (await recursivelyReadDirectory(HANDLERS_DIRECTORY)).map(
+    (filePath: string) => {
+      return path.resolve(filePath);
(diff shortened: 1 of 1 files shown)


tools/gencode.ts
```diff
@@ -117,13 +117,15 @@ const extractHandlersMetadata: () => Promise<
   GencodeIpcHandler[]
 > = async () => {
   // Load list of file paths in handlers directory, recursively visiting subfolders
-  const fileNames = (await recursivelyReadDirectory(HANDLERS_DIRECTORY)).map(
-    (filePath: string) => filePath.split(`${HANDLERS_DIRECTORY}/`)[1]
+  const filePaths = (await recursivelyReadDirectory(HANDLERS_DIRECTORY)).map(
+    (filePath: string) => {
+      return path.resolve(filePath);
```
chloebrett: nitpick but can inline this `.map((filePath: string) => path.resolve(filePath))` or even (assuming arguments align) point-free notation style `.map(path.resolve)`

chloebrett (approved):

tools/gencode.ts
ghost: Weirdly arguments don't align but I have reduced the number of lines. Thanks for pointing it out

PR #296, opened 2022-09-22
Take Detection - Ungroup Takes and UI Changes

exquisitebin:
## Summary
Allows takes to be ungrouped by pressing the block icon and confirming in a modal. Take text is aligned with the rest of the transcript. Words in takes correctly wrap to the width of the transcript box. The brackets change height dynamically.

Update 1:
When the user selects a take, the fact that the selection was made is stored in the project file. This is so that the yellow brackets don't open up again when the user reopens the project after saving. This was done by changing the TakeGroup shared type (old files without the extra attribute should still open fine).

Refactored/changed the way TranscriptionChunks are generated to allow for deleted take groups.

Update 2:
Take groups close on click away.

https://user-images.githubusercontent.com/58348527/192077933-6f782a15-2ef3-4bb6-a885-618d2948f09e.mp4

<hr/>

### OS

- [ ] Linux
- [ ] MacOS
- [x] Windows


src/renderer/components/Editor/SquareBracket.tsx
```diff
@@ -6,9 +6,10 @@ import SquareBracketHover from './SquareBracketHover';
 interface Props {
   isLast: boolean;
   isTakeGroupOpened: boolean;
+  takeHeight: number;
```
chloebrett: Make it clear that this is in `px`. Kind of hacky that we have to do this in JS in the first place but we have other JS things that should be CSS so I'll give it a pass

src/renderer/components/Editor/TakeComponent.tsx
```diff
@@ -111,68 +121,71 @@ const TakeComponent = ({
...
+        onMouseDown={onWordMouseDown}
+        onMouseEnter={onWordMouseEnter}
+        isInInactiveTake={!(isActive || isTakeGroupOpened) && !isFirstTimeOpen}
+        transcriptionLength={words.length}
+        {...passThroughProps}
+      />
+    );
+  });
+
+  useLayoutEffect(() => {
```
chloebrett, line 167: What happens if this is a regular `useEffect`? Is there a flicker?

src/renderer/components/Editor/TakeComponent.tsx
```diff
@@ -111,68 +121,71 @@ const TakeComponent = ({
...
-                    position: 'absolute',
-                    left: '-30px',
-                    transform: 'translateY(2px)',
-                    cursor: 'pointer',
-                  }}
-                >
-                  {takeIndex + 1}
-                </Avatar>
-              )}
+              {!isFirstTimeOpen && isTakeGroupOpened && NumberedButton}
```
chloebrett: This is a bit hard to read - maybe put the conditional in the declaration for `numberedButton`. I.e.

```
const numberedButton = (!isFirstTimeOpen && isTakeGroupOpened) ? (... stuff ...) : null;
```

And then just render it unconditionally.

Also `numberedButton` should start with a lowercase letter - the only things that should be in `PascalCase` are class names and component declarations (e.g. you declare a component `Thing` which can be built as `<Thing />`, this is not what `NumberedButton` is as we can't go `<NumberedButton />`)

src/renderer/components/Editor/TakeGroupComponent.tsx
```diff
@@ -118,12 +153,36 @@ const TakeGroupComponent = ({
...
+            </CustomRowStack>
+          </CustomColumnStack>
+          <UngroupTakesModal
+            isOpen={showUngroupModal}
+            closeModal={handleModalClose}
+            ungroupTakes={ungroupTakes}
+          />
+        </>
+      ) : (
+        <></>
```
chloebrett: `null` is cleaner than rendering an empty fragment

src/renderer/components/Editor/TakeGroupComponent.tsx
```diff
@@ -118,12 +153,36 @@ const TakeGroupComponent = ({
   });
 
   return (
-    <CustomColumnStack sx={{ marginTop: '10px', marginBottom: '10px' }}>
-      <CustomRowStack sx={{ justifyContent: 'flex-end' }}>
-        {!isFirstTimeOpen && isTakeGroupOpened && <BlockIcon />}
-      </CustomRowStack>
-      {takes}
-    </CustomColumnStack>
+    <>
```
chloebrett: Don't need this - fragments only needed when rendering multiple children at the top level

src/renderer/components/Editor/TakeGroupComponent.tsx
```diff
@@ -118,12 +153,36 @@ const TakeGroupComponent = ({
...
+      {!takeDissolved ? (
```
chloebrett: Flip this conditional - in fact could just do an early return i.e.

```
if (takeDissolved) {
  return null;
}

return ...
```

src/renderer/components/Editor/TakeGroupComponent.tsx
```diff
@@ -55,8 +73,25 @@ const TakeGroupComponent = ({
   transcription,
   ...passThroughProps
 }: TakeGroupComponentProps) => {
+  const [takeDissolved, setTakeDissolved] = useState(false);
```
chloebrett: Do we actually need this? Isn't `takeDissolved` implied by the take group being deleted, and then this wouldn't be rendered any more?

src/renderer/components/Editor/UngroupTakesModal.tsx
```diff
@@ -0,0 +1,77 @@
...
+  return (
+    <div>
+      <CustomModal open={isOpen} onClose={closeModal}>
+        <CustomModalInner>
+          <CustomStack>
+            <Typography variant="h1" color={colors.grey[400]}>
+              Ungroup Takes
+            </Typography>
+            <Typography variant="p-300" paddingTop="10px">
+              Are you sure you want to ungroup takes? This is not reversible.
```
chloebrett: Why shouldn't it be reversible? We can easily bake it into undo/redo. In fact if we don't, it kind of breaks the whole undo/redo system - you can't really have "some stuff is reversible and other things aren't". Even if the only way to get the takes back is by undoing, I think this message is misleading

chloebrett (changes requested):

src/renderer/components/Editor/TakeComponent.tsx
exquisitebin: I think it looks a little more smoother with useLayoutEffect but they're both not perfect

src/renderer/components/Editor/TakeComponent.tsx
exquisitebin: Yeah I wasn't sure about these conventions. Thanks for clearing them up.

src/renderer/components/Editor/TakeGroupComponent.tsx
exquisitebin: Yeah originally the components were still being rendered and I didn't know why. I've had a further look and found that the takeDetection.ts generated TakeGroup TranscriptionChunks under the assumption that take group ids always start from 0 and increase by 1 (which obviously doesn't support deleted take groups). I have just fixed it in this last commit and have also removed the useState.

src/renderer/components/Editor/UngroupTakesModal.tsx
exquisitebin: The original discussion when producing these UI designs for take detection was that it might be too difficult to undo the ungrouping of takes. Idk personally if that's the case. @emmapainter thoughts?

src/renderer/components/Editor/TakeGroupComponent.tsx
chloebrett: nice catch

src/renderer/components/Editor/UngroupTakesModal.tsx
chloebrett: Nah it's not difficult at all - destructive actions are perfectly allowed within the undo system, just pass enough info into the op so that the "undo" action can reconstruct whatever was destroyed. Although I guess it also won't *really* break undo if we don't make it undoable since take state is kind of orthogonal to the rest of the editor state, but the invariant of "should be able to restore a project to its original state" will be gone

package.json
```diff
@@ -274,10 +274,12 @@
     "history": "^5.3.0",
     "node-fetch": "2.6.6",
     "query-string": "^7.1.1",
+    "randomatic": "^3.1.1",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
     "react-redux": "^7.2.6",
     "redux": "^4.1.2",
+    "socket.io": "^4.5.2",
```
chloebrett: I think this is a merge problem, this shouldn't be a requirement for this change lol. And neither should randomatic (they're both collab server dependencies, not electron client dependencies)

src/renderer/store/transcription/reducer.ts
```diff
@@ -68,6 +68,7 @@ const transcriptionReducer: Reducer<Transcription | null, Action<any>> = (
   if ([SELECT_TAKE, DELETE_TAKE_GROUP].includes(action.type)) {
     // Update take groups first, so that updateOutputTimes uses the correct take groups
     const takeGroups = takeGroupsReducer(transcription.takeGroups, action);
+    console.log(takeGroups);
```
chloebrett: remove

src/renderer/utils/takeDetection.ts
```diff
@@ -19,45 +19,32 @@ export const generateTranscriptionChunks = (
...
-          id: 0,
+          id: takeInfo.takeGroupId,
           activeTakeIndex:
-            takeGroups.find((takeGroup) => takeGroup.id === 0)
-              ?.activeTakeIndex ?? 0,
+            takeGroups.find(
+              (takeGroup) => takeGroup.id === takeInfo.takeGroupId
+            )?.activeTakeIndex ?? 0,
+          takeSelected:
+            takeGroups.find(
```
chloebrett: repeated code, can extract this out as `matchedTakeGroup` or something

chloebrett (approved):

chloebrett:
Resolve the above and then it's all good, tested locally no issues

package.json
exquisitebin: I've just reverted that package.json/yarn commit. 

I've merged develop into my branch again, but I'm still having these local test cases fail due to those dependencies. I know it's got nothing to do with my branch, just thought it might be an easy fix.

image

chloebrett: `cd src/collabServer && yarn` should fix it

src/renderer/components/Editor/UngroupTakesModal.tsx
emmapainter: If it's not difficult then I'd love for it to be reversible - thanks for pulling this up @chloebrett, probably something we should have investigated before writing it off as too difficult. Make sure you change the modal text to align with this too.

emmapainter (changes requested):
Have a go at making ungroup takes reversible, and make sure you also change the text in the modal to explain what's going on - happy to work with you on copy for this if you're not sure. Love the changes in this PR, take detection is looking slick now 😍

chloebrett:
> Have a go at making ungroup takes reversible

I reckon do that in a separate PR, I need this merged so I can do the transcription block styling updates

src/renderer/components/Editor/UngroupTakesModal.tsx
exquisitebin: Sounds good, I've just started trying to implement this. Should I tack it on to this PR or to a new one?

exquisitebin: Lol I just refreshed the page and saw Emma and your comments

exquisitebin:
> Have a go at making ungroup takes reversible, and make sure you also change the text in the modal to explain what's going on - happy to work with you on copy for this if you're not sure. Love the changes in this PR, take detection is looking slick now 😍

Have started working on this. Probably easier to make it a separate PR so that Chloe can use the current code.

emmapainter (approved):

PR #298, opened 2022-09-22
Feature/transcription buffer before 02s

tngu0041:
## Summary
Adding a 0.2s silence buffer before Words based on discussion from https://github.com/chloebrett/mlvet/pull/272

Also changed the test to handle the case where available buffer size is smaller than 0.2s.

And also had to run the results from the Inject Pause logic through the roundToMs() function to resolve some floating point rounding issues.
<hr/>

### OS

- [ ] Linux
- [ ] MacOS
- [x] Windows


src/main/editDelete/preProcess.ts
```diff
@@ -28,35 +28,66 @@ const injectAttributes: MapCallback<PartialWord, Word> = (word, index) => ({
...
 
   if (nextWord === null) {
-    return totalDuration - wordEndTime;
+    return roundToMs(totalDuration - wordEndTime);
   }
   const gapTime = nextWord.startTime - wordEndTime;
 
-  return roundToMs(gapTime);
+  if (gapTime <= bufferBeforePadding) {
+    return roundToMs(0);
```
chloebrett: There won't be floating point errors with this int literal, you can just `return 0;`

src/main/editDelete/preProcess.ts
```diff
@@ -28,35 +28,66 @@ const injectAttributes: MapCallback<PartialWord, Word> = (word, index) => ({
...
+  prevWord: Word | null,
+  bufferBeforePadding: number
+) => number = (word, prevWord, bufferBeforePadding) => {
+  if (prevWord === null) {
+    return word.startTime;
+  }
+  const prevWordEndTime = prevWord.startTime + prevWord.duration;
+  const gapTime = word.startTime - prevWordEndTime;
+
+  // Half the gap used by this word, half by the previous word.
```
chloebrett: This comment is not accurate anymore

chloebrett (changes requested):

Irmy (approved):

src/main/editDelete/__tests__/preProcess.test.ts
```diff
@@ -78,9 +79,11 @@ describe('Test pre-processing JSON transcript into regular transcript', () => {
       defaultThreshold
     );
```
chloebrett, line 81: Remove the console.log

chloebrett (approved):

PR #300, opened 2022-09-23
Feature/Autosave

itsMoxMox:
## Summary

Autosave.

Saves after every action which modifies the project file during runtime.
Hopefully this will save some ~lives~ projects.

<hr/>

### OS

- [ ] Linux
- [ ] MacOS
- [X] Windows

### DIFF
--- a/src/renderer/StoreChangeObserver.tsx
+++ b/src/renderer/StoreChangeObserver.tsx
@@ -48,6 +48,8 @@ const StoreChangeObserver = () => {
     (store: ApplicationStore) => store.editWord?.index
   );
 
+  const collab = useSelector((store: ApplicationStore) => store.collab);
+
   // Debounce the selection to limit network requests for sharing selection with other clients
   const [debouncedSelection, setDebouncedSelection] =
     useDebounce(selfSelection);
@@ -195,10 +197,13 @@ const StoreChangeObserver = () => {
       currentProject.transcription &&
       currentProject.projectFilePath
     ) {
+      if (collab !== null && collab.isHost) return; // Disallow autosave if you are not the host in collab
+
       ipc.saveProject(currentProject);
-      ipc.setFileRepresentation(currentProject.projectFilePath, false); // Don't show a Save Prompt Dialog when closing.
+
+      currentProject.isEdited = false; // set the dirtiness to clean!
     }
-  }, [currentProject]);
+  }, [currentProject, collab]);
 
   // Component doesn't render anything
   return null;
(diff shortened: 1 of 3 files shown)


chloebrett (approved):
Havent tested locally but code looks fine, as long as it doesn’t do any weird infinite loops with the isDirty thing then looks good to me. Get someone else to test locally before merge

chloebrett:
Actually there is one more check to add to make this work with collab.

Load collab off the store if it’s not being loaded already (I think it is) and then:

`if (collab !== null && !collab.isHost)` then early exit. We don’t want to auto save if the client is a guest on someone else’s project

fwang1703 (commented):
Code looks good. Tested it on my Mac, and the autosaving was happening at each action, but when I tried to return to home, it still asked me if I wanted to save (even though autosaving has occurred). Is this expected behaviour?

chloebrett:
Yeah @moxchan should probably update the project file dirtiness before saving, just be careful to avoid an infinite loop with this (should be relatively easily to prevent)

src/renderer/StoreChangeObserver.tsx
```diff
@@ -188,6 +190,21 @@ const StoreChangeObserver = () => {
...
   }, [isShowingConfidenceUnderlines, words]);
 
+  // Autosave after currentProject changes.
+  useEffect(() => {
+    if (
+      currentProject &&
+      currentProject.transcription &&
+      currentProject.projectFilePath
+    ) {
+      if (collab !== null && collab.isHost) return; // Disallow autosave if you are not the host in collab
```
chloebrett: `!collab.isHost`, the way you've written it, it will early exit if they _are_ the host

src/renderer/StoreChangeObserver.tsx
```diff
@@ -188,6 +190,21 @@ const StoreChangeObserver = () => {
...
+
+      ipc.saveProject(currentProject);
+
+      currentProject.isEdited = false; // set the dirtiness to clean!
```
chloebrett: You need to dispatch this, if you just edit it it won't update the store (or at the very least, if it does work, it shouldn't and will cause unexpected side effects since redux is meant to be immutable)

You can do this by dispatching a `projectSaved` action. Just be aware this will cause an infinite loop so check that `currentProject.isEdited` is `true` here, otherwise early exit

chloebrett (changes requested):

src/renderer/StoreChangeObserver.tsx
moxchan: Will do

moxchan: Just realised that I was using `saveProject` from ipc, which kind of requires repeated code, at least when comparing to `saveProject` from `renderer/file`, im not sure how jank it is to import that function instead of rewriting some logic to dispatch a projectSaved

chloebrett (approved):
Looks good!

PR #308, opened 2022-09-25
Feature/restore video preview redux

MutazAshhab:
## Summary

When the user clicks on a deleted take it should then play the deleted take on the video preview with the following conditions:
1. Scrubber must play the deleted take (meaning 0 is the start of the take, and the right most section of the scrubber is the end of the deleted take)
2. Word highlight must not be displayed.

<hr/>

Yes, Andys #274 
<hr/>

### Demo recording

https://user-images.githubusercontent.com/86769131/192126803-7108aa7d-4545-4217-92fe-21f265bcd56d.mp4

<hr/>

### OS

- [x] Linux
- [ ] MacOS
- [ ] Windows

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,5 +38,3 @@ assets/thumbnails/thumbnail.png
 experiments/**/assets
 
 
-voskDLLs
-voskModel
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,7 @@ assets/audio/audio.wav
 assets/thumbnails/thumbnail.png
 
 experiments/**/assets
+
+
+voskDLLs
+voskModel
--- a/src/renderer/components/Editor/PlaybackManager.tsx
+++ b/src/renderer/components/Editor/PlaybackManager.tsx
@@ -8,6 +8,8 @@ import React, {
   useMemo,
   useState,
 } from 'react';
+import { useSelector } from 'react-redux';
+import { ApplicationStore } from 'renderer/store/sharedHelpers';
 import { RuntimeProject } from 'sharedTypes';
 import { bufferedWordDuration, isInInactiveTake } from 'sharedUtils';
 import { VideoPreviewControllerRef } from '../VideoPreview/VideoPreviewController';
@@ -41,6 +43,10 @@ const PlaybackManager = ({
   const [isPlaying, setIsPlaying] = useState(false);
   const [nowPlayingWordIndex, setNowPlayingWordIndex] = useState<number>(0);
 
+  const rangeOverride = useSelector(
+    (store: ApplicationStore) => store.playback.rangeOverride
+  );
+
   const play = useCallback(
     () => videoPreviewControllerRef?.current?.play(),
     [videoPreviewControllerRef]
@@ -73,6 +79,11 @@ const PlaybackManager = ({
       return;
     }
 
+    if (rangeOverride !== null) {
+      setNowPlayingWordIndex(-1);
+      return;
+    }
+
     const newPlayingWordIndex = currentProject.transcription.words.findIndex(
       (word) =>
         time >= word.outputStartTime &&
@@ -85,7 +96,12 @@ const PlaybackManager = ({
       setNowPlayingWordIndex(newPlayingWordIndex);
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [time, currentProject?.transcription]);
+  }, [
+    time,
+    currentProject?.transcription,
+    rangeOverride,
+    setNowPlayingWordIndex,
+  ]);
 
   return useMemo(
     () =>
--- a/src/renderer/components/Editor/PlaybackManager.tsx
+++ b/src/renderer/components/Editor/PlaybackManager.tsx
@@ -79,6 +79,8 @@ const PlaybackManager = ({
       return;
     }
 
+    // When rangeOverride is not null that means we are playing a deleted take (by clicking on the edit marker)
+    // then there should be no highlighted word, thus nowPlayingWordIndex is set to -1
     if (rangeOverride !== null) {
       setNowPlayingWordIndex(-1);
       return;
--- a/src/renderer/store/playback/reducer.ts
+++ b/src/renderer/store/playback/reducer.ts
@@ -20,37 +20,35 @@ const playbackReducer: Reducer<ApplicationStore['playback'], Action<any>> = (
 ) => {
   if (action.type === VIDEO_PLAYING) {
     const playState = action.payload as UpdatedPlaying;
+    const { isPlaying, lastUpdated } = playState;
+
     return {
-      isPlaying: playState.isPlaying,
-      time: playback.time,
-      lastUpdated: playState.lastUpdated,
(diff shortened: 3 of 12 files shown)


.gitignore
```diff
@@ -36,3 +36,7 @@ assets/audio/audio.wav
 assets/thumbnails/thumbnail.png
 
 experiments/**/assets
+
+
+voskDLLs
```
Mhudlark: Lol please remove these

src/renderer/store/playback/reducer.ts
```diff
@@ -21,6 +24,7 @@ const playbackReducer: Reducer<ApplicationStore['playback'], Action<any>> = (
       isPlaying: playState.isPlaying,
       time: playback.time,
       lastUpdated: playState.lastUpdated,
+      rangeOverride: playback.rangeOverride,
```
Mhudlark: Can just use object destructuring (spreading) of previous object state rather than setting individual properties based on previous state.
eg:
image

src/renderer/store/playback/reducer.ts
```diff
@@ -60,10 +66,27 @@ const playbackReducer: Reducer<ApplicationStore['playback'], Action<any>> = (
...
+
+    return {
+      ...playback,
+      rangeOverride,
+    };
+  }
+
+  if (action.type === CLEAR_RANGE_OVERRIDE) {
+    return {
+      ...playback,
```
Mhudlark, line 78: lol you did it here

src/transcriptProcessing/transcriptToCuts.ts
```diff
@@ -1,16 +1,27 @@
...
+): Array<Cut> => {
+  const wordsNotYetUpdated =
+    // filter out deleted words as well as words in inactive takes
+    rangeOverride === null
+      ? transcript.words.filter(
+          (word) =>
+            !word.deleted && !isInInactiveTake(word, transcript.takeGroups)
+        )
+      : // keep deleted words when rangeOverride is specified
+        transcript.words.slice(
```
Mhudlark, line 21: ah cool so we can use rangeOverride for take detection too if needed. Nice!

src/renderer/components/Editor/PlaybackManager.tsx
```diff
@@ -73,6 +79,11 @@ const PlaybackManager = ({
       return;
     }
 
+    if (rangeOverride !== null) {
```
Mhudlark, line 84: What does this logic do? Maybe add a comment for dumbos like me pls 😃

Mhudlark:
Love your work! Honestly great job figuring out video preview within the 0 comments I left for future developers (sorry 🤣 ).
I noticed an issue to do with seeking within a restore preview. Seems like if I seek it takes me out of the preview.
Please confirm that all standard transcription controls work while in the video preview.

src/renderer/components/Editor/PlaybackManager.tsx
chloebrett: Logic is, if we're playing from a deleted section instead of the main transcription (which is determined by rangeOverride !== null), then there should be no highlighted word so the nowPlayingWordIndex is set to -1

src/renderer/store/playback/reducer.ts
chloebrett: Yeah in this case something like

```
const { isPlaying, lastUpdated } = playState;

return {
  ...playback,
  isPlaying,
  lastUpdated,
}
```

would work. I think still be explicit about the things being updated rather than just going `{ ...playback, ...playState }`

src/renderer/store/playback/reducer.ts
chloebrett: yeah that's because this is a new action, he was just following what the previous author did for the other actions

src/transcriptProcessing/transcriptToCuts.ts
chloebrett: wdym by this? currently only used for previewing deleted sections but yeah it was architected so that it's flexible-ish

src/renderer/components/Editor/PlaybackManager.tsx
chloebrett: would rather it be `null` instead of `-1` but theres quite a few places that assume it's strictly a number

src/renderer/store/playback/reducer.ts
MutazAshhab: thanks Chloe for tagging along, I've tidied the reducers up and all of them now follow the same style you suggested. Here is the commit: [491b0c3](https://github.com/chloebrett/mlvet/pull/308/commits/491b0c33fb0b86a7d69062567a28691f293f0b1e)

src/renderer/components/Editor/PlaybackManager.tsx
MutazAshhab: Commit that adds the comment: [a5fbed8](https://github.com/chloebrett/mlvet/pull/308/commits/a5fbed804fd6a3bda0362a781d0ae51f3272f17b)

And yeah, basically what Chloe said

.gitignore
MutazAshhab: done: [1b20e83](https://github.com/chloebrett/mlvet/pull/308/commits/1b20e830e0b0e951b31bcc92cfa833b16b2392ad)

src/renderer/store/playback/reducer.ts
MutazAshhab: Chloe has my back :muscle: beat that @Mhudlark

src/transcriptProcessing/transcriptToCuts.ts
Mhudlark: the plan later down the track for take detection is to have preview of the take when selecting takes. This could be used for that

chloebrett: Yep so that would be able to reuse like half this code - basically video preview now supports either previewing the entire video or a fixed range of words

chloebrett: Yep so that would be able to reuse like half this code - basically video preview now supports either previewing the entire video or a fixed range of words

Irmy (changes requested):
Absolute beast, you smashed this! 
I think the restart playback issue when at the end of the snippet is left over form PR #274. Noticed if I replay the deleted section again by pressing space twice, with each replay a bit more of the video replay gets cut off at the end. 

Had the same issue with the scrubber that Hugh mentioned where it exits the restore text element and plays back the transciption. Maybe if we limit the cancel restore pop out area to the transcription? Happy to have your input here. Once the arrow shortcuts for playback are in that should make the user experience better too

Irmy:
Also had this error - I am on my bricked laptop so let me know if you cant replicate this error

https://user-images.githubusercontent.com/55219708/192187654-4b526f2b-3d9b-4e3f-9b44-962e2c7fb3f1.mp4

chloebrett:
Guess this can't be merged until we manage to fix what https://github.com/chloebrett/mlvet/pull/323 reverted

PR #309, opened 2022-09-25
Undo delete take group

exquisitebin:
## Summary

Allows users to undo the deletion of a take group using the undo stack.

<hr/>

So that users can undo/redo the deletion of a take group.

<hr/>

https://user-images.githubusercontent.com/58348527/192132371-d10b05c4-45a3-44e2-a517-aba769eb41d5.mp4

### OS

- [ ] Linux
- [ ] MacOS
- [x] Windows

### DIFF
--- a/src/renderer/components/Editor/TakeGroupComponent.tsx
+++ b/src/renderer/components/Editor/TakeGroupComponent.tsx
@@ -92,31 +92,56 @@ const TakeGroupComponent = ({
   const ungroupTakes = () => {
     const getTakeRanges = (): IndexRange[] => {
       const { words } = transcription;
-      const takeRanges = [] as IndexRange[];
-      let start = -1;
-      let end = -1;
-      let takeIndex = 0;
-      let takeInfo;
-      for (let i = 0; i < words.length; i += 1) {
-        takeInfo = words[i].takeInfo;
-        if (takeInfo !== null && takeInfo?.takeGroupId === takeGroup.id) {
-          if (start === -1) {
-            start = i;
-          } else if (takeInfo.takeIndex !== takeIndex) {
-            end = i;
-            takeRanges.push({
-              startIndex: start,
-              endIndex: end,
-            } as IndexRange);
-            takeIndex += 1;
-            start = i;
+
+      // Make a list of only the words in this take group,
+      // keeping track of what their indices in the current transcription were
+      const wordsInTakeGroup: {
+        word: Word;
+        indexInTranscription: number;
+      }[] = words
+        // Keep track of the indices in the transcription so that we can use them later
+        .map((word, i) => ({ word, indexInTranscription: i }))
+        // Filter to only include words in the current take group
+        .filter(({ word }) => word.takeInfo?.takeGroupId === takeGroup.id);
+
+      // Extract a list of take ranges from the word list we just made
+      const takeRanges = wordsInTakeGroup.reduce<IndexRange[]>(
+        (rangesSoFar, word, index) => {
+          const isLastWordInTakeGroup = index === wordsInTakeGroup.length - 1;
+          const nextWord = isLastWordInTakeGroup
+            ? null
+            : wordsInTakeGroup[index + 1];
+          const isLastWordInTake =
+            nextWord?.word.takeInfo?.takeIndex !==
+            word.word.takeInfo?.takeIndex;
+
+          // If we're in a new take compared to the previous word, or this is the last
+          // word in the take group, 'push' the current take to the list of takes.
+          if (isLastWordInTakeGroup || isLastWordInTake) {
+            // Find what the startIndex of the take should be. Since takes are always
+            // next to each other, the start index will just be the endIndex of the previous
+            // take range, or if this is the first take in the group, it will be the index of
+            // the first word in the take.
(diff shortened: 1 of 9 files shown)


src/renderer/components/Editor/TakeGroupComponent.tsx
```diff
@@ -91,7 +90,38 @@ const TakeGroupComponent = ({
   };
 
   const ungroupTakes = () => {
-    dispatch(deleteTakeGroup(takeGroup.id));
+    const getTakeRanges = (): IndexRange[] => {
```
chloebrett, line 93: This should definitely be a util function instead of being inline in the react component.

I tried to refactor it a little to make it easier to read, because I was struggling to understand the logic. In order to minimise usage of mutating variables (which tends to be the biggest reason functions get hard to follow), I ended up rewriting it entirely as a `reduce`:

```
const getTakeRanges: () => IndexRange[] = () => {
  const { words } = transcription;

  // Make a list of only the words in this take group,
  // keeping track of what their indices in the current transcription were
  const wordsInTakeGroup: {
    word: Word;
    indexInTranscription: number;
  }[] = words
    // Keep track of the indices in the transcription so that we can use them later
    .map((word, i) => ({ word, indexInTranscription: i }))
    // Filter to only include words in the current take group
    .filter(({ word }) => word.takeInfo?.takeGroupId === takeGroup.id);

  // Extract a list of take ranges from the word list we just made
  const takeRanges = wordsInTakeGroup.reduce<IndexRange[]>(
    (rangesSoFar, word, index) => {
      const isLastWordInTakeGroup = index === wordsInTakeGroup.length - 1;
      const nextWord = isLastWordInTakeGroup ? null : wordsInTakeGroup[index + 1];
      const isLastWordInTake =
        nextWord?.word.takeInfo?.takeIndex !==
        word.word.takeInfo?.takeIndex;

      // If we're in a new take compared to the previous word, or this is the last
      // word in the take group, 'push' the current take to the list of takes.
      if (isLastWordInTakeGroup || isLastWordInTake) {
        // Find what the startIndex of the take should be. Since takes are always
        // next to each other, the start index will just be the endIndex of the previous
        // take range, or if this is the first take in the group, it will be the index of
        // the first word in the take.
        const startIndex =
          rangesSoFar.length > 0
            ? rangesSoFar[rangesSoFar.length - 1].endIndex
            : wordsInTakeGroup[0].indexInTranscription;

        // Add a new range from startIndex up to the current word's index in the current
        // transcription, plus one as the end index is exclusive
        return rangesSoFar.concat([
          {
            startIndex,
            endIndex: word.indexInTranscription + 1,
          },
        ]);
      }

      // If this isn't the last word or a new take, just return the list of ranges unmodified
      return rangesSoFar;
    },
    []
  );

  return takeRanges;
};
```

I think this is a lot easier to reason about, even if it looks more verbose from the comments. Feel free to use this or to take the ideas and include them in your code. Also note that regardless of what the final function ends up being, this is exactly the kind of thing that needs unit tests so please write some. I've tested this with Max's sauce video but haven't tested three-take groups or anything more complex.

src/renderer/store/takeGroups/reducer.ts
```diff
@@ -27,6 +32,12 @@ const takeGroupsReducer: Reducer<TakeGroup[], Action<any>> = (
     return takeGroups.filter((takeGroup) => takeGroup.id !== takeGroupId);
   }
 
+  if (action.type === UNDO_DELETE_TAKE_GROUP) {
+    const { takeGroup } = action.payload as UndoDeleteTakeGroupPayload;
+
+    return takeGroups.concat(takeGroup).sort((a, b) => a.id - b.id);
```
chloebrett, line 38: Nice

src/renderer/store/transcriptionTakes/reducer.ts
```diff
@@ -21,6 +29,23 @@ const transcriptionTakesReducer: Reducer<Word[], Action<any>> = (
...
+    const updateWordTakeInfo = (takeIndex: number) => (word: Word) => ({
+      ...word,
+      takeInfo: {
+        takeGroupId: takeGroup.id,
+        takeIndex,
+      },
+    });
+
+    return takeRanges.reduce((wordsAcc, currRange, currIndex) => {
+      return mapInRange(wordsAcc, updateWordTakeInfo(currIndex), currRange);
```
chloebrett, line 45: Very cool, love the currying

chloebrett (changes requested):

src/renderer/components/Editor/TakeGroupComponent.tsx
exquisitebin: Yup that makes sense. Thanks for translating it into a more functional programming style, I wasn't sure how necessary it was so I just caved into imperative rip. I'll make sure to have comments next time at least.

I'll make some test cases up next, definitely a good idea.

emmapainter (approved):

MutazAshhab (approved):
Looks really good, just the comments that Chloe already mentioned. :+1:

chloebrett (approved):
You can merge this if you want, but test cases are never a bad thing. Realise we have like 10 days left anyway

PR #310, opened 2022-09-25
Fix dialog cancel error

zwan0202:
## Summary

Added error handling to `export`, `save`, and `saveAs` dialog cancelation. 

To eliminate the error screen when canceling a dialog.

<hr/>

### What did you change?

When canceling a dialog it will just return and skip other calls. 

Not sure if this is the best way to handle the dialog cancelation.

<hr/>

### OS

- [x] Linux
- [ ] MacOS
- [x] Windows


chloebrett (approved):
Seems fine

exquisitebin (approved):
Looks like its all fixed :+1:

PR #314, opened 2022-09-27
Fix reportBug handler declaration and run gencode

Mhudlark:
## Summary

Just a little fix to a handler function declaration so that `yarn gen` works again.

<hr/>

### What did you change?

Added a separate type declaration according to the following error given when `yarn gen` is run.

`Error: Error extracting type from handler in file reportBug.ts . Handler type must be declared separately from handler, with matching name in PascalCase. See existing handlers for examples.`


LiamTodd (approved):
Thanks for fixing this up!

PR #317, opened 2022-09-28
Feature/local transcription UI

### LINKED ISSUE #315: Multiple vosk transcriptions throws error
Vosk throws an error when trying to transcribe multiple videos in the same session.

image

### LINKED ISSUE #351: Download model sits at 100% for 20 seconds before saying 'Download complete'

Mhudlark:
## Summary

Adds the local transcription UI

**NOTE!!!!:**
When testing, it will be a lot faster to download the model if you change `line 62` of `downloadModel.ts` to be `return MODEL_SML_URL;`.

<hr/>

### What did you change?

Features added:
- User can choose between local transcription and cloud transcription during project creation
- User can switch between local transcription and cloud transcription from the menu
- If requirements are not met (assets not downloaded or API key not entered) for a transcription engine then user will be required to remedy this (download assets or enter key)

Bugs fixed:
- Fixed bug where an error was thrown if user tried to transcribe multiple times locally in the same session

<hr/>

<hr/>

### OS

- [ ] Linux
- [ ] MacOS
- [x] Windows


rileykeane (approved):
hot. 

A few observations. but if these decide to be done, they can be a separate PR 
- There is no way to cancel the download. Clicking the **x** closes the modal, but if I create a new project, it goes straight back into the download (it does keep the download progress, nice!) 
- Since Vosk does not offer punctuation, do we still want to keep the old approach where we add commas and full stops based on gaps? My understanding take detection won't work without punctuation. If this is true, and we won't have punctuation for Vosk in the near future, we should note this somewhere in the flow. 
- The designs do have remaining time on the download modal, but I'll let this one side. We could add a note saying '_This may take 5 -15 minutes depending on your internet connection_'. But I'm also mindful that the modal is already pretty full of text.

Mhudlark:
> hot.
> 
> A few observations. but if these decide to be done, they can be a separate PR
> 
> * There is no way to cancel the download. Clicking the **x** closes the modal, but if I create a new project, it goes straight back into the download (it does keep the download progress, nice!)
> * Since Vosk does not offer punctuation, do we still want to keep the old approach where we add commas and full stops based on gaps? My understanding take detection won't work without punctuation. If this is true, and we won't have punctuation for Vosk in the near future, we should note this somewhere in the flow.
> * The designs do have remaining time on the download modal, but I'll let this one side. We could add a note saying '_This may take 5 -15 minutes depending on your internet connection_'. But I'm also mindful that the modal is already pretty full of text.

Cancelling download is a fair point that I glossed over. Proper handling should cancel the download and progress should be lost. Whoops. I have created [an issue](https://github.com/chloebrett/mlvet/issues/325) to address later.

In regards to punctuation, I'm not sure we have to mention this. It feels weird to mention technical implementation in the app. Maybe in the help docs.

rileykeane:
> In regards to punctuation, I'm not sure we have to mention this. It feels weird to mention technical implementation in the app. Maybe in the help docs.

Fair. The "_What does this mean?_" link could explain that with local, take detection is unavailable. 

This reminds me, @Irmy we need to create a doc for explaining cloud vs local and update the link in the app.

emmapainter:
> In regards to punctuation, I'm not sure we have to mention this. It feels weird to mention technical implementation in the app. Maybe in the help docs.

I mean you wouldn't mention technical implementation, you could just have a little note that says something like "Take detection is unavailable with local transcription"

Mhudlark:
> > In regards to punctuation, I'm not sure we have to mention this. It feels weird to mention technical implementation in the app. Maybe in the help docs.
> 
> I mean you wouldn't mention technical implementation, you could just have a little note that says something like "Take detection is unavailable with local transcription"

It will still be available

PR #323, opened 2022-09-30
revert: #274 play back redux

MutazAshhab:
## What is this PR? 

This undoes what was introduced in `feature/play-pause-redux`.

## Why?

1. While the code introduced in `feature/play-pause-redux` cleaned up the codebase it created an error with video preview. I've attached a recording of the bug that it introduced.
2. It also was doing multiple plays/pauses one after the other.

## Video/Demo of bug

Watch how after counting to 5 then clicking play (twice, because once doesnt work)... The video starts previewing seconds into the transcript and not and 0 seconds.

https://user-images.githubusercontent.com/86769131/193235113-ecdb882e-da5a-464a-b0ad-09735d98832d.mp4


Mhudlark (approved):

rileykeane (approved):
All working well! The bugs introduced on the latest develop all seem to be gone now. 

I did notice the media controls (play/pause) on the keyboard (in my case headphones) are paying/pausing the video preview itself but not the clock/scrubber bar. I know an issue was raised to turn off the ability to use device media controls. I'm not sure if this was fixed and has come back at some point, or never fixed at all. 

Regardless, I think this should be fixed in a separate PR. I have opened issue #324 for this. 

@Irmy I believe this was on your teams list, so tagging you here for visibility.

rileykeane:
@Taz17 let's get this merged. I want to push out a new release today for testing.

PR #326, opened 2022-10-01
Add punctuation to local transcription

Mhudlark:
## Summary

Adds punctuation to local transcription.

<hr/>

Punctuation looks better and is needed for take detection.

<hr/>

image

<hr/>

Please test punctuation on a variety of videos. Currently set to 0.3 which is pretty good and I think is what we had originally.

<hr/>

### OS

- [ ] Linux
- [ ] MacOS
- [x] Windows


rileykeane (approved):
Thanks for getting this in so fast. 

Let's get it merged asap so we can get out the new release.

PR #332, opened 2022-10-01
FIX / Bugfix on autosave disallowing mp4 export

itsMoxMox:
## Summary

Fixed a bug where mp4 export would not properly work after an autosave has executed.

<hr/>

### OS

- [ ] Linux
- [ ] MacOS
- [X] Windows


exquisitebin (approved):
Tested locally, do be looking like it works.

PR #333, opened 2022-10-01
feature: restore video preview

MutazAshhab:
## Summary

When the user clicks on a deleted take it should then play the deleted take on the video preview with the following conditions:
1. Scrubber must play the deleted take (meaning 0 is the start of the take, and the right most section of the scrubber is the end of the deleted take)
2. Word highlight must not be displayed.

<hr/>

### Demo recording

Demo recording is to big for github. [Here is the link to it on slack](https://mlvideo.slack.com/files/U0362TURRFC/F045FKXG7H6/restorevideopreviewdemo.mp4)

Note: all the `console.log` I was talking about was only on my local and is not included in this PR

<hr/>

### TODO

1. Find out why the transcript highlight starts as soon as you open the project (and no audio plays) then fix it!!!
<hr/>

### OS

- [x] Linux
- [x] MacOS
- [ ] Windows


rileykeane (changes requested):
This is really awesome work! Thank you for your effort. 

I want to preface this with the fact I have not reviewed the code. I'll leave that to someone more qualified than me. 

For the most part, everything works well. I have noticed a few bugs. Some need to be fixed in this PR and I think some can be handled in a later PR. 

**Bugs that need to be fixed as part of this PR**
- Clicking the edit marker while the restore popover is open throws an error
- When you press enter to restore, the video preview stays stuck on the restored segment and I can't seem to get it to go back to the full video. 

Here is a video showing both issues 
https://drive.google.com/file/d/1tdUBFbCXmqryThb6xhlqEuEiY3IocvRD/view?usp=sharing

**Bugs that can be resolved in a future PR**
- The click away issue is an interesting one. When you don't have a large transcript, this is not a huge issue. However if the transcription block was full of text, I can see how a user would be confused as to how to close the restore popover. One option could be to have the esc key close the restore window. Ideally, clicking words should close it, but I don't think it's worth spending too much time on. 

I found a UI bug with the restore popover when the text is longer than the max width. It should show the start of the sentence and be truncated with an ellipsis but it looks like it's just showing the end. I've opened issue #338 to be fixed in a later PR.

src/renderer/components/VideoPreview/VideoPreviewController.tsx
```diff
@@ -34,6 +34,8 @@ export interface VideoPreviewControllerRef {
 interface Props {
   setTime: (time: number) => void;
   setIsPlaying: (isPlaying: boolean) => void;
+  outputVideoLength: number;
+  setOutputVideoLength: React.Dispatch<React.SetStateAction<number>>;
```
Mhudlark: The type for this can probably just be `(outputVideoLength: number) => void`

src/renderer/components/Editor/PlaybackManager.tsx
```diff
@@ -85,7 +98,13 @@ const PlaybackManager = ({
       setNowPlayingWordIndex(newPlayingWordIndex);
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [time, currentProject?.transcription]);
+  }, [
+    time,
+    currentProject?.transcription,
+    rangeOverride,
```
Mhudlark, line 104: Just want to confirm that we want to add rangeOverride to this useEffect rather than have its own useEffect. Seems to work fine, just checking is all 😄

src/renderer/components/VideoPreview/VideoPreviewController.tsx
```diff
@@ -225,6 +249,28 @@ const VideoPreviewControllerBase = (
...
+    if (currentProject !== null && currentProject?.transcription !== null) {
+      cuts.current = convertTranscriptToCuts(
+        currentProject.transcription,
+        rangeOverride
+      );
+
+      if (cuts.current.length === 0) {
+        return;
+      }
+      const lastCut = cuts.current[cuts.current.length - 1];
```
Mhudlark: Just thinking of DRY as much as applicable, maybe these two lines can be their own function used in both useEffects? eg: `updateOutputVideoLength()`

src/renderer/components/VideoPreview/index.tsx
```diff
@@ -54,6 +54,17 @@ const VideoPreviewBase = ({ src }: Props, ref: Ref<VideoPreviewRef>) => {
         controls: {
           includeControls: [],
         },
+        keyboardShortcuts: {
```
Mhudlark, line 57: Probably unrelated to this PR, has the functionality for the keyboard shortcuts (removed when redux was reverted) for play, skip, etc been added back with useKeypress somewhere in the repo? If not, can you please create an issue or a ticket for it?

rileykeane:
All the issues I mentioned above are now fixed! 🎉

I can see that clicking words now dismisses the popover which is awesome. There is one bug when clicking a word to dismiss the popover and then clicking play. The video plays from the correct position however the clock and highlighted words are reset back to zero/start. 

**Steps to reproduce**
- Open the restore popover 
- Click a word in the transcription that is not the first word 
- Popover is closed and clicked word is highlighted (correct behavior) 
- Click play 

**Expected behavior** 
- Clock is set to start time of the selected word
- Video starts playing from the selected word
- Words highlight on playback starting from the selected word 
- Highlighted words and clock are in sync with the video

**Actual Behavior** 
- Clock is set to the start time of the selected word
- After hitting play, clock is reset to 0
- Highlighted words start from the beginning
- Video does playback does stay in the correct position however is no longer in sync with the clock and highlighted words.

src/renderer/components/VideoPreview/VideoPreviewController.tsx
MutazAshhab: It works, dispite setOutputVideoLength is of time `React.Dispatch<React.SetStateAction<number>>` when pass it down from the Project.tsx

src/renderer/components/Editor/PlaybackManager.tsx
MutazAshhab: Yup, in this useEffect rather than its own.

src/renderer/components/VideoPreview/index.tsx
MutazAshhab: Github issue here: https://github.com/chloebrett/mlvet/issues/343

src/renderer/components/VideoPreview/VideoPreviewController.tsx
MutazAshhab: Done here: [c47c7b6](https://github.com/chloebrett/mlvet/pull/333/commits/c47c7b6658e82dd601349eaa762a40e2848a2878)

src/renderer/components/VideoPreview/VideoPreviewController.tsx
MutazAshhab: I updated it here: [1474959](https://github.com/chloebrett/mlvet/pull/333/commits/1474959a58a034dfdea00a870c7e59eb18e83ca3)

MutazAshhab:
@rileykeane, the issue you are talking about is already in develop and is not caused by the changes i have introduced. Could you create an issue for this?

Irmy (approved):
After such a huge effort I feel like it would be insulting for me to just write "Looks awesome" and approve. So ... looks absolutely epic!

rileykeane (approved):
Thanks @Taz17 

The issue mentioned above has been fixed in #345. This is good to go!

Mhudlark (approved):
Great work fixing all the bugs and preparing for take detection review feature too!

PR #340, opened 2022-10-02
Fix crash when undoing a paste at the end of transcription; add isMer…

### LINKED ISSUE #290: App Crash when undoing paste at the end of the video
Steps to reproduce: 
1. Cut/copy a piece of text
2. Paste this at the end of the project, so that the words pasted are the last ones in the document
3. Undo the paste. The app crashes with this error: 
Error 1 of 1

Expected behaviour: the pasted words are removed from the end of the document after undo

Tested on MacOS ARM, develop branch commit 81776dd4e81258a4766d487e76941bee539c762c

chloebrett:
…geSplitAllowed tests; remove isInOriginalOrder check from merge

Resolves #290 

Notes:
- Fixes crash when undoing a paste to the end of the transcription
- Adds a test suite for the `isMergeSplitAllowed` check which was the source of this bug
- Specifically also adds a test that fails if the range clamp on `src/renderer/store/selection/helpers.ts:27` is removed (the more specific source of this bug)
- Removes the requirement that words to merge must be in originalIndex order - a check that wasn't necessary and that never returned false anyway since the range check had a comparison bug (the conditions in `if (range.startIndex < index || range.endIndex >= index)` should have been inverted)
- Adds a new check that exactly one word must be selected in order to split it (previously, you could select multiple words and it would just split the first one)


jhea0008 (approved):
Thanks for the fix!

PR #341, opened 2022-10-02
Fix bug causing undos of nested deletes to restore more words than should be restored

### LINKED ISSUE #292: Undoing the cut of a section containing a delete marker restores the delete
Steps to reproduce: 
1. delete a word
2. Make a selection that contains the delete marker
3. Delete or cut this section
4. Undo once (Ctrl-Z)

Expected behaviour: 
The text is restored to the state it looks like in step 2

Actual behaviour: 
Full text is restored including what was deleted in step 1

Tested on MacOS ARM, develop branch commit 81776dd4e81258a4766d487e76941bee539c762c

chloebrett:
Resolves #292 

- See bug description - basically, fixes that bug
- Renames `DELETE_SELECTION` to `DELETE_WORDS` as it is a better descriptor of what the action does
- Also updates tests accordingly

Implementation details:
- "delete words" operations are now expressed as a list of indices to delete rather than a single range. The selection is still a single range / contiguous. Using indices allows us to exclude words that are already deleted from the range to delete.

This allows deletions of arbitrary depth (i.e. deletions where the selection contains sections that were already deleted by a previous delete) to "just work" - see the demo below.

Demo:

https://user-images.githubusercontent.com/6735055/193451799-e097e312-81ea-45dc-9bd0-f00e747837e4.mov

### DIFF
--- a/src/collabServer/__tests__/SessionManager.test.ts
+++ b/src/collabServer/__tests__/SessionManager.test.ts
@@ -4,8 +4,7 @@ import { Socket } from 'socket.io-client';
 import { Transcription } from 'sharedTypes';
 import { UndoStack } from 'renderer/store/undoStack/helpers';
 import { CollabServerSessionState } from 'collabServer/types';
-import { rangeLengthOne } from 'renderer/utils/range';
-import { makeDeleteSelection } from 'renderer/store/transcriptionWords/ops/deleteSelection';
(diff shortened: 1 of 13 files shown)


rileykeane (approved):
Tested functionality, all looks good!

Thanks for fixing this one 😄

jhea0008 (approved):
Looks good!

PR #347, opened 2022-10-04
Feature/home button and minimum window size

jhea0008:
## Summary

Adds a home button to the main screen. Adds a minimum window size so as to not crush UI with a too-small window. Re-structures the VideoControllerBox to instead be a menu bar type thing

<hr/>

User Experience

<hr/>

### What did you change?
Home button uses the already present go to home function used by the menu bar item
Took VideoControllerBox out of VideoController.tsx and put it in to project.tsx, and re-structured it so it's more of a menu bar box

<hr/>

Screen Shot 2022-10-04 at 17 47 49

### OS

- [ ] Linux
- [x] MacOS
- [ ] Windows

### DIFF
--- a/src/main/window.ts
+++ b/src/main/window.ts
@@ -22,6 +22,8 @@ const createWindow = async () => {
     show: false,
     width: 1024,
     height: 728,
+    minWidth: 500,
+    minHeight: 250,
     icon: getAssetPath('icon.png'),
     webPreferences: {
       preload: path.join(__dirname, 'preload.js'),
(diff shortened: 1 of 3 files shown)


emmapainter (approved):

Mhudlark (approved):
Good work!
One thing. Currently I think its a bit too easy to go to home, probably because changes are autosaved. As a result, I lose my undo stack. Maybe a modal to check the user wants to leave?

PR #352, opened 2022-10-05
Feature/download model time remaining

### LINKED ISSUE #351: Download model sits at 100% for 20 seconds before saying 'Download complete'

Mhudlark:
## Summary

Adds the time remaining (in minutes or seconds) to the download model UI

<hr/>

In the designs but not included in original local transcription UI PR.

<hr/>

### What did you change?

Added some extra states to the download model redux so that time remaining can be calculated based off of how much progress has been achieved in a certain time interval.

<hr/>

<hr/>

image

image

<hr/>

### Future work

Currently, the time remaining jumps around a fair bit. This can most easily be seen in the number of seconds remaining. We can store lots of time remaining values and then take the average but that's a fair bit of computation for something that happens every couple of milliseconds.

<hr/>

Download the model again. Sorry. Means deleting the `localTranscriptionAssets` folder from your app data path.
On windows that is `C:\Users\<insert-user-here>\AppData\Roaming\Electron\mlvet\localTranscriptionAssets`

<hr/>

### OS

- [ ] Linux
- [x] MacOS
- [x] Windows

### DIFF
--- a/src/renderer/store/downloadModel/reducer.ts
+++ b/src/renderer/store/downloadModel/reducer.ts
@@ -5,7 +5,11 @@ import {
   FINISH_DOWNLOAD,
 } from './actions';
 import { ApplicationStore, initialStore } from '../sharedHelpers';
-import { calculateTimeRemaining, DownloadModel } from './helpers';
+import {
+  calculateTimeRemaining,
+  DownloadModel,
+  DownloadStateUpdatePayload,
+} from './helpers';
 import { Action } from '../action';
 
 // The time between updates to the download time remaining, in milliseconds
@@ -13,7 +17,7 @@ const timeRemainingUpdateInterval = 1000;
 
 const downloadModelReducer: Reducer<
   ApplicationStore['downloadModel'],
-  Action<any>
+  Action<DownloadStateUpdatePayload>
 > = (downloadModel = initialStore.downloadModel, action) => {
   if (action.type === START_DOWNLOAD) {
     return {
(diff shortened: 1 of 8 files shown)


src/renderer/store/downloadModel/reducer.ts
```diff
@@ -18,10 +21,35 @@ const downloadModelReducer: Reducer<
...
       isDownloadComplete: false,
       downloadProgress: 0,
+      lastUpdated: new Date(),
+      previousDownloadProgress: 0,
+      timeRemaining: null,
     } as DownloadModel;
   }
 
   if (action.type === DOWNLOAD_PROGRESS_UPDATE) {
+    const newLastUpdated = new Date();
```
chloebrett: Reducers should be deterministic pure functions so that they are replayable, this breaks that invariant - can instead pass in new date as an argument to the action

Mhudlark: Good point. I've altered the reducer to not be deterministic. However, I didn't want frontend or handler code to have to worry about passing in the date in the action dispatch so its defined inside the action as that is essentially a function. It is no longer defined in the reducer. Thank you! 😃

chloebrett: Yep this is the way to go, nice job. Just on terminology it IS deterministic now not the other way round :)

Mhudlark: oh lol thank you

rileykeane (approved):

PR #354, opened 2022-10-05
Feature/take video preview

MutazAshhab:
## Demo

https://user-images.githubusercontent.com/86769131/194013879-94786830-95c8-46e8-9029-a0dbaa4d5fd3.mp4

### OS

- [x] Linux
- [ ] MacOS
- [ ] Windows

### DIFF
--- a/src/renderer/components/Editor/TakeGroupComponent.tsx
+++ b/src/renderer/components/Editor/TakeGroupComponent.tsx
@@ -9,7 +9,12 @@ import colors from 'renderer/colors';
 import { EditWordState } from 'renderer/store/sharedHelpers';
 import { deleteTakeGroup } from 'renderer/store/takeGroups/actions';
 import { IndexRange, TakeGroup, Transcription, Word } from 'sharedTypes';
-import { isEventInElement, transcriptionContentId } from 'renderer/utils/ui';
+import {
+  getWordElementIdFromClick,
+  isEventInElement,
+  isWordInTake,
+  transcriptionContentId,
+} from 'renderer/utils/ui';
 import { clearRangeOverride } from 'renderer/store/playback/action';
 import { useEventListener } from 'renderer/utils/hooks';
 import TakeComponent from './TakeComponent';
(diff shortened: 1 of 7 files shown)


src/renderer/components/Editor/TakeGroupComponent.tsx
```diff
@@ -95,8 +100,15 @@ const TakeGroupComponent = ({
...
   };
 
-  const clickAway = () => {
-    if (!isFirstTimeOpen) {
+  const onClickAway = (event: Event) => {
+    const isInTranscriptionContent = isEventInElement(
+      event,
+      transcriptionContentId
+    );
+    const isInTakeGroup = isEventInElement(event, takeGroupId);
```
Mhudlark, line 108: What will happen here if the user clicks away from the current take group onto another take group? Do the id's need to be unique?

Mhudlark:
During testing I noticed a couple of bugs.
1. The take video preview cuts off the end of the last word. If I select a take and then play the entire transcription the word will not be cut off but when I play just that take in the preview, it will be.
2. Clicking on a word in the take video preview (to seek to that time) doesn't seem to work

src/renderer/components/Editor/TakeGroupComponent.tsx
MutazAshhab: The ID's dont need to be unique. Clicking on another take group will simply open it

Mhudlark: Is that the behaviour we want? Shouldn't clicking on another take group close the current one? @emmapainter

emmapainter: Yeah @exquisitebin is assigned to "Clicking away from take group closes take group and chooses currently selected take" - so clicking another take group should close the current one

exquisitebin: Thats correct. Clicking on another Take Group closes the currently opened one. That functionality is in develop.

exquisitebin (approved):
Tested locally and seems to be working well. Good job!

PR #358, opened 2022-10-10
Update bug reporter url

LiamTodd:
## Summary

Changes the url for the bug-reporter api.
3 commits, one line changed. Some sort of record(?).

<hr/>

Because the bug-reporter server code was moved to the organisation repository, so the server needed to be redeployed under a new domain. 

<hr/>

IMPORTANT PLS READ!
This depends on the mlvet repo being moved to the organisation account. The bug-report feature will not work until that happens, and will instead tell you to "try again later". This is because the server code already points to the to-be location of the mlvet repo, not the current location.

### OS

- [ ] Linux
- [ ] MacOS
- [x] Windows

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -2,6 +2,6 @@
 export const CURRENT_SCHEMA_VERSION = 2;
 export const URL_USER_FEEDBACK_FORM = 'https://forms.gle/YNozsPZxF4kxSuud8';
 export const BUG_REPORT_API_URL =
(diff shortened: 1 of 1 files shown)


Mhudlark (commented):
MLVET has been moved, does it work now?

LiamTodd:
> MLVET has been moved, does it work now?

Yep, it works

Mhudlark (approved):

PR #363, opened 2022-10-11
Fix export related issues

### LINKED ISSUE #349: No 'export completed' message
When exporting as a MP4 has completed the progress bar goes away but there is nothing to indicate the export has been completed

### LINKED ISSUE #355: Keyboard shortcuts modal needs to be updated for export MP4
Currently the keyboard shortcuts menu, accessed via the 'Help' menu, is out of date. It only references export and does not specifically say export to EDL or MP4. Please add this.

zwan0202:
## Summary

- Updated export to EDL or MP4 keyboard shortcuts in keyboard shortcuts modal
- Added export complete message to indicate the export has been completed

<hr/>

To address issue [#349](https://github.com/MLVETDevelopers/mlvet/issues/349) and [#355 ](https://github.com/MLVETDevelopers/mlvet/issues/355)

<hr/>

### What did you change?

- Updated `KeyboardShortcutsDialog`
- Added an action and logic in reducer to `exportIo` to set `isExporting` to `false` for close export card
- Updated `ExportCard` to change text from `Exporting...` to `Export Complete` and added a close button when export progress reach 1. 
- Design to [Figma](https://www.figma.com/file/vPh49OC7hJpMD7cQMAnBUo/MLVET-UI-Design?node-id=2024%3A3531)

Capture

<hr/>

### OS

- [ ] Linux
- [ ] MacOS
- [x] Windows


src/renderer/components/ExportCard.tsx
```diff
@@ -9,18 +10,38 @@ const Card = styled(Stack)({
...
+  onClose: () => void;
   progress: number;
 }
 
-const ExportCard = ({ progress }: Props) => {
+const ExportCard = ({ onClose, progress }: Props) => {
+  const exportText = progress === 1 ? 'Export Complete' : 'Exporting...';
+  const closeButton =
+    progress === 1 ? (
+      <IconButton
```
Mhudlark, line 25: Mui has a component called IconButton which you can use. Sorry for not mentioning earlier 🤣 
https://mui.com/material-ui/react-button/
https://mui.com/material-ui/api/icon-button/

Mhudlark:
Nice work! (This is first time mp4 export has properly worked for me and its f*cking great!) 
One little comment in code.
Couple of things when testing:
1. Noticed that export throws an error when project has not yet been saved. Could you please fix that (handle error or force save before export) in this PR?
image
2. `temp` folder from cuts is not deleted when individual cuts are deleted. Could you please fix that?

src/renderer/components/ExportCard.tsx
zwan0202: It returns a `IconButton` when progress reaches 1, because we don't want it during the exporting

zwan0202:
> Nice work! (This is first time mp4 export has properly worked for me and its f*** great!) One little comment in code. Couple of things when testing:
> 
> 1. Noticed that export throws an error when project has not yet been saved. Could you please fix that (handle error or force save before export) in this PR?
>    image
> 2. `temp` folder from cuts is not deleted when individual cuts are deleted. Could you please fix that?

1. So we are not allowing users to export when the project is not saved? (sorry i'm not sure how other app handle this)
2. I thought we decided just keep `temp` folder after export

src/renderer/components/ExportCard.tsx
Mhudlark: Yeah I just meant use the MUI IconButton component rather than defining your own. Just more concise that way and its not an extra package.

Mhudlark (approved):
Great work!

chloebrett:
I feel like it should be fine to allow export even if the project isn't saved - other software does this. Just use the media path as the default save location instead of the project file path in that case

emmapainter (approved):
Agreed that being able to export without saving would be ideal. Given we are two days out from the deadline feel free to merge this as is @zwan0202 since I think the current state is fine, but if you have time you can change it to align with Chloe's suggestion.

PR #365, opened 2022-10-11
Feature/undo take select

exquisitebin:
## Undo Take Select
Adds the ability to undo selecting a take.

<hr/>

So that the user is able to intuitively use the undo stack to revert take selection changes.

### Screenshots
<hr/>

https://user-images.githubusercontent.com/58348527/195025475-7f9df7b2-2a6b-4820-a4e9-0c2a560b7b5c.mp4

<hr/>

### OS

- [ ] Linux
- [x] MacOS
- [x] Windows

### DIFF
--- a/src/renderer/components/Editor/TakeComponent.tsx
+++ b/src/renderer/components/Editor/TakeComponent.tsx
@@ -1,12 +1,6 @@
 import styled from '@emotion/styled';
 import { Avatar, Box } from '@mui/material';
-import {
-  IndexRange,
-  TakeGroup,
-  TakeInfo,
-  Transcription,
-  Word,
-} from 'sharedTypes';
+import { IndexRange, Transcription, Word } from 'sharedTypes';
 import React, {
   RefObject,
   useCallback,
--- a/src/renderer/components/Editor/TakeComponent.tsx
+++ b/src/renderer/components/Editor/TakeComponent.tsx
@@ -1,8 +1,12 @@
 import styled from '@emotion/styled';
 import { Avatar, Box } from '@mui/material';
-import { useDispatch } from 'react-redux';
-import { selectTake } from 'renderer/store/takeGroups/actions';
-import { IndexRange, TakeInfo, Transcription, Word } from 'sharedTypes';
+import {
+  IndexRange,
+  TakeGroup,
+  TakeInfo,
+  Transcription,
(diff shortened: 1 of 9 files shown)


src/renderer/components/Editor/TakeComponent.tsx
```diff
@@ -87,19 +93,36 @@ const TakeComponent = ({
...
   const takeRef = useRef<HTMLDivElement>(null);
 
-  const dispatch = useDispatch();
-
   const onSelectTake = useCallback(() => {
-    dispatch(selectTake(takeWords[0].takeInfo as TakeInfo));
+    const takeInfo = takeWords[0].takeInfo as TakeInfo;
+    const takeGroup = transcription.takeGroups.find(
+      (tg) => tg.id === takeInfo.takeGroupId
+    ) as TakeGroup;
```
chloebrett: Can replace cast with an early exit if undefined which is more explicit about edge cases

src/renderer/store/takeGroups/reducer.ts
```diff
@@ -25,6 +29,12 @@ const takeGroupsReducer: Reducer<TakeGroup[], Action<any>> = (
     );
   }
 
+  if (action.type === UNDO_SELECT_TAKE) {
+    const { takeGroup } = action.payload as UndoSelectTakePayload;
+
+    return takeGroups.map((tg) => (tg.id === takeGroup.id ? takeGroup : tg));
```
chloebrett, line 35: Rather than overriding entire take group, wouldn't it make more sense just to store the originally active take ID (or none if there was no selected take in the group) and change the active one to the one that was active before?

exquisitebin: We need to undo both activeTakeIndex (which denotes the selected take) and takeSelected (boolean denoting if a take has been selected yet) and we need the id to find it, which are all the values contained in TakeGroup anyway.

chloebrett: Yep okay, this is fine then. If we weren't in the second last week though I'd say account for possible extensions of the TakeGroup because if the type ever changes this code will break

emmapainter (approved):
Tested this - works great 🙂

MutazAshhab (approved):

PR #367, opened 2022-10-12
Bugfix/play all takes

### LINKED ISSUE #366: Video preview only plays first take
When a take has not been selected yet, the video preview only plays the first take despite all takes appearing on the UI. It should play through all takes in order.

exquisitebin:
## Fix Skipping Takes / Play All Takes
Note: changes may only work for new project files

<hr/>

Before this change, when a take group had no selected/active take (e.g. after the project is created), the playback would only play the first take in the group, skipping over the other takes. Intuitively this should not occur. 

Therefore this bugfix works with the assumption that all takes in a group should be considered part of the transcript until one is selected.

<hr/>

### What did you change?
Words from take group that has no selected take are considered active (return `false` in `isInInactiveTake`)

<hr/>

<hr/>

https://user-images.githubusercontent.com/58348527/195279943-be70f15d-f79b-4b3c-8082-6fe43ff57881.mp4

<hr/>

### OS

- [ ] Linux
- [ ] MacOS
- [x] Windows

### DIFF
--- a/src/sharedUtils.ts
+++ b/src/sharedUtils.ts
@@ -132,7 +132,7 @@ export const isInInactiveTake: (
 
   const takeGroup = takeGroups.find((group) => group.id === takeGroupId);
 
-  if (takeGroup?.activeTakeIndex === takeIndex) {
+  if (takeGroup?.activeTakeIndex === takeIndex || !takeGroup?.takeSelected) {
     return false;
   }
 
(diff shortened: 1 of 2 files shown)


zwan0202 (approved):
Thanks for fixing this!

chloebrett (approved):
Makes sense

PR #382, opened 2022-10-14
Bugfix/copy paste mac

### LINKED ISSUE #264: Cut/copy/paste/select all doesn't work in regular text fields on mac

### LINKED ISSUE #373: Unable to paste transcription API key into input field
2 users reported that they were unable to paste the API key into the required field, and instead needed to manually type it in

User OS - Mac

chloebrett:
Resolves #264 
Resolves #373

The problem this fixes:
- On Mac, cut/copy/paste/selectAll worked in the _app_ for modifying the transcription with our custom logic, but didn't work in regular textfields (e.g. pasting API key)

The solution:
- Makes cut/copy/paste/selectAll work natively on Mac
- The app menu first does the native action and then the front end intercepts the native action to do the app-specific action as well
- For selectAll, this is tied to the 'Ctrl/Cmd+A' key shortcut, so "select all" in the app doesn't work when clicking in the menu - this was the case on Linux/Windows before anyway so I don't think it's a huge issue to now have that also be the case on Mac
- Additional note, cut/copy/paste are now unconditionally enabled in the menu - before, they were disabled if there was no transcription range selected

Please test on
- [x] Mac
- [ ] Windows
- [ ] Linux

So we can confirm it doesn't break anything else as I've only tested on Mac


MutazAshhab:
Code looks good. I've tested it on linux, and it all woks as expected.

jhea0008:
Could we get native support for undo alongside the cut, copy and paste? Otherwise all good

MutazAshhab (approved):

chloebrett:
> Could we get native support for undo alongside the cut, copy and paste? Otherwise all good

We could make it work with keyboard shortcuts, but if we do this then clicking the menu buttons will only do the native event and not the app-specific undo/redo which is a bit sad. Can't think of a way around this right now. I think the current behaviour while not ideal is slightly better

chloebrett:
Merging now and if there are issues / follow ups let's put them in the issues tab

PR #383, opened 2022-10-14
Add tooltips to items in the menu bar

### LINKED ISSUE #339: Entire feedback button should be clickable
Current behaviour: only icon is clickable

Screen Shot 2022-10-02 at 9 19 58 pm

Expected behaviour: the entire area around the icon and the text are a single click target

Reasoning: I went to click it to test it out and assumed that clicking the text would make the popup appear, but you actually have to click the icon itself - and there is also almost no padding, so you have to click it exactly. There is basically no downside to making the whole area clickable since there are no other click targets nearby to get confused with.

jhea0008:
## Summary

Improves the tooltips to be closer to the design and adds them to the video controller

<hr/>

aesthetics

<hr/>

### What did you change?
New componet: ProjectToolTip with styling in it

<hr/>

Screen Shot 2022-10-14 at 12 51 53

<hr/>

### OS

- [ ] Linux
- [X] MacOS
- [ ] Windows

### DIFF
--- a/src/renderer/components/Editor/MenuBarButton.tsx
+++ b/src/renderer/components/Editor/MenuBarButton.tsx
@@ -15,11 +15,7 @@ const MenuBarButton = ({
 }: MenuBarButtonProps) => {
   return (
     <ProjectTooltip title={text}>
-      <IconButton
-        sx={{ color: colors.grey[300], display: 'flex' }}
-        onClick={onClick}
-        {...rest}
-      >
+      <IconButton sx={{ color: colors.grey[300] }} onClick={onClick} {...rest}>
         <div>{children}</div>
       </IconButton>
     </ProjectTooltip>
--- a/src/renderer/components/Editor/MenuBarButton.tsx
+++ b/src/renderer/components/Editor/MenuBarButton.tsx
@@ -1,0 +1,29 @@
+import { ButtonProps, colors, IconButton } from '@mui/material';
+import ProjectTooltip from './ProjectTooltip';
+
+interface MenuBarButtonProps extends ButtonProps {
+  text: string;
+  children: React.ReactNode;
+  onClick: () => void;
+}
+
+const MenuBarButton = ({
+  text,
+  children,
+  onClick,
+  ...rest
+}: MenuBarButtonProps) => {
+  return (
+    <ProjectTooltip title={text}>
+      <IconButton
+        sx={{ color: colors.grey[300], display: 'flex' }}
+        onClick={onClick}
+        {...rest}
+      >
+        <div>{children}</div>
+      </IconButton>
+    </ProjectTooltip>
+  );
+};
+
+export default MenuBarButton;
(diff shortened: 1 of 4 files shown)


jhea0008:
Addresses issue: 
https://github.com/MLVETDevelopers/mlvet/issues/339

src/renderer/pages/Project.tsx
```diff
@@ -117,56 +118,63 @@ const ProjectPage = () => {
...
-                      <ContentCopyIcon />
-                    </IconButton>
-                    <IconButton
-                      sx={{ color: colors.grey[300] }}
-                      onClick={pasteText}
-                      title="Paste Text"
-                    >
-                      <ContentPasteIcon />
-                    </IconButton>
+                    <ProjectTooltip title="Exit to Home">
```
chloebrett: This whole list is pretty repetitive, could we just have a list of objects containing title, action and icon and then render the buttons by mapping over the list? Or at least a helper function to generate a button given a tooltip title, action and icon

src/renderer/components/Editor/ProjectTooltip.tsx
```diff
@@ -0,0 +1,21 @@
+import { styled, Tooltip, tooltipClasses, TooltipProps } from '@mui/material';
+import colors from 'renderer/colors';
+
+const ProjectTooltip = styled(({ className, ...props }: TooltipProps) => (
```
chloebrett: I think there's too much going on inline here, can we separate out into a couple of steps so it's clearer what's going on? Might be able to get a definitive type here as well if that's still an issue

src/renderer/pages/Project.tsx
jhea0008: Created a new component to encapsulate more of the features common to all the buttons

src/renderer/components/Editor/ProjectTooltip.tsx
jhea0008: Yea, I just copy pasted this from the mui examples, so I'm not 100% clear on it either. I have separated it out a bit I think it's clearer now

src/renderer/components/Editor/ProjectTooltip.tsx
```diff
@@ -0,0 +1,22 @@
...
+    color: colors.grey[900],
+    fontSize: '16px',
+    fontFamily: 'Rubik',
+  },
+}));
+
+export default function toolTipFunction(props: TooltipProps) {
+  return (
+    <ProjectTooltip title={props.title}>
+      <span>{props.children}</span>
```
Mhudlark: Can it just be `{props.children}`? Is the wrapping <span> needed?

src/renderer/components/Editor/MenuBarButton.tsx
```diff
@@ -0,0 +1,25 @@
...
+const MenuBarButton = ({
+  text,
+  children,
+  onClick,
+  ...rest
+}: MenuBarButtonProps) => {
+  return (
+    <ProjectTooltip title={text}>
+      <IconButton sx={{ color: colors.grey[300] }} onClick={onClick} {...rest}>
+        <div>{children}</div>
```
Mhudlark: Can it just be `{props.children}`? Is the wrapping <div> needed?

Mhudlark (approved):
Nice work!
Approving because it works well when tested and little time. Few small comments regarding code implementation, would be nice if they were fixed before merging.

emmapainter (approved):
Thanks for smashing this out Justin, this will do the job just fine 😄

PR #386, opened 2022-10-15
Chore/update local transcription assets paths

Mhudlark:
## Summary

Changes the local transcription asset paths to be stored in a configurable file.

<hr/>

Allows the end user to change the URL that the model is downloaded from. Means that the AWS S3 bucket holding the transcription assets can be removed and after 1 year the local transcription feature will still work.

<hr/>

### What did you change?

- Updated the local transcription asset download URLs to be stored in a json file that can be configured by developers or by the end user (does not get compiled during packaging)
- Added a utility function to extract the model correctly from the downloaded files and put it in the right location for local transcription. This means that if the download file dir structure changes, the download function will still work.
- Updates the URLs to be:
    - 1. Libs will download from a git repository in the MLVETDevelopers organisation
    - 2. Models will download directly from vosk webpage

<hr/>

- I have tested both develop and packaged app (as they have slightly different `assets` paths)

<hr/>

### OS

- [ ] Linux
- [ ] MacOS
- [x] Windows


Mhudlark:
Tests are failing because of issue being fixed in https://github.com/MLVETDevelopers/mlvet/pull/384

tngu0041 (approved):

MutazAshhab (approved):

PR #388, opened 2022-10-16
Feature/take group improvements

MutazAshhab:
## Summary
- Users can now choose takes by pressing the the digits on their keyboard (numpad digits as well)
- Users can now close take group by pressing enter (numpad enter as well). Pressing this will close the take group.

<hr/>

Feature

<hr/>

### Demo

https://user-images.githubusercontent.com/86769131/196017400-66f90a2d-2196-44d9-a03f-6cb132e9a814.mp4

<hr/>

### OS

- [x] Linux
- [ ] MacOS
- [ ] Windows


src/renderer/components/Editor/TakeGroupComponent.tsx
```diff
@@ -224,6 +231,45 @@ const TakeGroupComponent = ({
...
+      selectTake(takeGroup.activeTakeIndex);
+    } else if (isDigit(eventCode[eventCode.length - 1])) {
+      const numberPressed = Number(eventCode[eventCode.length - 1]);
+      selectTake(numberPressed - 1); // minus one since the UI displays the index + 1
+    }
+  };
+
+  useKeypressWithEvent(
+    handleKeypressWithEvent,
+    true,
```
Mhudlark: Instead of `true`, could this be `isTakeGroupOpened`?

emmapainter (changes requested):
Thanks for doing this! 

For some reason there is some unexpected behaviour. I tried to take a screen recording but since I was using keyboard shortcuts it isn't clear what I'm doing. Here's what happened:

1. Open a take group
2. Using number keys to toggle between takes works great!
3. Press enter to confirm current selection - still working as expected
4. Reopen the take group by clicking the take indicator
5. Use number keys to toggle between takes - now it does some funky behaviour where it starts selecting takes for other take groups?

Here's the screen recording:
https://user-images.githubusercontent.com/100325471/196321701-b217f1b5-645d-4a89-a2bc-53b884db12f8.mov


Also if I press enter while no take groups are active but no takes have been selected it starts selecting takes on the take groups (should do nothing).

PR #389, opened 2022-10-16
Chore/async local transcription

### LINKED ISSUE #327: UI hangs while transcribing locally
Currently, the UI is non-responsive when transcribing locally

Mhudlark:
## Summary

Makes local transcription asynchronous.

<hr/>

The UI was hanging during local transcription.

<hr/>

### What did you change?

- Made a cmd-line script `asyncVosk.ts` that runs local transcription
- Made the `voskTranscribeFunction` start the `asyncVosk.ts` script in a separate process using `exec`
- Changed vosk so that the path to the Vosk DLL libs is passed into vosk, not retrieved from within vosk

<hr/>

- Used `child_process.exec` and not `fork` or `spawn` because I couldn't get them to work.

<hr/>

Try interacting with UI during local transcription.

<hr/>

### OS

- [ ] Linux
- [ ] MacOS
- [x] Windows


Mhudlark:
Note this bug which exists for both cloud and local (but consequences for local are more severe).
https://github.com/MLVETDevelopers/mlvet/issues/390

Mhudlark:
Update: This is not working with the packaged app. Turns out child_process doesn't really work from inside an app packaged with asar.

